PR #4, opened 2018-10-02
Travis support

mrlukasbos:


mrlukasbos:
I made quite some  progress on Travis support. Unfortunately, the system does not work  yet. One of the main causes of this is the dependency Roboteam_utils, which is quite complex and needs to be built also. This Pull request would be interesting again when we got rid of roboteam utils in roboteam_world and roboteam_ai.

PR #28, opened 2018-11-15
Safe constant handling

rolfvdhulst:
If anyone knows a better solution for handling static const doubles as class member definitions please let me know. googled for 30+ minutes.


mrlukasbos:
Using a namespace the problem can be solved as well.

PR #62, opened 2019-01-08
make the ball a pointer

### LINKED ISSUE #36: Make ball class robust to disappearances
Ideally it would become a pointer in World::getBall() so that it can also dissappear, like our Robots.

mrlukasbos:
### Summary
The system would crash when skills were run when no ball has been seen in the field yet. This should prevent it. What it does not fix is that the ball does not 'dissappear' in the system when it physically dissapears. (it keeps the last seen location of the ball, this might be fixed in World.cpp in another PR) 

### Related issues
fixes #36 

### Comments for the reviewer
In the skills and conditions that require the ball a check is needed. I might have missed some of these, since there are a lot of them.

**functionality**  
- [ ] The code is properly tested
- [ ] The tests are passing
- [ ] The code works as expected
- [ ] The code is understandable and easy to read
- [ ] Proper abstractions
- [ ] No duplicate code

**style**  
- [ ] Variables/objects/classes have proper names
- [ ] No unused variables
- [ ] No magic numbers
- [ ] The code is well documented
- [ ] No unused imports
- [ ] No leftover code / unused comments
- [ ] Proper indentation
- [ ] Proper use of namespaces


roboteam_ai/src/conditions/BallKickedToOurGoal.cpp
```diff
@@ -7,16 +7,17 @@ namespace rtt {
...
+    auto ball = World::getBall();
     Vector2 goalCentre = Field::get_our_goal_center();
     double goalWidth = Field::get_field().goal_width;
     double margin = constants::BALL_TO_GOAL_MARGIN;
     Vector2 lowerPost = goalCentre + Vector2(0.0, - (goalWidth + margin));
     Vector2 upperPost = goalCentre + Vector2(0.0, goalWidth + margin);
-    Vector2 ballPos = ball.pos;
-    Vector2 ballPredPos = Vector2(ball.pos) + Vector2(ball.vel)*constants::BALL_TO_GOAL_TIME;
+    Vector2 ballPos = ball->pos;
+    Vector2 ballPredPos = Vector2(ballPos) + Vector2(ballPos)*constants::BALL_TO_GOAL_TIME;
```
rolfvdhulst: You changed a velocity to position here, please keep using the velocity (ball->vel)

roboteam_ai/src/skills/Attack.cpp
```diff
@@ -7,29 +7,22 @@
...
-    robot = getRobotFromProperties(properties);
-}
+Attack::Attack(string name, bt::Blackboard::Ptr blackboard) : Skill(std::move(name), std::move(blackboard)) { }
 
 /// Get an update on the skill
 bt::Node::Status Attack::onUpdate() {
-    updateRobot();
-    if (! robot) return Status::Running;
-    Vector2 ball = World::getBall().pos;
+    if (!robot || !ball) return Status::Running;
```
rolfvdhulst: Maybe make it fail in case there is no ball visible? We also keep the ball/robot for some time in Roboteam World already.

rolfvdhulst: Is probably not relevant to this PR but for another issue

roboteam_ai/src/skills/Dribble.cpp
```diff
@@ -101,17 +99,14 @@ void Dribble::onInitialize() {
 }
 
 Dribble::Status Dribble::onUpdate() {
-
-    ball = World::getBall(); //TODO: sanity checking if ball is actually there?
-
-    if (currentProgress == Progression::FAIL) {
+    if (currentProgress == Progression::FAIL || !ball) {
```
rolfvdhulst: This is actually undesirable behavior as we saw in the designlab, as we want to robot to keep dribbling if it is in front of the ball from the camera's angle.

roboteam_ai/src/skills/Dribble.cpp
```diff
@@ -101,17 +99,14 @@ void Dribble::onInitialize() {
...
         return Status::Failure;
     }
     else if (currentProgress == Progression::WAITING) {
         return Status::Waiting;
     }
 
-    deltaPos = targetPos - Vector2(ball.pos.x, ball.pos.y);
+    deltaPos = targetPos - Vector2(ball->pos.x, ball->pos.y);
```
rolfvdhulst: You can write Vector2(ball->pos) here.

roboteam_ai/src/skills/GetBall.cpp
```diff
@@ -60,8 +58,8 @@ void GetBall::onInitialize() {
     count = 0;
 }
 GetBall::Status GetBall::onUpdate() {
-    ball = World::getBall(); //TODO: sanity checking if ball is actually there
-    deltaPos = Vector2(ball.pos.x, ball.pos.y) - Vector2(robot->pos.x, robot->pos.y);
+    if (!ball) return Status::Running;
```
rolfvdhulst: Should be Status::failure

roboteam_ai/src/skills/GoToPos.cpp
```diff
@@ -30,19 +27,17 @@ void GoToPos::onInitialize() {
 
 /// Get an update on the skill
 bt::Node::Status GoToPos::onUpdate() {
-    updateRobot();
     if (! robot) return Status::Running;
+    if (!ball) return Status::Running;
```
rolfvdhulst: We do not want to check for the ball if the GoToPos does not need the ball to compute its path. Surround this with an if (goToBall or goBehindBall) statement.

roboteam_ai/src/skills/GoToPosLuTh.cpp
```diff
@@ -33,8 +33,10 @@ GoToPosLuTh::Status GoToPosLuTh::onUpdate() {
     displayData.clear();
 
     if (goToBall) {
-        auto ball = World::getBall();
-        targetPos = ball.pos;
+        if (!ball) {
+            ROS_ERROR("GoToPosLuTh: No ball found! assuming (%f, %f)", ball->pos.x, ball->pos.y);
```
rolfvdhulst: ball is a null pointer in this case, so this will segfault as you try to access it in the ros error.

roboteam_ai/src/skills/Skill.cpp
```diff
@@ -6,6 +6,8 @@ namespace ai {
 Skill::Skill(std::string name, bt::Blackboard::Ptr blackboard)
         :bt::Leaf(std::move(name), std::move(blackboard)), ioManager(false, true) {
     robot = std::make_shared<roboteam_msgs::WorldRobot>();
+    ball = std::make_shared<roboteam_msgs::WorldBall>();
```
rolfvdhulst: I like this.

roboteam_ai/src/utilities/Coach.cpp
```diff
@@ -105,12 +105,13 @@ int Coach::pickOpponentToCover(int selfID) {
 }
 
 Vector2 Coach::getPositionBehindBall(double distanceBehindBall) {
-    auto ball = static_cast<Vector2>(World::getBall().pos);
+    auto ball = World::getBall();
+    auto ballPos = static_cast<Vector2>(ball->pos);
+    if (!ball) { ROS_ERROR("No ball found. assuming ball at position (%f, %f)", ballPos.x, ballPos.y); }
```
rolfvdhulst: It will segfault on 109 if 110 is true.

roboteam_ai/src/utilities/RobotDealer.cpp
```diff
@@ -78,15 +78,27 @@ int RobotDealer::claimRobotForTactic(RobotDealer::RobotType feature, std::string
                 return - 1;
 
             case closeToBall: {
-                rtt::Vector2 ball = rtt::ai::World::getBall().pos;
-                id = getRobotClosestToPoint(ids, ball);
+                auto ball = rtt::ai::World::getBall();
+                rtt::Vector2 ballPos = ball->pos;
```
rolfvdhulst: same as above, it will segfault (null pointer)

roboteam_ai/src/utilities/RobotDealer.cpp
```diff
@@ -78,15 +78,27 @@ int RobotDealer::claimRobotForTactic(RobotDealer::RobotType feature, std::string
...
+                            static_cast<int>(ballPos.x), static_cast<int>(ballPos.y));
+                }
+                id = getRobotClosestToPoint(ids, ballPos);
                 break;
             }
 
             case betweenBallAndOurGoal: {
-                rtt::Vector2 ball = rtt::ai::World::getBall().pos;
+                auto ball = rtt::ai::World::getBall();
+                rtt::Vector2 ballPos = ball->pos;
```
rolfvdhulst: idem to above

roboteam_ai/src/skills/Attack.cpp
mrlukasbos: No. If there ball is not there it should not immediately fail: it can also have lost the ball temporarily (i.e. a robot is in front of it); then it should not instantly fail

PR #68, opened 2019-01-08
smoothed controlgotoposbasic

kJh19:
### Summary
I (re)moved the static class
and smoothed some equations

### Related issues
If there are related issues, reference them here.

fixes ...  
fixes ...

### Comments for the reviewer
If there are comments for the reviewer, leave them here

**functionality**  
- [ ] The code is properly tested
- [ ] The tests are passing
- [ ] The code works as expected
- [ ] The code is understandable and easy to read
- [ ] Proper abstractions
- [ ] No duplicate code

**style**  
- [ ] Variables/objects/classes have proper names
- [ ] No unused variables
- [ ] No magic numbers
- [ ] The code is well documented
- [ ] No unused imports
- [ ] No leftover code / unused comments
- [ ] Proper indentation
- [ ] Proper use of namespaces

### DIFF
--- a/roboteam_ai/src/control/ControlGoToPos.cpp
+++ b/roboteam_ai/src/control/ControlGoToPos.cpp
@@ -87,10 +87,7 @@ void ControlGoToPos::goToPosBasic(RobotPtr robot, Vector2 &targetPos) {
 //        return;
 //    }
 
-    static Controller pidBasic(3, 0, 0.5);
-    Vector2 error;
-    error.x = targetPos.x - robot->pos.x;
-    error.y = targetPos.y - robot->pos.y;
+    Vector2 error = targetPos - robot->pos;
     double dist = error.length();
     static bool far = true;
     if (dist > rtt::ai::constants::ROBOT_RADIUS and ! far) {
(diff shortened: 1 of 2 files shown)


roboteam_ai/src/control/ControlGoToPos.cpp
```diff
@@ -101,11 +98,14 @@ void ControlGoToPos::goToPosBasic(RobotPtr robot, Vector2 &targetPos) {
         pidBasic.setD(0);
         far = false;
     }
+    //if (dist > rtt::ai::constants::ROBOT_RADIUS) {
```
rolfvdhulst: No commented code please.

roboteam_ai/src/control/ControlGoToPos.cpp
```diff
@@ -101,11 +98,14 @@ void ControlGoToPos::goToPosBasic(RobotPtr robot, Vector2 &targetPos) {
...
+    //    std::cout << "Robot:" << robot->id << "TargetPos:" << dist << std::endl;
+    //}
     Vector2 delta = pidBasic.controlPIR(error, robot->vel);
     Command command;
     command.id = robot->id;
     command.use_angle = 1;
-    command.w = static_cast<float>(delta.angle());
+    command.w = 0;
```
rolfvdhulst: why did you change this functionality?

kJh19: because they spin more with 0, the angle changes every time, and so I felt it would interfere with possible debugging

rolfvdhulst: So now they always look at the same way? We want our robots to face the direction they are going.

kJh19: yeah, but let's fix that when we fix the constant turning, inputting a changing angle might also make them turn constantly, so I would like to keep those issues separate.

rolfvdhulst: Still: Why change it here then? Make a separate branch for this and fix the entire problem with turning there at once. We do not want partial solutions on development, because then people might think this was the intended functionality.

kJh19: yeah, but wasn't the problem somewhere in grsim? This is the least bug-prone/least-intended-side-effects version

mrlukasbos:
@kJh19  don't forget to update according to the review!

PR #96, opened 2019-01-18
Fix rotating of robotcommand

rolfvdhulst:
This rotates the Robot Command by 180 degrees if we are on the right. Tested with GrSim. Keeper and Field properties have been tested too to ensure they rotate correctly.

Tests are passing.


roboteam_ai/src/skills/Skill.cpp
```diff
@@ -20,7 +21,13 @@ Skill::Skill(std::string name, bt::Blackboard::Ptr blackboard) : bt::Leaf(std::m
 
 }
 void Skill::publishRobotCommand(roboteam_msgs::RobotCommand cmd) {
-    ioManager.publishRobotCommand(cmd);
+    ros::NodeHandle nh;
+    std::string ourSideParam;
+    nh.getParam("our_side",ourSideParam);
+    if(ourSideParam=="right"){
+        rotateRobotCommand(cmd);
```
mrlukasbos: This is a c-style solution which is fast but not very intuitive in readability.

rolfvdhulst: I fixed this in new commits

PR #109, opened 2019-01-23
Formations

mrlukasbos:
Puts the keeper in front of the goal, and puts the rest of the robots in a line at 1/4 of the field, aimed at opponents goal


roboteam_ai/src/skills/EnterFormation.cpp
```diff
@@ -0,0 +1,44 @@
...
+bt::Node::Status EnterFormation::onUpdate() {
+    auto robotPos = rtt::Vector2(robot->pos);
+    Vector2 targetLocation = coach::Coach::getFormationPosition(robot->id);
+    Vector2 targetToLookAtLocation = Field::get_their_goal_center();
+
+    roboteam_msgs::RobotCommand cmd;
+    cmd.id = robot->id;
+    cmd.use_angle = 1;
+
+    if (robotPos.dist(targetLocation) > 0.1) {
```
rolfvdhulst: I don't really like this design if I'm honest, because both getting into position and setting final angle. Also, goToPos already has a check like this in it (I believe with currently >0.3) so it will stop the velocities before it is actually there.

PR #121, opened 2019-01-31
Go Around Pos

rolfvdhulst:
Adds the skill GoAroundPos, which goes around the ball primarily.


roboteam_ai/src/ApplicationManager.cpp
```diff
@@ -103,7 +103,7 @@ void ApplicationManager::notifyTreeStatus(bt::Node::Status status) {
     case Status::Running:break;
     case Status::Success:ROS_INFO_STREAM("Status returned: Success");
         ROS_INFO_STREAM(" === TREE CHANGE === ");
-        BTFactory::setCurrentTree("haltStrategy");
+        BTFactory::setCurrentTree("keeperStrategy");
```
mrlukasbos: you need to initialize on halt

roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -198,14 +198,27 @@ int ControlUtils::rotateDirection(double currentAngle, double targetAngle){
 }
 
 /// Limits velocity to maximum velocity
-Vector2 ControlUtils::VelocityLimiter(Vector2 vel) {
-    if (vel.length() > rtt::ai::constants::MAX_VEL) {
-        vel = vel.stretchToLength(rtt::ai::constants::MAX_VEL);
+Vector2 ControlUtils::VelocityLimiter(Vector2 vel,double maxVel) {
```
mrlukasbos: why?

roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -198,14 +198,27 @@ int ControlUtils::rotateDirection(double currentAngle, double targetAngle){
...
+    if (vel.length() > maxVel) {
+        vel = vel.stretchToLength(maxVel);
         return vel;
     }
     else return vel;
 }
 
+Vector2 ControlUtils::VelocityLimiter(Vector2 vel, double maxVel,double minVel){
```
mrlukasbos: You can merge this with the previous function; by initializing the parameters in de .h file

roboteam_ai/src/skills/GoAroundPos.h
```diff
@@ -0,0 +1,34 @@
...
+//
+
+#ifndef ROBOTEAM_AI_GOAROUNDPOS_H
+#define ROBOTEAM_AI_GOAROUNDPOS_H
+#include "Skill.h"
+namespace rtt {
+namespace ai {
+
+class GoAroundPos : public Skill {
+        enum Progression{FAIL,ROTATING,STOPPING,DONE};
```
mrlukasbos: add the private keyword for consistency

roboteam_ai/src/treeinterp/Switches.cpp
```diff
@@ -81,7 +82,7 @@ std::vector<std::string> Switches::tacticJsonFileNames =
          "EnterFormationTactic",
          "BallPlacementUsTactic",
          "AvoidBallForBallPlacementTactic",
-         "randomTactic" // used for testing, do not remove it!
+         "randomTactic", // used for testing, do not remove it!
```
mrlukasbos: remove this comma

roboteam_ai/src/utilities/Constants.h
```diff
@@ -62,6 +62,14 @@ const double GETBALL_OVERSHOOT=.02;//m
 //GoToPos
 const double MAX_CALCULATION_TIME=20.0; //max time in ms
 
+//GoAroundPos
+const double MAX_GOAROUND_STOP_TIME=2.0; //s
```
mrlukasbos: nice

roboteam_ai/src/control/ControlUtils.cpp
rolfvdhulst: so it works for any velocity (e.g. you can pass a maximum velocity) (see header file)

roboteam_ai/src/control/ControlUtils.cpp
rolfvdhulst: No, they do functionally different things as one checks for the minimum speed and the other does not.

roboteam_ai/src/ApplicationManager.cpp
rolfvdhulst: This is not initialization, but the tree it switches to after completing a tree.

roboteam_ai/src/control/ControlUtils.cpp
rolfvdhulst: It defaults to the MAX_VEL

roboteam_ai/src/control/ControlUtils.cpp
rolfvdhulst: No, since you are then still checking for a minimum speed where in the other one you're not. This gives weird behaviour if minVel>maxVel

PR #128, opened 2019-02-01
Interface tests

mrlukasbos:


roboteam_ai/test/InterfaceTests/MainWindowTest.cpp
```diff
@@ -46,7 +48,7 @@ TEST(MainWindowTest, it_shows_the_visualizer_properly) {
     vis->setShowTactics(true);
     vis->setShowVelocities(true);
 
-    ASSERT_TRUE(vis->showAngles
+    EXPECT_TRUE(vis->showAngles
```
rolfvdhulst: instead of making this an && statement I would say run a separate EXPECT_TRUE for all of them, so that you can see which button specifically breaks every time instead of knowing that one of them breaks.

roboteam_ai/test/InterfaceTests/MainWindowTest.cpp
```diff
@@ -64,18 +66,68 @@ TEST(MainWindowTest, it_shows_the_visualizer_properly) {
     vis->setShowTactics(false);
     vis->setShowVelocities(false);
 
-    ASSERT_FALSE(vis->showAngles
+    EXPECT_FALSE(vis->showAngles
```
rolfvdhulst: Similar to above

PR #130, opened 2019-02-07
Feature/constants switching

mrlukasbos:
This PR does a major change to how constants work in Roboteam AI.
It allows us to have different variables when running GrSim or not. 

The constants need to be initialized since it needs `ros::init()` to be run beforehand, otherwise it can't check for the presence of grsim. Therefore **it is not possible anymore to initialize constant variables in static classes!**

Please be critical in the review, and check if all constants have been edited correctly!

### DIFF
--- a/roboteam_ai/src/control/ControlUtils.cpp
+++ b/roboteam_ai/src/control/ControlUtils.cpp
@@ -23,7 +23,7 @@ double ControlUtils::calculateAngularVelocity(double robotAngle, double targetAn
         direction = - 1;                //  clockwise rotation
     }
     if (angleDiff > 1)angleDiff = 1;
-    return direction*(std::pow(rotFactor, angleDiff - 1)*rtt::ai::constants::MAX_ANGULAR_VELOCITY - 1/rotFactor);
+    return direction*(std::pow(rotFactor, angleDiff - 1) * Constants::getDouble("MAX_ANGULAR_VELOCITY") - 1/rotFactor);
 }
 
 // Efficient implementation, see this: https://stackoverflow.com/questions/2049582/how-to-determine-if-a-point-is-in-a-2d-triangle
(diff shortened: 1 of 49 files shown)


roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -159,11 +159,11 @@ bool ControlUtils::lineSegmentsIntersect(Vector2 lineAStart, Vector2 lineAEnd, V
 rtt::Arc ControlUtils::createKeeperArc() {
     double goalwidth = rtt::ai::Field::get_field().goal_width;
     Vector2 goalPos = rtt::ai::Field::get_our_goal_center();
-    double diff = rtt::ai::constants::KEEPER_POST_MARGIN - rtt::ai::constants::KEEPER_CENTREGOAL_MARGIN;
+    double diff = Constants::getDouble("KEEPER_POST_MARGIN") - Constants::getDouble("KEEPER_CENTREGOAL_MARGIN");
 
     double radius = diff*0.5 + goalwidth*goalwidth/(8*diff); //Pythagoras' theorem.
     double angle = asin(goalwidth/2/radius); // maximum angle (at which we hit the posts)
-    Vector2 center = Vector2(goalPos.x + rtt::ai::constants::KEEPER_CENTREGOAL_MARGIN + radius, 0);
+    Vector2 center = Vector2(goalPos.x +  Constants::getDouble("KEEPER_CENTREGOAL_MARGIN") + radius, 0);
```
RobZelluf: one space too many between + and Constants

roboteam_ai/src/utilities/Constants.cpp
```diff
@@ -0,0 +1,187 @@
...
+    std::cerr << "Double constant not found: " << name.c_str() << std::endl;
+    return 0.0;
+}
+
+int Constants::getInt(const std::string &name) {
+    if (integers.find(name) != integers.end() ) {
+        return integers[name];
+    }
+    std::cerr << "Integer constant not found: " << name.c_str() << std::endl;
+    return 0;
```
RobZelluf: Do you want these to return a default? This could definitely break stuff and you might not always check the error log.

roboteam_ai/src/utilities/Constants.cpp
```diff
@@ -0,0 +1,187 @@
...
+    return { 0, 0, 0, 0 }; // transparent
+}
+
+void Constants::init() {
+    ros::NodeHandle nh;
+    std::string robotOutputTarget;
+    nh.getParam("robot_output_target", robotOutputTarget);
+    useGrSim = robotOutputTarget != "serial"; // use the grSim as default; only serial is specifically defined
+
+    bools["SHOW_LONGEST_TICK"]              = true;
```
RobZelluf: Will this allow CLion to auto-finish constant names for you. That's what I liked about the old implementation of constants. Besides that I really like this!

roboteam_ai/src/utilities/Constants.cpp
mrlukasbos: well... i have to

mrlukasbos:
I did not like the fact that autocomplete would not be possible anymore. So I started another PR which is even better

PR #132, opened 2019-02-08
Feature/constants switching 2

mrlukasbos:
replaces  #130
this implementation is neater. Only values that are dependent on ros need to be initialzed after ros::init(), not all.


mrlukasbos:
thanks, you too!

PR #133, opened 2019-02-08
Feature/condition tests

mrlukasbos:


roboteam_ai/test/ConditionTests/BallOutOfFieldtest.cpp
```diff
@@ -0,0 +1,51 @@
+//
+// Created by mrlukasbos on 8-2-19.
+//
+
+//
```
rolfvdhulst: Copy pasta leftovers I think :P

PR #134, opened 2019-02-12
Goal visibility calculations

mrlukasbos:
This PR adds functions that detect blockades in front of a goal of choice.

- If two robots overlap than they become one obstacle.
- If a robot is at exactly the same point as the point which you try to calculate, the robot itself is not seen as an obstacle
- you can get an array with the obstacles
- you can get an array with the visible parts
- you can get the percentage of visible parts of the goal, rounded to integers (maybe make this floats?) 

fully tested!

Fully tested and should be fully functional.


roboteam_ai/src/utilities/Field.cpp
```diff
@@ -78,6 +81,153 @@ bool Field::pointIsInField(Vector2 point, float margin) {
...
 }
 
+double Field::getPercentageOfGoalVisibleFromPoint(bool ourGoal, Vector2 point){
+    auto field = Field::get_field();
+    double goalWidth = field.goal_width;
+    double blockadeLength = 0;
+    for (auto const &blockade : getBlockadesMappedToGoal(ourGoal, point)) {
+        blockadeLength += blockade.first.dist(blockade.second);
+    }
+    return std::max(100 - round(blockadeLength/goalWidth * 100), 0.0);
```
rolfvdhulst: I like fmax a bit better since you're also using round() from the math.h header for consistency

roboteam_ai/src/utilities/Field.cpp
```diff
@@ -78,6 +81,153 @@ bool Field::pointIsInField(Vector2 point, float margin) {
...
+    // everytime we add a vector from the lowest goalside to the lowest part of the obstacle we remember the upper part of the obstacle
+    // That upper part is stored as the lowerhook again: and we can repeat the process
+    for (auto const &blockade : blockades) {
+        auto lowerbound = std::min(blockade.first.y, blockade.second.y);
+
+        // if the lowerbound is the same as the lower hook then the visible part has a length of 0 and we don't care about it
+        if (lowerbound != lowerHook.y) {
+            visibleParts.emplace_back(std::make_pair(lowerHook, Vector2(blockade.first.x, lowerbound)));
+        }
+        auto upperbound = std::max(blockade.first.y, blockade.second.y);
```
rolfvdhulst: same as before

roboteam_ai/src/utilities/Field.cpp
```diff
@@ -78,6 +81,153 @@ bool Field::pointIsInField(Vector2 point, float margin) {
...
+    auto lowerHook = lower;
+    std::vector<std::pair<Vector2, Vector2>> visibleParts = {};
+
+
+    // we start from the lowerhook, which is the lowest goal side at the start.
+    // The obstacles are sorted on their smallest value.
+    // everytime we add a vector from the lowest goalside to the lowest part of the obstacle we remember the upper part of the obstacle
+    // That upper part is stored as the lowerhook again: and we can repeat the process
+    for (auto const &blockade : blockades) {
+        auto lowerbound = std::min(blockade.first.y, blockade.second.y);
```
rolfvdhulst: I would use fmin()

roboteam_ai/src/utilities/Field.cpp
```diff
@@ -78,6 +81,153 @@ bool Field::pointIsInField(Vector2 point, float margin) {
...
+            auto point2 = util::twoLineIntersection(point, upperSideOfRobot, lowerGoalSide, upperGoalSide);
+
+            // remove all obstacles that are completely out of the goal
+            bool bothPointsBelowGoal = point1.y < lowerGoalSide.y && point2.y < lowerGoalSide.y;
+            bool bothPointAboveGoal = point1.y > upperGoalSide.y && point2.y > upperGoalSide.y;
+            if (!bothPointsBelowGoal && !bothPointAboveGoal) {
+
+                // constrain the blockades to within the goal
+                if (point1.y > point2.y) { // point1 is largest
+                    point1.y = std::min(point1.y, upperGoalSide.y);
```
rolfvdhulst: Same

PR #141, opened 2019-02-19
Feature/pause and Demo Joystick

Barisimre:
Adds the pause button that pauses

Add also the logic for embedded world demo joystick things


roboteam_ai/src/interface/InterfaceValues.cpp
```diff
@@ -111,6 +125,22 @@ void InterfaceValues::setUseRefereeCommands(bool useRefereeCommands){
     std::lock_guard<std::mutex> lock(RefMutex);
     InterfaceValues::useRefereeCommands = useRefereeCommands;
 }
+void InterfaceValues::setHaltText() {
```
mrlukasbos: this is not the place, see comment above.

roboteam_ai/src/interface/mainWindow.cpp
```diff
@@ -229,6 +229,11 @@ void MainWindow::updatePID_luth() {
 /// send a halt signal to stop all trees from executing
 void MainWindow::sendHaltSignal() {
     InterfaceValues::sendHaltCommand();
+    InterfaceValues::setHaltText();
+    haltBtn->setText(InterfaceValues::getHaltText());
```
mrlukasbos: you could enter the text and background handling in here?

PR #144, opened 2019-02-22
Demo logic

Barisimre:


roboteam_ai/src/interface/mainWindow.cpp
```diff
@@ -229,6 +230,10 @@ void MainWindow::updatePID_luth() {
 /// send a halt signal to stop all trees from executing
 void MainWindow::sendHaltSignal() {
     InterfaceValues::sendHaltCommand();
+    std::cout << "Pause" << std::endl;
```
mrlukasbos: This log can be removed

roboteam_ai/src/demo/JoystickDemo.cpp
```diff
@@ -19,14 +19,26 @@ bool JoystickDemo::isDemo() {
...
     }
+    else if (msg.halt == 1 && msg.id == -1) {
+        pause.setPause(true);
+        pause.haltRobots();
+    }
+    else if (msg.halt == 2 && msg.id == -1) {
+        pause.setPause(false);
+    }
+    else {
+        std::cout << "Pause Else should never be hit" << std::endl;
```
mrlukasbos: `std::cerr` since this looks like something we might wanna print in red

PR #146, opened 2019-02-26
Feature/new trees

Barisimre:
works tm


roboteam_ai/src/ApplicationManager.cpp
```diff
@@ -116,6 +116,7 @@ void ApplicationManager::updateDangerfinder() {
 void ApplicationManager::handleRefData() {
     ai::StrategyManager strategyManager;
     std::string strategyName = strategyManager.getCurrentStrategyName(refereeMsg.command);
+    BTFactory::getFactory().init();
```
mrlukasbos: I suppose this rebuilds the trees. it looks like you reinitialize the trees every tick

PR #152, opened 2019-03-05
Add hot swapping

Barisimre:
Add json hot swapping
No need to pause or restart


roboteam_ai/src/treeinterp/BTFactory.cpp
```diff
@@ -21,6 +21,8 @@ BTFactory &BTFactory::getFactory() {
 /// Initiate the BTFactory
 void BTFactory::init() {
 
+    std::cout << "init" << std::endl;
```
mrlukasbos: remove this, or make it more clear what we are initializing

PR #154, opened 2019-03-06
Fix segfaults in Conditions, change to onInitialize/onUpdate

ThijsLuttikhuis:
Fixed bunch of seg faults related to conditions

Refactor some names

Add general drawer in Drawer

Improve NumTreePosControl


roboteam_ai/src/conditions/Condition.cpp
```diff
@@ -6,5 +6,35 @@ namespace ai {
...
+std::string Condition::node_name() {
+    return name;
+}
+
+void Condition::initialize() {
+    if (! properties->getString("ROLE").empty()) {
+        robot = getRobotFromProperties(properties);
+        ball = World::getBall(); // update ball position
+        if (! robot) return;
+        if (! ball) return;
```
mrlukasbos: This allows all conditions only to run when the ball is there. is that what we want?

roboteam_ai/src/conditions/BallOutOfField.cpp
```diff
@@ -12,7 +12,7 @@ namespace ai {
 BallOutOfField::BallOutOfField(std::string name, bt::Blackboard::Ptr blackboard)
         :Condition(std::move(name), std::move(blackboard)) { };
 
-bt::Node::Status BallOutOfField::update() {
+bt::Node::Status BallOutOfField::onUpdate() {
     Vector2 ballPos;
     auto ball = World::getBall();
```
mrlukasbos: the ball object is already available here so don't need to get it

roboteam_ai/src/conditions/BallInDefenseAreaAndStill.cpp
```diff
@@ -13,14 +13,14 @@ namespace ai {
...
     if (properties->hasBool("theirDefenceArea")) {
         theirDefenceArea = properties->getBool("theirDefenceArea");
     }
     else {theirDefenceArea=false;}
 }
 
-bt::Node::Status BallInDefenseAreaAndStill::update() {
+bt::Node::Status BallInDefenseAreaAndStill::onUpdate() {
     Vector2 ballPos;
     auto ball=World::getBall();
```
mrlukasbos: the ball object is already available here so don't need to get it

roboteam_ai/src/conditions/BallKickedToOurGoal.cpp
```diff
@@ -12,7 +12,7 @@ namespace ai {
 BallKickedToOurGoal::BallKickedToOurGoal(std::string name, bt::Blackboard::Ptr blackboard)
         :Condition(std::move(name), std::move(blackboard)) { };
 
-bt::Node::Status BallKickedToOurGoal::update() {
+bt::Node::Status BallKickedToOurGoal::onUpdate() {
     auto ball = World::getBall();
```
mrlukasbos: the ball object is already available here so don't need to get it

roboteam_ai/src/conditions/HasBall.cpp
```diff
@@ -18,7 +18,7 @@ namespace ai {
 
 HasBall::HasBall(std::string name, bt::Blackboard::Ptr blackboard) : Condition(name, blackboard) { }
 
-bt::Node::Status HasBall::update() {
+bt::Node::Status HasBall::onUpdate() {
     robot = getRobotFromProperties(properties);
```
mrlukasbos: the robot object is already available here so don't need to get it

roboteam_ai/src/conditions/IsRobotClosestToBall.cpp
```diff
@@ -11,7 +11,7 @@ namespace ai{
 IsRobotClosestToBall::IsRobotClosestToBall(std::string name, bt::Blackboard::Ptr blackboard)
 : Condition(std::move(name), std::move(blackboard)) { }
 
-bt::Node::Status IsRobotClosestToBall::update() {
+bt::Node::Status IsRobotClosestToBall::onUpdate() {
     robot = getRobotFromProperties(properties);
```
mrlukasbos: the robot object is already available here so don't need to get it

roboteam_ai/src/skills/Skill.cpp
```diff
@@ -38,28 +28,31 @@ Skill::Status Skill::update() {
...
+    if (! ball) return Status::Waiting;
     return onUpdate();
 }
 
 void Skill::initialize() {
     robot = getRobotFromProperties(properties);
     ball = World::getBall();
-    if (!robot) return;
+    if (! robot) return;
+    if (! ball) return;
```
mrlukasbos: now skills don't work if there is no ball! this can be annoying

roboteam_ai/src/skills/Skill.h
```diff
@@ -14,29 +14,27 @@ namespace ai {
...
  * \brief Base class for all skills. Provides no additional functionality.
  */
 class Skill : public bt::Leaf {
     private:
         roboteam_msgs::RobotCommand rotateRobotCommand(roboteam_msgs::RobotCommand &cmd);
-protected:
-        io::IOManager ioManager = io::IOManager(false,true);
-
+    protected:
+        using Control = control::ControlUtils;
```
mrlukasbos: I would make it cutils instead of Control, which is confusing with the other namespace; 'control'

roboteam_ai/src/utilities/World.cpp
```diff
@@ -131,7 +131,13 @@ std::shared_ptr<roboteam_msgs::WorldRobot> World::getRobotClosestToPoint(const V
 /// returns the ball msg
 std::shared_ptr<roboteam_msgs::WorldBall> World::getBall() {
     std::lock_guard<std::mutex> lock(worldMutex);
-    return std::make_shared<roboteam_msgs::WorldBall>(world.ball);
+    //TODO:
```
mrlukasbos: What do you need to do?

roboteam_ai/test/ConditionTests/BallOutOfFieldtest.cpp
```diff
@@ -22,6 +22,7 @@ TEST(BallOutOfFieldTest, it_detects_ball_out_of_field) {
     worldMsg.ball.pos.x = 0.0;
     worldMsg.ball.pos.y = 0.0;
     worldMsg.ball.visible = 1;
+    worldMsg.ball.area = 99999;
```
mrlukasbos: what does the ball area do?

mrlukasbos (changes requested):
In the conditions and skill you now force the existence of a ball. I think that is questionable. some skills don't need a ball to operate properly; and for demo's it might get annoying. maybe a setting on skills/condition that sets if the existence of a ball is necessary? 

Secondly, the skills are not properly updated using the `ball` and `robot`  inherited objects which of course means shorter code! 

Finally, I see some changes with ball.area in the world messages. I am unsure why you set it and I am unsure why you set it so often to 99999.

roboteam_ai/src/conditions/BallInDefenseAreaAndStill.cpp
ThijsLuttikhuis: Yes and no,

Currently the implementation in Condition.h is: *if* there is a "ROLE" in properties, get the robot corresponding to that role, and get the ball. 
No ball or no robot? Then return;

If the Condition is used in a non-role scenario though, (which is possible for Conditions, not for Skills) you might want to get the ball in the condition itselves at 'onUpdate' if needed, as it is not done in 'update'.

roboteam_ai/src/conditions/Condition.cpp
ThijsLuttikhuis: I've tested it irl with ssl-vision and a ball:

There is always a ball, once there was one ball in the world, so shouldn't be a problem. 
In ssl-vision, as well as grsim, our world state has a ball area of 99999, so in the tests I put that at 99999 as well. Reason for the area in the first place, is that I needed a way to check if the ball is defined yet. I check if the area of the ball is not 0, because that means the ball is not defined yet

roboteam_ai/src/skills/Skill.cpp
ThijsLuttikhuis: More annoying if the skill segfaults when there is no ball ;)

roboteam_ai/src/skills/Skill.h
ThijsLuttikhuis: Copied from Skill.h ;-)

roboteam_ai/test/ConditionTests/BallOutOfFieldtest.cpp
ThijsLuttikhuis: Ball area is default 0, now I check if the ball has an area 'not 0', so all balls in world need an area now.

In grsim and ssl vision (or probably in vision/world) the ball always has area 99999, so I just make it that.

roboteam_ai/src/skills/Skill.cpp
mrlukasbos: You should then be able to say if a skilll can /cannot run without a ball.

Barisimre (dismissed):
get rekt

mrlukasbos (approved):

PR #199, opened 2019-03-19
restructure coach

mrlukasbos:
This PR restructures the coach into multiple files.

For defense and formations the coach functionality was not necessary and therefore removed and replaced by static functions within the skills themselves. 

I tried to avoid static classes in this PR as a matter of test. 

Defensive and offensive coach functionality will be added in later pull requests.


RobZelluf (approved):
Looks good!

PR #250, opened 2019-04-08
Feature/more pids interface

mrlukasbos:
for every gtp you can have a separate pid and update it in the interface now. I made a new abstraction for it called pidBox, which emits a QT signal upon change and then that change gets propagated towards the positioncontrollers


roboteam_ai/src/interface/mainWindow.cpp
```diff
@@ -104,7 +82,31 @@ MainWindow::MainWindow(QWidget* parent)
...
+    basicPidBox->setPid(InterfaceValues::getBasicPid());
+
+    QObject::connect(numTreePidBox, static_cast<void (PidBox::*)(pidVals)>(&PidBox::pidChanged),
+                     [=](const pidVals &pid) { InterfaceValues::setNumTreePid(pid); });
+
+    QObject::connect(forcePidBox, static_cast<void (PidBox::*)(pidVals)>(&PidBox::pidChanged),
+                     [=](const pidVals &pid) { InterfaceValues::setNumTreePid(pid); });
+
+    QObject::connect(basicPidBox, static_cast<void (PidBox::*)(pidVals)>(&PidBox::pidChanged),
+                     [=](const pidVals &pid) { InterfaceValues::setNumTreePid(pid); });
```
Barisimre: something something haskell

PR #251, opened 2019-04-08
Feature/acceleration limiter

simenbootsma:
- [x] The tests are passing

### Summary
Limit acceleration to have more stable and smooth movement of the robot. Acceleration can be higher in the forward direction of the robot as compared to the sideways direction because of the orientation of the wheels. I used a linear mapping of the angle between the orientation and the direction of driving to implement this. 

Added the accelerationlimiter to positioncontroller PID function, so all functions using poscontrol should automatically limit it now.


roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -232,6 +231,25 @@ Vector2 ControlUtils::velocityLimiter(Vector2 vel, double maxVel,double minVel){
 }
 
 
+/// Limits acceleration to maximum acceleration
+Vector2 ControlUtils::accelerationLimiter(Vector2 vel, double maxAcc,double prevVel){
+    if (vel.length() > (prevVel + maxAcc/Constants::TICK_RATE())) {
```
rolfvdhulst: we should probably prevent writing division by tickRate since we can measure the time intervals now between the last worldstates

rolfvdhulst: so you could just do maxAcc*time interval

roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -232,6 +231,25 @@ Vector2 ControlUtils::velocityLimiter(Vector2 vel, double maxVel,double minVel){
...
+    }
+    return vel;
+}
+
+
+/// Calculate the maximum acceleration based on the direction of driving.
+/// Acceleration is the lowest in the sideways direction and highest in the forward direction.
+double ControlUtils::calculateMaxAcceleration(Vector2 vel, double angle){
+    double angleDiff = ControlUtils::constrainAngle(vel.angle() - angle);
+    angleDiff = (abs(angleDiff) > 0.5*M_PI) ? (M_PI - abs(angleDiff)) : abs(angleDiff); // angle from vel.angle() to local x-axis of the robot
```
rolfvdhulst: I would kind of like to see this thing (determining angle relative to robots local angle) as a seperate function as we will use it in other places too.

roboteam_ai/test/ControlTests/ControlUtilsTest.cpp
```diff
@@ -45,6 +45,37 @@ TEST(ControlUtils, velocityLimiter) {
     }
 }
 
+TEST(ControlUtils, accelerationLimiter) {
+    rtt::Vector2 testVel;
+    double prevTestVel;
+    for (int i = 1; i < 200; i ++) {
+        testVel = Vector2(5*cos(i/100.0*M_PI), 3*sin(i/30.0*M_PI));
+        prevTestVel = Vector2(5*cos((i-1)/100.0*M_PI), 3*sin((i-1)/30.0*M_PI)).length();
+        EXPECT_LE(cr::ControlUtils::accelerationLimiter(testVel, rtt::ai::Constants::MAX_ACC_UPPER(), prevTestVel).length(),
```
rolfvdhulst: use using statements so  something like using c=rtt::ai::constants so this is actually kind of readable.

rolfvdhulst: this goes for all the tests

roboteam_ai/src/skills/GoToPos.cpp
```diff
@@ -35,7 +35,11 @@ bt::Node::Status GoToPos::onUpdate() {
     }
 
     control::PosVelAngle pva = gotopos.getPosVelAngle(robot, targetPos);
-    pva.vel = control::ControlUtils::velocityLimiter(pva.vel, maxVel);
+    pva.vel = control::ControlUtils::velocityLimiter(pva.vel, maxVel, minVel);
```
rolfvdhulst: I think it would be nice if all these control calls/ad hoc changes to commands can be in one function call that we can easily implement in skill or such. For now this is fine but if we want to change it structurally this needs to be changed structurally

roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -258,6 +257,36 @@ Vector2 ControlUtils::velocityLimiter(const Vector2 &vel, double maxVel, double
...
+    if (vel.length() > (prevVel + maxAcc/Constants::TICK_RATE())) {
+        return vel.stretchToLength(prevVel + maxAcc/Constants::TICK_RATE());
+    }
+    return vel;
+}
+
+/// Calculate the maximum acceleration based on the direction of driving.
+/// Acceleration is the lowest in the sideways direction and highest in the forward direction.
+//double ControlUtils::calculateMaxAcceleration(const Vector2 &vel, double angle){
+//    Angle angleDiff = vel.toAngle() - angle;
```
mrlukasbos: commented code should be removed

roboteam_ai/src/jsons/keeper/Keeper.json
```diff
@@ -1 +0,0 @@
...: {"children": ["442wnhvzcos5l6fgc", "l3f41xsptc9v4mkp1"], "id": "3r3vdkvrq0qy2k8tq", "title": "Sequence"}, "h2xe4k2qeyxz98": {"children": ["vgze2gu8e6uovxjw9", "3r3vdkvrq0qy2k8tq"], "id": "h2xe4k2qeyxz98", "title": "MemSequence"}, "v97vjfarlmuf9bf5": {"id": "v97vjfarlmuf9bf5", "title": "BallKickedToOurGoal"}, "5f53g1cyfsq6s": {"properties": {"Keeper": "true"}, "id": "5f53g1cyfsq6s", "title": "InterceptBall"}, "gsh09zbfjy9lj1558": {"id": "gsh09zbfjy9lj1558", "title": "Kick"}, "a2c0bwakz850sit": {"properties": {"maxTime": "5.0"}, "id": "a2c0bwakz850sit", "title": "GetBall"}, "vgze2gu8e6uovxjw9": {"properties": {"theirDefenceArea": "false"}, "id": "vgze2gu8e6uovxjw9", "title": "BallInDefenseAreaAndStill"}, "upe8dim19wwr9o5ts": {"properties": {"Position": "{-1.6,0}", "distanceBehindBall": "0.01", "goBehindBall": "true", "maxVel": "0.8"}, "id": "upe8dim19wwr9o5ts", "title": "GoToPos"}, "phgpzxou1ceze6ua": {"id": "phgpzxou1ceze6ua", "title": "HasBall"}, "l3f41xsptc9v4mkp1": {"children": ["upe8dim19wwr9o5ts", "a2c0bwakz850sit", "exhsxbmfw5lchsb81", "gsh09zbfjy9lj1558"], "id": "l3f41xsptc9v4mkp1", "title": "MemSequence"}}, "root": "o3jcbqg12iz4ba9", "title": "Keeper"}]}, "name": "Keeper"}
```
mrlukasbos: why is this file removed?

roboteam_ai/src/utilities/Constants.h
```diff
@@ -34,6 +34,9 @@ class Constants {
     static double MIN_ANGLE()                   { return -M_PI; };
     static double MAX_ANGLE()                   { return M_PI; };
     static double MAX_VEL()                     { return 8.0; };
+    static double MIN_VEL()                     { return 0.2; };  // Minimum velocity to make the robot move
```
mrlukasbos: this looks like it's not used

roboteam_ai/src/jsons/keeper/Keeper.json
ThijsLuttikhuis: no idea :x I'll put it back in

roboteam_ai/src/control/ControlUtils.cpp
ThijsLuttikhuis: yes

roboteam_ai/src/utilities/Constants.h
ThijsLuttikhuis: useful if you want a minimum velocity

PR #253, opened 2019-04-08
Feature/canshoottotarget

mrlukasbos:
This is probably useful for passing and detecting passes within trees


rolfvdhulst (commented):
Looks fine, but how does this differ from HasClearshot? Both even call the same clearLine function. Suggest maybe merging those

mrlukasbos:
Has clearshot specifically only looks at their goal...

rolfvdhulst (commented):
Yes, so it would be identical to setting target to theirGoal in your skill. Feels kind of like redundant c++ to recompile if we change constants to me.

mrlukasbos:
yes. that other skill can probably be removed and this one should be extended. I prefer doing that later though

roboteam_ai/src/conditions/CanShootToTarget.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace rtt {
+namespace ai {
+
+CanShootToTarget::CanShootToTarget(std::string name, bt::Blackboard::Ptr blackboard)
+        :Condition(std::move(name), std::move(blackboard)) { };
+
+bt::Node::Status CanShootToTarget::onUpdate() {
+
+    // get the target position to which we want a clear line
+    if (properties->hasVector2("target")) {
```
ThijsLuttikhuis: how do you get the target? is it a hardcoded vector2?

roboteam_ai/src/conditions/CanShootToTarget.cpp
```diff
@@ -0,0 +1,35 @@
...
+		target = properties->getVector2("target");
+	} else {
+		return bt::Node::Status::Failure;
+	}
+
+    // get the margin, which is the min distance from the line
+	if (properties->hasDouble("margin")) {
+		margin = properties->getDouble("margin");
```
ThijsLuttikhuis: in passing, the margin is an angle from the shooting robot in stead of a distance to the passline,
maybe change this to the same definition of can shoot to target

roboteam_ai/src/conditions/CanShootToTarget.cpp
```diff
@@ -0,0 +1,35 @@
...
+	} else {
+		margin = MAX_DIST_FROM_LINE;
+	}
+
+	bool canShoot = control::ControlUtils::clearLine(robot->pos, target, world::world->getWorld(), margin, true);
```
ThijsLuttikhuis: clearLine is a vague name for a function, on it's own it would mean that you 'clear' the line i.e. delete/reset it.

ThijsLuttikhuis (changes requested):
currently not usable without hardcoding the position of the target Vector2

mrlukasbos:
I guess this is getting outdated.

PR #257, opened 2019-04-09
Fix Receive skill

RobZelluf:
Restructure receive so that it also works when ballPlacement is not done (it will just aim to receive the ball at the position the robot is already standing).


roboteam_ai/src/skills/Reflect.h
```diff
@@ -0,0 +1,25 @@
+//
```
mrlukasbos: this should not be in this PR

roboteam_ai/src/skills/Reflect.cpp
```diff
@@ -0,0 +1,5 @@
+//
```
mrlukasbos: remove from this PR

roboteam_ai/src/skills/Receive.h
```diff
@@ -46,15 +46,15 @@ class Receive : public Skill {
     bool readyToReceivePass = false;
 
 
-    bool isInPosition(Vector2 behindTargetPos);
+    bool isInPosition(const Vector2& behindTargetPos = {0, 0});
```
mrlukasbos: What! no! this is bad. the ballplacement target is not always {0,0}. it also mistyfies the behaviour of isInPosition

roboteam_ai/src/skills/Receive.cpp
```diff
@@ -22,10 +22,29 @@ Receive::Status Receive::onUpdate() {
...
+                ball->pos);
+
+        moveToCatchPosition(behindTargetPos);
+
+        if (isInPosition(behindTargetPos)) {
+            coach::g_pass.setReadyToReceivePass(true);
+        }
+    } else {
+        // Check if robot is in position, otherwise turn towards ball
+        if (isInPosition()) {
```
mrlukasbos: so i'm guessing this still always receives at {0,0}

roboteam_ai/src/skills/Receive.cpp
```diff
@@ -22,10 +22,29 @@ Receive::Status Receive::onUpdate() {
...
+            coach::g_pass.setReadyToReceivePass(true);
+        } else {
+            command.w = static_cast<float>((Vector2(ball->pos) - robot->pos).angle());
+            publishRobotCommand();
```
mrlukasbos: this function could be shortened

mrlukasbos (changes requested):
There are weird things going on

mrlukasbos (approved):
I tested it locally and apparently there is a ballplacement check in inPosition. So for now it's OK.

roboteam_ai/src/skills/Receive.cpp
mrlukasbos: nvm, this had a check

roboteam_ai/src/skills/Receive.h
mrlukasbos: nvm, this had a check

PR #258, opened 2019-04-10
Feature/keeper tree switching

### LINKED ISSUE #233: Keeper referee communication

mrlukasbos:
The keeper also can listen to ref commands now. 

- It **defaults to 'singlekeeper strategy'.** 
- for ballblacement (both us and them) it becomes an **avoider**
- letting the keeper do ballplacement (or even the pass) might not be desirable because you want the keeper to be in the goal as much as possible. What if the referee lets kick off when the keeper is not yet back after ballplacement? Also, that would require a significant amount of extra work.

fixes #233


roboteam_ai/src/skills/Skill.cpp
```diff
@@ -15,9 +15,10 @@ void Skill::publishRobotCommand() {
     std::string ourSideParam;
     nh.getParam("our_side", ourSideParam);
 
-//    if(ourSideParam=="right"){
-//        cmd=rotateRobotCommand(cmd);
-//    }
+    if(ourSideParam=="right"){
+        command=rotateRobotCommand(command);
+    }
```
ThijsLuttikhuis: fix this with grsim/real-life

mrlukasbos: done

PR #265, opened 2019-04-11
Bugfix/fix major issues

mrlukasbos:
- fix a segfault on boot (tree was terminated while it was non-existent)
- show robothub output when running roboteam ai two-teams.launch
- when a node is never ticked but gets terminated it sometimes sends a command with vel = 0, but the robot_id is -1 so i added a check for that as well in iomanager


roboteam_ai/test/UtilTests/IOTest.cpp
```diff
@@ -63,14 +63,26 @@ TEST(IOTest, it_subscribes) {
     ros::Subscriber robotCommandSub = nh.subscribe<roboteam_msgs::RobotCommand>(rtt::TOPIC_COMMANDS, 0,
             &robotCommandCallback);
 
+    // without a valid ID no command should be sent
     roboteam_msgs::RobotCommand cmd;
+    cmd.id = -1;
     cmd.x_vel = 10;
```
ThijsLuttikhuis: use x_vel and y_vel lower than the max_velocity(length < 8 or something)  for a command, there might be some check (maybe later in time, as extra check for robocup) somewhere that will make it fail in another way than tested here

PR #267, opened 2019-04-11
Interface upgrades

mrlukasbos:
- add a switch side button
- add a switch INTERFACE side button checkbox. this is useful for playing two teams on the same pc: you can make viewing much easier then. it is standard set to false because otherwise things may get obscure during debugging.
- it adds some visual hints in the interface which side is where


roboteam_ai/src/interface/mainWindow.cpp
```diff
@@ -223,6 +243,17 @@ void MainWindow::toggleOurColorParam() {
     setToggleColorBtnLayout();
 }
 
+/// toggle the ROS param 'our_color'
+    void MainWindow::toggleOurSideParam() {
+        ros::NodeHandle nh;
+        std::string ourColorParam, newParam;
+        nh.getParam("our_side", ourColorParam);
+        newParam = ourColorParam == "left" ? "right" : "left";
```
ThijsLuttikhuis: dude wat?

roboteam_ai/src/interface/widget.cpp
```diff
@@ -83,6 +84,48 @@ void Visualizer::drawFieldLines(QPainter &painter) {
...
+    ourLineLower = toScreenPosition(ourLineLower);
+
+
+    auto color = weAreYellow ? QColor(255,255,0,255) : QColor(80,80,255,255);
+    pen.setBrush(color);
+    pen.setColor(color);
+    painter.setPen(pen);
+    painter.drawLine(ourLineUpper.x, ourLineUpper.y, ourLineLower.x, ourLineLower.y);
+
+    ///////////
```
ThijsLuttikhuis: blank

PR #271, opened 2019-04-11
Feature/receive to kick

RobZelluf:
Add the Reflect skill to try to kick the ball immediately at the goal when it is received


roboteam_ai/src/coach/CoachHeuristics.cpp
```diff
@@ -1,3 +1,5 @@
+#include <utility>
```
rolfvdhulst: You don't need this one (alt entering makes clion go stupid sometimes)

roboteam_ai/src/utilities/Constants.h
```diff
@@ -49,12 +49,13 @@ class Constants {
     static int TICK_RATE()                      { return 60; };
     static double CLOSE_TO_BORDER_DISTANCE()    { return 1.2 * ROBOT_RADIUS(); };
     static int GAME_ANALYSIS_TICK_RATE()        { return 30; };
+    static double DISTANCE_TO_KICKER()          { return 0.06; };
```
rolfvdhulst: This is almost exactly the same as CentreToFront() (which is ~0.059) which is the exact distance to the front of the robot. I suggest we don't make duplicate constants for the same concept (so either change this one or use the other one that does the proper measurement).

rolfvdhulst: Essentially this is just a duplicate.

roboteam_ai/src/skills/ReflectKick.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+void ReflectKick::onInitialize() {
+    auto field = world::field->get_field();
+    goalTarget = getFarSideOfGoal();
+    reflectionPos = robot->pos;
+    numTreeGtp.setAvoidBall(false);
+}
+
+ReflectKick::Status ReflectKick::onUpdate() {
+    angleToGoalTarget = (goalTarget - getKicker()).toAngle();
```
rolfvdhulst: This can be split in more functions and be easier to read. It's hard to follow what you're doing.

rolfvdhulst: It's hard to follow where you are doing the mathematics that is unique (e.g. the angles that are different from normal intercept), maybe make a small comment or put them in one function

rolfvdhulst (changes requested):
Looks ok, your skill is hard to follow at times, some small things need to be changed (See comments)

rolfvdhulst (approved):

PR #279, opened 2019-04-14
Refactor/clean conditions

mrlukasbos:
I got bored and decided to clean things up such that everything looks the same and is documented. 
This PR does that for all conditions.


roboteam_ai/src/conditions/BallKickedToOurGoal.h
```diff
@@ -11,15 +7,15 @@ namespace rtt {
...
 
 class BallKickedToOurGoal : public Condition {
-        const double BALL_TO_GOAL_MARGIN = 0.0215;
-        const double BALL_TO_GOAL_TIME = 2.5;
-    public:
-        explicit BallKickedToOurGoal(std::string name = "BallKickedToOurGoal", bt::Blackboard::Ptr blackboard = nullptr);
-        Status onUpdate() override;
-        std::string node_name() override { return "BallKickedToOurGoal"; };
+private:
+    const double BALL_TO_GOAL_MARGIN = 0.0215; // m
```
rolfvdhulst: Should probably just make this constants ball radius explicitly

roboteam_ai/test/ConditionTests/BallInDefenseAreaAndStillTest.cpp
```diff
@@ -19,7 +19,7 @@ TEST(DetectsDefenseArea, BallInDefenseAreaAndStill)
     EXPECT_TRUE(nodeTheirDefenceArea.theirDefenceArea); // check if the property is handled properly
 
     BBpointer->setBool("theirDefenceArea", false);
-    rtt::ai::BallInDefenseAreaAndStill node("Test", BBpointer);
+    rtt::ai::BallInDefenseAreaAndStill node("BallInDefenseAreaAndStill", BBpointer);
```
rolfvdhulst: I personally think for tests we should make sure the nodes have an explicitly different name so there is no chance in hell they can ever get confused. (that's why it's different I think)

PR #298, opened 2019-04-18
Improve passing

RobZelluf:
Improve passing logic such that passing and receiving always terminates when it either takes too long or fails. 

Currently receiver speed is doubled, but this should probably be fixed in goToPos later on.


roboteam_ai/src/coach/PassCoach.cpp
```diff
@@ -15,9 +15,15 @@ void PassCoach::resetPass() {
     passed = false;
     readyToReceivePass = false;
     robotBeingPassedTo  = -1;
+    timerStarted = false;
 }
 
 int PassCoach::initiatePass(int passerID) {
+    if (robotBeingPassedTo != -1) {
+        if(!passTakesTooLong()) {
```
ThijsLuttikhuis: unclear what this does, make a comment or something

roboteam_ai/src/coach/heuristics/CoachHeuristics.cpp
```diff
@@ -62,10 +62,10 @@ double CoachHeuristics::calculateDistanceToOpponentsScore(const Vector2 &positio
 
 double CoachHeuristics::calculateBehindBallScore(const Vector2 &position, CoachHeuristics::WorldData world) {
     double xDistanceBehindBall = world.ball.pos.x - position.x;
-    if (xDistanceBehindBall < 0) {
-        return 1.0;
+    if (xDistanceBehindBall > 0) {
+        return 0.0;
```
ThijsLuttikhuis: why is this not a bool?

ThijsLuttikhuis (changes requested):
please revert the json changes

ThijsLuttikhuis (approved):

PR #307, opened 2019-04-23
Line class tests

rolfvdhulst:
First merge the line class in utils before you look at this. 
Adds tests for Line class. Done like this because apparently we are never running the tests in roboteam_utils


mrlukasbos (changes requested):
I do not think this is a proper approach. Running this test to test a non-that-much-changing codebase is not a good solution.  This test should obviously be in Roboteam Utils. 

As a solution, I propose that I will enable Travis builds for RTT utils to make sure the tests work.

mrlukasbos (approved):

mrlukasbos:
approved after short discussion. Conclusion: these will be moved to utils later.

PR #308, opened 2019-04-23
remove robotdealer logic

mrlukasbos:
some early monday refactor


roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -12,21 +12,6 @@ namespace rtt {
 namespace ai {
 namespace control {
 
-/// return the angular velocity for a robot to go from robotAngle to targetAngle
```
rolfvdhulst: I would argue against removing this function as it was an easy way for us to test angular velocity control instead of global control if ever needed. It´s not like there is many functions like this in our system.

PR #312, opened 2019-04-23
fix hungarian

mrlukasbos:
The hungarian had a bug that places where switched upon refresh. apparently it sorted the robots on ID, which was quite awesome, but not the idea. 

The problem was that the distance matrix rounded distances to integers, which means whole meters. Then distances that where the same where sorted on ID instead because other params wouldn't matter anyway.


rolfvdhulst (approved):
We should have a tally for every time we get an integer division or cast bug.

ThijsLuttikhuis (approved):
bjutivol

PR #315, opened 2019-04-24
Fix terminating and unequal use of hungarian

RobZelluf:
Terminating is only done when the node was initialized


roboteam_ai/src/bt/Node.cpp
```diff
@@ -73,10 +74,15 @@ Node::Status Node::NodeUpdate() {
 
 void Node::NodeInitialize() {
     initialize();
+    init = true;
```
ThijsLuttikhuis: um wat? y u need dis

roboteam_ai/src/coach/OffensiveCoach.h
```diff
@@ -35,21 +35,19 @@ class OffensiveCoach {
...
-
     std::vector<Vector2> getDefaultLocations();
-    std::vector<Vector2> getNewOffensivePositions(int numberOfRobots);
-
-    const set<RobotPtr> &getSideAttackers() const;
+    void updateOffensivePositions();
+    std::vector<Vector2> getOffensivePositions(int numberOfRobots);
 
-    void addSideAttacker(RobotPtr);
+    void addSideAttacker(const RobotPtr&);
```
ThijsLuttikhuis: add name of thing

ThijsLuttikhuis (dismissed):
change plox

roboteam_ai/src/coach/OffensiveCoach.cpp
```diff
@@ -92,40 +92,55 @@ std::vector<Vector2> OffensiveCoach::getNewOffensivePositions(int numberOfRobots
...
 }
 
-void OffensiveCoach::removeSideAttacker(const OffensiveCoach::RobotPtr& robot) {
-    int size = sideAttackers.size();
-    for (auto &sideAttacker : sideAttackers) {
-        if(sideAttacker->id == robot->id) {
-            sideAttackers.erase(sideAttacker);
-        }
+std::vector<Vector2> OffensiveCoach::getOffensivePositions(int numberOfRobots) {
+    numberOfRobots += numberOfRobots % 2;
```
rolfvdhulst: ehh wtf is this black magic. This needs a comment at least

roboteam_ai/src/skills/SideAttacker.h
```diff
@@ -19,7 +19,7 @@ class SideAttacker : public Skill {
     Vector2 deltaPos;
     Vector2 targetPos;
 
-    int zone = -1;
+    int zone;
```
rolfvdhulst: looks like a testing artifact to me that you removed this, was there a good reason?

rolfvdhulst (changes requested):
Looks okay, see comments

roboteam_ai/src/coach/OffensiveCoach.cpp
RobZelluf: done

roboteam_ai/src/skills/SideAttacker.h
RobZelluf: Removed it, all this logic is moved to the OffensiveCoach

rolfvdhulst (approved):

PR #318, opened 2019-04-25
shotcontroller

mrlukasbos:
This PR creates a general implementation for shooting.

Features:
- [x] Go to the ball under the right angle
- [x] Shoot at different speeds
- [x] Shoot at different precision
- [x] Shoot with or without geneva (the shotcontroller can decide for you)
- [ ] ~Force a geneva shot (will force to use geneva=1 or geneva=5~ lets implement this later
- [x] implement it in skills
- [ ] Fix attack rules


roboteam_ai/src/skills/Attack.cpp
```diff
@@ -40,30 +39,6 @@ bt::Node::Status Attack::onUpdate() {
     command.kicker_vel = shotData.kickSpeed;
     command.geneva_state = shotData.genevaState;
 
-
```
mrlukasbos: This needs to be re - implemented somehow. (maybe just with 'canmoveInDefenseArea' etc from GTP

PR #321, opened 2019-04-25
Feature/avoid ball on pass

RobZelluf:
Add a condition that checks whether the robot is on the pass line, and add the type "passing" to avoidBall to let it avoid the passLine if needed


roboteam_ai/src/skills/AvoidBall.cpp
```diff
@@ -15,8 +15,16 @@ using cu = control::ControlUtils;
 
 AvoidBall::AvoidBall(std::string name, bt::Blackboard::Ptr blackboard)
 : Skill(std::move(name), std::move(blackboard)) {
+}
+
+void AvoidBall::onInitialize() {
+    minRobotDistanceForForce = 0.7;
```
mrlukasbos: this variable should probably be somewhere else. either in header or in constants (this value is necessary to comply with the rules!!)

PR #331, opened 2019-04-26
make sure ros is properly used in world

mrlukasbos:
While working on referee fixes I noticed that there was a lot of unused code in worldmanager.
Basically that whole thing was unnecessary, since it was checking if a message changed while ROS just notifies you on that. 

tested the performance with htop and it is about equal.


ThijsLuttikhuis (approved):
yes

PR #337, opened 2019-04-26
fixPenaltyKeeper

rolfvdhulst:
Small nasty bug and some finetuning


roboteam_ai/src/skills/PenaltyKeeper.h
```diff
@@ -24,6 +24,9 @@ class PenaltyKeeper : public Skill  {
         control::BasicPosControl gtp;
         PenaltyState updateState(PenaltyState currentState);
         bool isBallShot();
+        //testing:
```
mrlukasbos: meh. remove these.

mrlukasbos (changes requested):

roboteam_ai/src/skills/PenaltyKeeper.h
rolfvdhulst: I will 100% need these when we have real robots. Don't see the point of removing them to add them again.

rolfvdhulst: removed them but this results in a lot of commented code, pick your poison.

mrlukasbos (approved):

PR #359, opened 2019-04-28
Create check for passing takes too long in main loop

RobZelluf:


mrlukasbos (approved):
for now fine. for later I think its nice to build a pass class which keeps track if these timers etc.

PR #363, opened 2019-04-29
vary max vel and sometimes fix velocities that were too low

mrlukasbos:


roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -205,6 +205,7 @@ int ControlUtils::rotateDirection(double currentAngle, double targetAngle) {
 
 /// Limits velocity to maximum velocity
 Vector2 ControlUtils::velocityLimiter(const Vector2 &vel, double maxVel, double minVel) {
```
rolfvdhulst: You should really add a comment here explaining that it automatically is defaulting to the maxvel as called by the referee

rolfvdhulst (approved):
See the one comment.

PR #365, opened 2019-04-29
Feature/aim util

rolfvdhulst:
Adds a small function that aims for you. It tries to aim to the corners or otherwise just shoots in the middle of the goal (uses goalVision). It is not completely ideal under really small angles, but for general attacking this will work fine.


roboteam_ai/src/coach/OffensiveCoach.cpp
```diff
@@ -145,17 +146,55 @@ std::vector<Vector2> OffensiveCoach::getOffensivePositions(int numberOfRobots) {
...
-    for(int i=0; i < numberOfPositions; i++) {
+    for (int i = 0; i < numberOfPositions; i ++) {
         positionVectors.emplace_back(offensivePositions[i].position);
     }
 
     return positionVectors;
 }
 
+/// this function decides what point in the goal to aim at from a position on which the ball will be/where the robot is
+Vector2 OffensiveCoach::getShootPoint(const Vector2 &fromPoint) {
```
mrlukasbos: the name should indicate that its about a goal

mrlukasbos (approved):
nice

PR #368, opened 2019-04-29
Feature/robots without geneva

mrlukasbos:
allow setting in the constants if robots have non-working geneva's.


roboteam_ai/src/utilities/Constants.h
```diff
@@ -106,6 +106,32 @@ class Constants {
     static bool STD_SHOW_AVAILABLE_PASSES()     { return false; };
     static bool STD_TIMEOUT_TO_TOP()            { return false; };
 
+    static std::vector<bool> ROBOTS_WITH_WORKING_GENEVA() {
```
rolfvdhulst: Prefer to see this as a map personally as you want unique ID's

rolfvdhulst: Like using ID's as element access is not as pretty as using ID's as key.

PR #371, opened 2019-04-29
catch ball on side of kicker + don't allow changing geneva while it still turns

mrlukasbos:
This should improve the real-life performance of shotControl.


roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -188,6 +194,20 @@ void ShotController::makeCommand(ShotData data, roboteam_msgs::RobotCommand &com
...
+Vector2 ShotController::getGenevaLineOffsetPoint(Vector2 point, int genevaState) {
+    if (genevaState == 3) {
+        return point;
+    }
+    std::map<int, double> genevaLineOffset;
+    genevaLineOffset[1] = 0.02;
+    genevaLineOffset[2] = 0.01;
+    genevaLineOffset[4] = -0.01;
+    genevaLineOffset[5] = -0.02;
+    point = point + point.rotate(M_PI/2).stretchToLength(genevaLineOffset[genevaState]);
```
Barisimre: Use M_PI_2

PR #392, opened 2019-04-30
shotcontroller improvements

mrlukasbos:
- [x] don't get stuck in Grsim (the robots there have another frontside shape so they need different measurements) 
- [x] fix bug where attackers not instantly kick.

### DIFF
--- a/roboteam_ai/src/control/shotControllers/ShotController.cpp
+++ b/roboteam_ai/src/control/shotControllers/ShotController.cpp
@@ -36,7 +36,18 @@ ShotData ShotController::getShotData(world::Robot robot, Vector2 shotTarget) {
    ShotData shotData;
    if (isOnLineToBall && isBehindBall) {
        bool hasBall = world::world->ourRobotHasBall(robot.id, Constants::MAX_KICK_RANGE());
-       shotData = hasBall && !genevaIsTurning ? shoot(robot, shotTarget) : moveStraightToBall(robot, currentDesiredGeneva);
+
+
+       if (hasBall && genevaIsTurning) {
+
+           std::cout << "I WANT TO SHOOT BUT THE GENEVA WONT LET ME. it turns for " << secondsToTurnGeneva << "s" << std::endl;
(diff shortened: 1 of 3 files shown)


roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -218,10 +228,10 @@ Vector2 ShotController::getGenevaLineOffsetPoint(Vector2 point, int genevaState)
         return point;
     }
     std::map<int, double> genevaLineOffset;
-    genevaLineOffset[1] = 0.02;
```
ThijsLuttikhuis: has this been tested in real life?

roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -61,10 +71,10 @@ void ShotController::determineGenevaAndPosition(const world::Robot &robot, const
         if (genevaDifference != 0) {
             genevaIsTurning = true;
             // each turn should increase the time which the geneva is turning
-            secondsToTurnGeneva = genevaDifference * 0.5;
+            secondsToTurnGeneva = genevaDifference * 0.3;
```
ThijsLuttikhuis: this might depend on real life / grSim ( and == magic number )

roboteam_ai/src/control/shotControllers/ShotController.cpp
mrlukasbos: not yet, but obviously i will.

roboteam_ai/src/control/shotControllers/ShotController.cpp
mrlukasbos: yes. Have to calculate for that.

mrlukasbos: will make an issue (it needs to become a function, since switching multiple states is faster per state than only one state, since the geneva has a pid)

PR #394, opened 2019-04-30
Fix starting before ball in play

Barisimre:


roboteam_ai/src/conditions/BallIsMoving.cpp
```diff
@@ -1,6 +1,7 @@
 
 /*
  * returns SUCCESS if the ball is moving
+ * USED FOR CHECKING GAME STATE CHANGES DONT USE IF YOU DONT KNOW WHEN TO
```
ThijsLuttikhuis: make it clearer from the title of the condition or w/e

roboteam_ai/src/conditions/BallIsMoving.cpp
```diff
@@ -17,8 +18,9 @@ bt::Node::Status BallIsMoving::onUpdate() {
     Vector2 ballVel=ball->vel;
 
     bool ballIsLayingStill = ballVel.length() < Constants::BALL_STILL_VEL();
+    auto refCommand = static_cast<RefGameState>(rtt::ai::Referee::getRefereeData().command.command);
 
-    if (ballIsLayingStill){
+    if (ballIsLayingStill || refCommand != RefGameState::NORMAL_START){
```
ThijsLuttikhuis: ugly as fuck. no.

roboteam_ai/src/conditions/BallIsMoving.cpp
ThijsLuttikhuis: make the name of the condition different

PR #400, opened 2019-05-01
add chip in shotcontroller

mrlukasbos:


roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -172,17 +174,25 @@ ShotData ShotController::moveStraightToBall(world::Robot robot, int genevaState)
...
     ShotData shotData(pva);
 
     // set the kicker and kickforce
-    shotData.kick = true;
-    shotData.kickSpeed = determineKickForce(ball->pos.dist(shotTarget));
+    if (chip) {
+        shotData.chip = true;
+        shotData.kick = false;
 
+        // TODO calibrate chip speed
```
rolfvdhulst: Make this an issue/add it to the testing list.

PR #403, opened 2019-05-01
Feature/chip pass

Barisimre:
Needs  #400 

- [x] Chip after 2 failed passes
- [x] Remove receiver changing position on terminate


roboteam_ai/src/conditions/HasClearShot.h
```diff
@@ -9,7 +9,7 @@ namespace ai {
 
 class HasClearShot : public Condition {
 private:
-    const double MIN_VIEW_AT_GOAL = 0.4;
+    const double MIN_VIEW_AT_GOAL = 0.25;
```
rolfvdhulst: How is this relevant to chip pass?

PR #404, opened 2019-05-01
Robots stay in field

Barisimre:


roboteam_ai/src/skills/gotopos/GTPSpecial.cpp
```diff
@@ -15,39 +15,43 @@ void GTPSpecial::gtpInitialize() {
...
+            Vector2 ballPos = rtt::ai::world::world->getBall()->pos;
+
+            Vector2 penaltyThem = rtt::ai::world::field->getPenaltyPoint(false);
+            targetPos = (ballPos + (penaltyThem - ballPos).stretchToLength((penaltyThem - ballPos).length()/2.0));
+            errorMargin = 0.05;
+            break;
+        }
+        case getBackIn: {
+            targetPos = {robot->pos.x, 0};
+            break;
```
ThijsLuttikhuis: what if robot->pos.x is outside field?

PR #405, opened 2019-05-01
Fix empty commands and a nasty positioning bug

rolfvdhulst:
Fixes 2 bugs 
* blocking too far forwards sometimes
* defenders not receiving commands


roboteam_ai/src/skills/CoachDefend.cpp
```diff
@@ -20,6 +20,10 @@ bt::Node::Status CoachDefend::onUpdate() {
     if (! targetLocation) {
 //        std::cerr << "Could not find the location of defender " << robot->id << " in calculated positions!"
 //                  << std::endl;
+        command.x_vel = 0;
```
ThijsLuttikhuis: looks good, after practice tournament this might be improved to finding a 'default good position' or something

ThijsLuttikhuis (approved):
primaaa

roboteam_ai/src/skills/CoachDefend.cpp
rolfvdhulst: Yes.

Barisimre (approved):

PR #421, opened 2019-05-02
Hotfix/fix terminates

rolfvdhulst:
cleans up a ton of terminates. Should not break stuff but prevent errors in general.


roboteam_ai/src/skills/Attack.cpp
```diff
@@ -46,6 +46,8 @@ bt::Node::Status Attack::onUpdate() {
 }
 
 void Attack::onTerminate(Status s) {
+    shot=false;
+    shotControl = std::make_shared<control::ShotController>(control::ShotPrecision::LOW ,control::BallSpeed::MAX_SPEED,false);
```
mrlukasbos: shot is deprecated and the shotcontroller is set on initialize so this terminate is not necessary

roboteam_ai/src/skills/AvoidBall.cpp
```diff
@@ -20,7 +20,7 @@ AvoidBall::AvoidBall(std::string name, bt::Blackboard::Ptr blackboard)
 void AvoidBall::onInitialize() {
     minRobotDistanceForForce = 0.7;
     stop = properties->getBool("Stop");
-    if(stop) minRobotDistanceForForce *= 1.5;
+    if(stop) minRobotDistanceForForce = 0.7*1.5;
```
mrlukasbos: ehm, magic?

PR #423, opened 2019-05-04
Practice tournament

mrlukasbos:


mrlukasbos:
This branch literally breaks things, even in GrSim so I think we should be very careful merging it. If we want to merge it at all.

PR #425, opened 2019-05-07
Interface api

mrlukasbos:
In this PR I make the functions for drawing to the interface much more powerful.

### debug points
Drawing a vector of points to the interface for debugging is as easy as: 
`interface::Input::drawDebugData(your_vector, [optional things]); `
The interface will clear the points for you.

### checkboxes
If a certain visualization is really useful you can draw it with an enumeration in front:
`interface::Input::drawData(PATHFINDING, your_vector, [optional things]); `
then it will only be shown when  PATHFINDING checkbox is shown in the interface


roboteam_ai/src/control/PositionUtils.cpp
```diff
@@ -85,7 +85,9 @@ std::vector<Vector2> PositionUtils::getPenaltyPositions(int number) {
                                  {lengthOffset, widthOffset},
                                  {lengthOffset, - widthOffset},
                                  {0, - widthOffset},
-                                 {- lengthOffset, - widthOffset}};
+                                 {- lengthOffset, - widthOffset},
```
rolfvdhulst: If I understand correctly this is just adding two positions in case we have a penalty, am I correct?

roboteam_ai/test/InterfaceTests/TreeVisualizerTest.cpp
```diff
@@ -83,6 +83,8 @@ TEST(TreeVisualizerTest, it_sets_proper_color_for_status) {
     EXPECT_EQ(treeVis->getColorForStatus(bt::Node::Status::Waiting), Qt::darkGray);
 }
 
+
```
rolfvdhulst: Why?

roboteam_ai/src/control/PositionUtils.cpp
mrlukasbos: yes :-) This fixed a segfault because with 8 robots there is one without a formation (and 2 if there is no keeper set)

roboteam_ai/test/InterfaceTests/TreeVisualizerTest.cpp
mrlukasbos: auto generated stuff did things

roboteam_ai/src/interface/api/Toggles.h
```diff
@@ -0,0 +1,41 @@
...
+};
+
+enum showType {NO_ROBOTS = 0, SELECTED_ROBOTS = 1, ALL_ROBOTS = 2};
+struct Toggle {
+    Visual vis;
+    showType defaultShowType;
+    QString title;
+};
+
+static std::vector<Toggle> toggles = {
```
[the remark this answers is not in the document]
mrlukasbos: it is designed to be easy. you have to define it in `toggles` and you are good to go.

PR #426, opened 2019-05-07
Fix shotcontroller problems

mrlukasbos:
fix shotcontroller inaccuracies

### DIFF
--- a/roboteam_ai/src/control/shotControllers/ShotController.cpp
+++ b/roboteam_ai/src/control/shotControllers/ShotController.cpp
@@ -57,7 +57,7 @@ ShotData ShotController::getShotData(world::Robot robot, Vector2 shotTarget, boo
         bool hasBall = world::world->ourRobotHasBall(robot.id, Constants::MAX_KICK_RANGE());
         if (hasBall && !genevaIsTurning) {
             shotData = shoot(robot, lineToDriveOver, shotTarget, chip);
-            std::cout<<" SHOOT";
+          //  std::cout<<" SHOOT";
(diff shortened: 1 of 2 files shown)


roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -124,196 +95,167 @@ bool ShotController::onLineToBall(const world::Robot &robot, std::pair<Vector2,
...
-    double diff=abs(aim-robot.angle);
-    if (precision==HIGH){
-        return diff<0.05;
-    }
-    if (precision==MEDIUM) {
-        return diff<0.25;
-    }
-    return diff<0.3;
-}
-std::pair<Vector2,Vector2> ShotController::offsetLine(std::pair<Vector2,Vector2> line, int genevaState){
```
rolfvdhulst: Should probably add this function back in

roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -124,196 +95,167 @@ bool ShotController::onLineToBall(const world::Robot &robot, std::pair<Vector2,
...
-    if (genevaState==3){
-        return line;
+    // determine the angle between the robot position and the shotline
+    Angle angleWithShotline = robotToBall.toAngle() - preferredShotVector.toAngle();
+    if (angleWithShotline.getAngle() > toRadians(20)) {
```
rolfvdhulst: We should add geneva state 1 and 5 back in

Issue #436, opened 2019-05-10
Decceleration limiter


simenbootsma:
Add a decceleration limiter to the existing acceleration limiter. Take into account the issue about having a unique position controller for every robot #417 . The purpose of this thing is to omit backflips when they are not intended.

Issue #417, opened 2019-05-02
Every robot needs to have a unique controller


rolfvdhulst:
Acceleration limiter now resets when we change from one type to the other within a skill- also the error's between controllers are not propagated nicely.

PR #443, opened 2019-05-10
Update robot objects in worldstate instead of replacing them

mrlukasbos:
needs  #449
first step to #417 and #436 

### Previously
Previously in the world state constantly a `worlddata` object was being created with a new vector of Robots. `world->getUs()` and `world->getThem()` returned a pointer to the most recent vector.  This made the robot objects basically memoryless and that's something we don't want anymore because of acceleration/deceleration limiting, as well as keeping track of geneva states and who knows what more. 

### New
In this PR the robot vector is created only once and then the corresponding objects are updated whenever new information is available. This makes the robot object persistent through our lifecycle and we can store shotcontroller/gotopos/geneva/pid/etc in there, which is going to help us a lot with control.

### Todo
- [x] Write/fix tests
- [x] Refactor some duplicate lines in world.cpp


roboteam_ai/src/world/World.cpp
```diff
@@ -141,7 +158,7 @@ const World::RobotPtr World::getRobotForId(int id, bool ourTeam) {
         worldCopy = worldDataPtr;
 
     }
-    const std::vector<Robot> &robots = ourTeam ? worldCopy->us : worldCopy->them;
+    const std::vector<Robot> robots = ourTeam ? worldCopy->us : worldCopy->them;
```
ThijsLuttikhuis: why not getUs / getThem, it's one line and you don't need the lock_guard

roboteam_ai/src/world/World.cpp
```diff
@@ -14,46 +10,67 @@ World worldObj;
...
+}
 
-    // update distance to ball, if it has ball etc.
-    for (auto &robot : newWorldData.us) {
-        robot.updateRobot(newWorldData.ball);
-    }
-    for (auto &robot : newWorldData.them) {
-        robot.updateRobot(newWorldData.ball);
-    }
+void World::updateRobotsFromData(Robot::Team team, const std::vector<roboteam_msgs::WorldRobot> &robotsFromMsg, std::vector<Robot> &robots, const Ball &ball, unsigned long worldNumber) const {
```
ThijsLuttikhuis: how is this function const.. ahh c++ plz

sideEffects by the way :)

PR #448, opened 2019-05-11
Feature/robot specific control

mrlukasbos:
needs #443 

Put all the control  under robots instead of skills. This has quite some consequences.

- [x] Update robots instead of replacing them (see #443) 
- [ ] ~Defence area / fieldsides  become robot specific as well. But we can probably hook this under RefStates as well, since they are dependent on the referee anyway. So these things are basically global and the same everywhere.~ **This can be done later in a seperate PR: #451** 
- [x] ball avoidance can be skill specific, so we need to be able to set that specifically. (A ballhandle GTP would also help a lot @ThijsLuttikhuis )
- [x] implement this in all skills
- [ ] implement acceleration limiter
- [ ] ~implement deceleration limiter~ **Later PR**
- [ ] ~move global geneva + geneva timing from shotcontroller to robot~ **Later PR**
- [x] shotcontroller accuracy and kickSpeed should be used in getShotData() function params instead of constructor.


roboteam_ai/src/control/positionControllers/PosController.cpp
```diff
@@ -77,6 +77,7 @@ double PosController::getAvoidBall() const {
 }
 
 void PosController::setAvoidBall(double ballDistance) {
+    std::cerr << "Manually setting 'avoidball' distance in skill" << std::endl;
```
Barisimre: print

PR #449, opened 2019-05-11
Hotfix/keeper claiming and fix defence area implementation

mrlukasbos:
Prevents the keeper problems we saw during the practice tournament.

Besides checking if we want to use a keeper we also need to check if the keeper is actually visible. otherwise just don't claim and don't tick it. 

and, if there is no keeper we need to claim ALL robots for the other strategies instead of ALL robots - 1 

I wrote tests for it. 

Besides that, I found some problems with defence areas. This has now been resolved (i think).
lastly, there was a segfault in the keeper as well, the fieldmutex was not always used, giving a sporadic segfault when assigning a keeper.

### DIFF
--- a/roboteam_ai/src/ApplicationManager.cpp
+++ b/roboteam_ai/src/ApplicationManager.cpp
@@ -114,7 +114,7 @@ void ApplicationManager::runOneLoopCycle() {
                 ai::robotDealer::RobotDealer::setKeeperID(ai::world::world->getUs().at(0).id);
             }
             keeperTree = BTFactory::getKeeperTree();
-            if (keeperTree) {
+            if (keeperTree && rtt::ai::robotDealer::RobotDealer::keeperExistsInWorld()) {
                 keeperTree->tick();
             }
         }
--- a/roboteam_ai/src/analysis/DecisionMaker.cpp
+++ b/roboteam_ai/src/analysis/DecisionMaker.cpp
@@ -14,8 +14,15 @@ namespace analysis {
 PlayStyle DecisionMaker::getRecommendedPlayStyle(BallPossession possession) {
     int amountOfRobots = world::world->getUs().size();
 
+    // if we use the keeper we should check if it is currently visible
+    // if it is not visible we should claim all robots visible in world
+    // otherwise we should subtract the keeper
+    bool keeperIsInWorld = std::find_if(world::world->getUs().begin(), world::world->getUs().end(), [](world::Robot robot) {
+        return robot.id == robotDealer::RobotDealer::getKeeperID();
+    }) != world::world->getUs().end();
+
     // subtract one robot if we need a keeper
-    if (robotDealer::RobotDealer::usesSeparateKeeper()) {
+    if (robotDealer::RobotDealer::usesSeparateKeeper() && keeperIsInWorld) {
         amountOfRobots = std::max(0, amountOfRobots-1);
     }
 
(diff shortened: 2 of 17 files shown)


roboteam_ai/test/ConditionTests/IsInDefenseAreaTest.cpp
```diff
@@ -30,6 +30,8 @@ TEST(DetectsInOurDefenseArea, IsInDefenseAreaTest)
     EXPECT_EQ(node.update(), bt::Node::Status::Failure);
 
     roboteam_msgs::GeometryFieldSize field;
+    field.field_width = 12;
```
Barisimre: I guess this is by design but looks weird

roboteam_ai/test/WorldTests/FieldTest.cpp
```diff
@@ -16,23 +16,30 @@ TEST(FieldTest, it_gets_and_sets_the_field) {
...
     for (int i = 0; i < 100; i ++) {
         auto x = testhelpers::WorldHelper::getRandomValue(- 6, - 4);
         auto y = testhelpers::WorldHelper::getRandomValue(0, 8);
 
-        // all points should be in our defence area
-        rtt::ai::world::field->set_field(field);
         bool inOurDefenceArea = rtt::ai::world::field->pointIsInDefenceArea(rtt::Vector2(x, y), true, 0.0);
+
+        if (!inOurDefenceArea) {
+            std::cout << rtt::Vector2(x, y) << std::endl;
```
Barisimre: Maybe remove the print

roboteam_ai/src/ApplicationManager.cpp
```diff
@@ -114,7 +114,7 @@ void ApplicationManager::runOneLoopCycle() {
                 ai::robotDealer::RobotDealer::setKeeperID(ai::world::world->getUs().at(0).id);
             }
             keeperTree = BTFactory::getKeeperTree();
-            if (keeperTree) {
+            if (keeperTree && rtt::ai::robotDealer::RobotDealer::keeperExistsInWorld()) {
```
Barisimre: We could have a get keeper back in vision tactic that we tick if the first one is true or in any case not ignore the fact that we used to have a keeper and now we dont

roboteam_ai/src/analysis/DecisionMaker.cpp
```diff
@@ -14,11 +14,10 @@ namespace analysis {
 PlayStyle DecisionMaker::getRecommendedPlayStyle(BallPossession possession) {
     int amountOfRobots = world::world->getUs().size();
 
-    // subtract one robot if we need a keeper
-    if (robotDealer::RobotDealer::usesSeparateKeeper()) {
+    // subtract one robot if we have a keeper
+    if (robotDealer::RobotDealer::usesSeparateKeeper() && robotDealer::RobotDealer::keeperExistsInWorld()) {
```
Barisimre: This means that if the keeper is lost in vision we think that we dont use a keeper anymore? Now sure how this will change behavior, in a good or bad way?

roboteam_ai/src/utilities/RobotDealer.cpp
```diff
@@ -367,6 +365,14 @@ void RobotDealer::setUseSeparateKeeper(bool useSeparateKeeper) {
     RobotDealer::useSeparateKeeper = useSeparateKeeper;
 }
 
+bool RobotDealer::keeperExistsInWorld() {
+    auto us = world::world->getUs();
+
+    return std::find_if(us.begin(), us.end(), [](world::Robot & robot) {
+        return robot.id == getKeeperID();
+    }) != us.end();
```
Barisimre: I think this really doesn't need to be a lambda since not many people this year or next will be proficient in functional style programming

PR #455, opened 2019-05-14
Feature/add dribbler to robot

ThijsLuttikhuis:
- made robot->geneva not static, added some features
- add dribbler to robot in the same way as geneva


roboteam_ai/src/world/Robot.cpp
```diff
@@ -89,21 +91,72 @@ double Robot::findBallDistance(const Vector2 &ballPos) {
...
+    abs(dribblerState - previousDribblerState)*timeToChangeOneDribblerLevel;
+}
+
+void Robot::setDribblerState(unsigned char dribbler) {
+
+    if (dribbler < 0 || dribbler > 7) {
+        std::cout << "setting invalid dribbler state (" << (int)dribbler <<
+                  ") for robot with id " << id << std::endl;
     }
+    else if (! workingGeneva) {
```
ThijsLuttikhuis: need to change this still :)

roboteam_ai/src/world/Robot.cpp
```diff
@@ -89,21 +91,72 @@ double Robot::findBallDistance(const Vector2 &ballPos) {
...
+        std::cout << "setting dribbler state (" << (int)dribbler <<
+                  ") for robot without working dribbler with id " << id << std::endl;
+    }
+    else {
+        previousGenevaState = genevaState;
```
ThijsLuttikhuis: and this

roboteam_ai/src/utilities/Constants.h
```diff
@@ -99,9 +99,11 @@ class Constants {
     static bool STD_TIMEOUT_TO_TOP();
 
     static std::map<int, bool> ROBOTS_WITH_WORKING_GENEVA();
+        static std::map<int, bool> ROBOTS_WITH_WORKING_DRIBBLER();
```
rolfvdhulst: ree indentation

roboteam_ai/src/world/Robot.cpp
```diff
@@ -12,24 +12,26 @@ namespace rtt {
...
+    else {
+        workingGeneva = false;
+        workingDribbler = false;
     }
 }
 
 Robot::Robot()
-        : distanceToBall(-1.0), iHaveBall(false), id(-1), angle(-1.0),
-          angularVelocity(-1.0), team(invalid), lastUpdatedWorldNumber(0) {
+        :distanceToBall(- 1.0), iHaveBall(false), lastUpdatedWorldNumber(0), genevaState(0), workingGeneva(false),
```
rolfvdhulst: would prefer if you initialize angle and angularvelocity with 0.0

roboteam_ai/src/world/Robot.cpp
```diff
@@ -12,24 +12,26 @@ namespace rtt {
...
-
-    if (id != -1 && genevaState.find(id) == genevaState.end()) {
-        genevaState[id] = 3;
+Robot::Robot(const roboteam_msgs::WorldRobot &copy, Team team,
+        unsigned char genevaState, unsigned char dribblerState, unsigned long worldNumber)
+        :distanceToBall(- 1.0), iHaveBall(false), lastUpdatedWorldNumber(worldNumber), genevaState(genevaState),
+         dribblerState(dribblerState),
+         id(copy.id), angle(copy.angle), pos(copy.pos), vel(copy.vel), angularVelocity(copy.w), team(team) {
+
+    if (id > - 1 && id < 16) {
```
rolfvdhulst: kind of prefer >=0 for readability here

roboteam_ai/src/world/Robot.cpp
```diff
@@ -89,21 +91,72 @@ double Robot::findBallDistance(const Vector2 &ballPos) {
...
+        previousGenevaState = genevaState;
+        genevaState = state;
+        timeGenevaChanged = world::world->getTime();
+    }
 }
 
-    const unsigned long &Robot::getLastUpdatedWorldNumber() const {
-        return lastUpdatedWorldNumber;
+bool Robot::isGenevaReady() const {
+    return world->getTime() - timeGenevaChanged >
```
rolfvdhulst: use brackets around things you group for comparison,  so that it's clear what you are comparing to what.

roboteam_ai/src/world/Robot.cpp
```diff
@@ -89,21 +91,72 @@ double Robot::findBallDistance(const Vector2 &ballPos) {
...
+bool Robot::hasWorkingGeneva() const {
+    return workingGeneva;
+}
+
+unsigned char Robot::getDribblerState() const {
+    return dribblerState;
+}
+
+bool Robot::isDribblerReady() const {
+    return world->getTime() - timeDribblerChanged >
```
rolfvdhulst: same as above.

rolfvdhulst: I don't like this function because this is not really how the dribbler works. We don't have to wait for a dribbler to change it's level.

roboteam_ai/src/utilities/Constants.h
ThijsLuttikhuis: fixed

roboteam_ai/src/world/Robot.cpp
ThijsLuttikhuis: does not matter, is - and should not be - used in any other way than creating an empty robot.

roboteam_ai/src/world/Robot.cpp
ThijsLuttikhuis: it is a check to see if the ball had time to match the speed of the dribbler

PR #395, opened 2019-04-30
Feature/ball handle pos control

ThijsLuttikhuis:
- [x] The tests are passing

### Summary
Give a short summary of what the changes are and why you made them. If applicable, apply screenshots as well. 

### Related issues
If there are related issues, reference them here.

fixes ...  
fixes ...

### Comments for the reviewer
If there are comments for the reviewer, leave them here

### Checklist for the reviewer
This acts as a guide to help you and/or the reviewer to analyze the code. 

**functionality**  
- Is the code is properly tested?
- Is the code working as expected
- Is the code is understandable and easy to read
- Are proper abstractions made? (no duplicate code, nice use of functions/objects)

**style**  
- Variables/objects/classes have proper names
- Proper use of variables: nu unused vars, no magic numbers
- No possible memory leaks. Mutexes properly used when multithreading? possible pointer issues?
- Are smart pointers used?
- unused variables?
- magic numbers?
- well documented?
- No unused imports?
- No leftover code / unused comments
- Proper indentation
- Proper use of namespaces

PR #457, opened 2019-05-14
Feature/dribble forward

RobZelluf:
Dribble forward when cannot pass. First needs better ball control (#343 or #395) ???


Barisimre (approved):

RobZelluf:
Put in #476

PR #464, opened 2019-05-14
Feature/fix receive defense area

### LINKED ISSUE #452: Receive should not pick receive points in the defence area's during normal play.
It does do this. Can probably just pick points on the line segment parts that are close but not in the defence area.

rolfvdhulst:
Needs the polygon and line classes in roboteam_utils to work, but fixes an important bug in field.

This PR fixes receiving going into the defense area.
fixes #452


roboteam_ai/src/skills/Receive.cpp
```diff
@@ -14,7 +15,7 @@ Receive::Receive(string name, bt::Blackboard::Ptr blackboard) : Skill(std::move(
 
 void Receive::onInitialize() {
     readyToPassSet = false;
-   // basicGtp.setCanMoveInDefenseArea(false);
+    canMoveInDefenseArea=properties->getBool("canMoveInDefenseArea");
```
mrlukasbos: this boolean should actually be dependent on refstate!

roboteam_ai/src/skills/Receive.cpp
```diff
@@ -62,9 +63,37 @@ void Receive::onTerminate(Status s) {
...
 Vector2 Receive::computeInterceptPoint(const Vector2& startBall, const Vector2& endBall) {
-    return robot->pos.project(startBall, endBall);
-}
+    double margin = 0.3;
+    Vector2 projectPos = Line(startBall,endBall).project(robot->pos);
+    if (canMoveInDefenseArea) {
+        return projectPos;
+    }
+    if (world::field->pointIsInDefenceArea(projectPos, true, margin)||world::field->pointIsInDefenceArea(projectPos, false, margin)) {
+        Polygon dUs(world::field->getDefenseArea(false, margin)), dThem(world::field->getDefenseArea(true, margin));
```
mrlukasbos: `dUs `is a weird name?

roboteam_ai/src/skills/Receive.cpp
```diff
@@ -62,9 +63,37 @@ void Receive::onTerminate(Status s) {
...
+        interface::Input::drawData(interface::DEBUG,world::field->getDefenseArea(false, margin),Qt::red,-1,interface::Drawing::LINES_CONNECTED);
```
mrlukasbos: we should probably always draw this thing on debug, but we should move it to the interface

roboteam_ai/src/skills/Receive.cpp
rolfvdhulst: Could be, but we could also manually set it in the relevant trees that change based on refState regardless.

rolfvdhulst: I think that is probably easier.

roboteam_ai/src/skills/Receive.cpp
rolfvdhulst: refactored to defUs

roboteam_ai/src/skills/Receive.cpp
rolfvdhulst: I removed it from here now, since we use a lot of different margins.

roboteam_ai/src/skills/Receive.cpp
ThijsLuttikhuis: dusssss

ThijsLuttikhuis: dthemmmm

ThijsLuttikhuis: please initialize the two polygons on seperate lines

roboteam_ai/src/skills/Receive.cpp
```diff
@@ -69,22 +69,24 @@ Vector2 Receive::computeInterceptPoint(const Vector2& startBall, const Vector2&
         return projectPos;
     }
     if (world::field->pointIsInDefenceArea(projectPos, true, margin)||world::field->pointIsInDefenceArea(projectPos, false, margin)) {
-        Polygon dUs(world::field->getDefenseArea(false, margin)), dThem(world::field->getDefenseArea(true, margin));
-        interface::Input::drawData(interface::DEBUG,world::field->getDefenseArea(false, margin),Qt::red,-1,interface::Drawing::LINES_CONNECTED);
+        Polygon defUs(world::field->getDefenseArea(true, margin));
```
ThijsLuttikhuis: still wrong name

roboteam_ai/src/skills/Receive.cpp
```diff
@@ -62,9 +63,39 @@ void Receive::onTerminate(Status s) {
...
 Vector2 Receive::computeInterceptPoint(const Vector2& startBall, const Vector2& endBall) {
-    return robot->pos.project(startBall, endBall);
-}
+    double margin = 0.3;
+    Vector2 projectPos = Line(startBall,endBall).project(robot->pos);
+    if (canMoveInDefenseArea) {
+        return projectPos;
+    }
+    if (world::field->pointIsInDefenceArea(projectPos, true, margin)||world::field->pointIsInDefenceArea(projectPos, false, margin)) {
+        Polygon defenceAreaUs(world::field->getDefenseArea(true, margin));
```
ThijsLuttikhuis: perfect

roboteam_ai/src/skills/Receive.cpp
mrlukasbos: but also more prone to errors. Just keep it for now, in another branch I am redesigning the application of rules anyway

PR #469, opened 2019-05-15
Feature/acceleration limiter

rolfvdhulst:
updates the acceleration limiter to be kind to our motors.


roboteam_ai/src/control/ControlUtils.cpp
```diff
@@ -235,7 +235,22 @@ double ControlUtils::calculateMaxAcceleration(const Vector2 &vel, double angle)
     double a = abs(toVectorDiff.x);
     return Constants::MAX_ACC_UPPER() * (a) + Constants::MAX_ACC_LOWER() * (1-a);
 }
-
+Vector2 ControlUtils::accelerationLimiterNew(const Vector2 &vel,double robotAngle, const Vector2 &prevVel) {
+    // calculate max
+    Vector2 diff=vel-prevVel;
+    Angle angleDiff=diff.angle()-robotAngle;
+    Vector2 toVectorDiff=angleDiff.toVector2();
+    double a=abs(toVectorDiff.x);
```
simenbootsma: Why use toVectorDiff, isn't `a=abs(cos(angleDiff))` sufficient?

rolfvdhulst: This was exactly the same code as you used previously. It's functionally the same, so I don't mind changing it.

roboteam_ai/src/control/positionControllers/PosController.cpp
```diff
@@ -35,12 +35,17 @@ PosVelAngle PosController::controlWithPID(const RobotPtr &robot, PosVelAngle tar
     pidCommand.vel = calculatePIDs(robot, target);
 
     // acceleration limiter
-    double maxAcc = control::ControlUtils::calculateMaxAcceleration(pidCommand.vel, pidCommand.angle);
-    if (prevVel == 0.0) prevVel = robot->vel.length();
-    pidCommand.vel = control::ControlUtils::accelerationLimiter(pidCommand.vel, maxAcc, prevVel);
+    if (prevVel == Vector2(0.0,0.0)) {
+        prevVel = robot->vel;
+    }
+    //OLD
```
mrlukasbos: we have git so you don't need this

rolfvdhulst: this was for testing purposes so recompiling does not take ages. We were still testing this

PR #479, opened 2019-05-22
Change getBall to use ballHandlePosControl

RobZelluf:
GetBall uses BallHandle now,
Add feature in BallHandle to keep the ball in the field, or more generally, take the velocity of the ball into account when grabbing it.


roboteam_ai/src/control/positionControllers/BallHandlePosControl.cpp
```diff
@@ -125,54 +128,47 @@ RobotCommand BallHandlePosControl::travelWithBall(TravelStrategy travelStrategy)
...
+    case no_preference:return {};
     }
 
     std::cout << "travel case not handled" << std::endl;
-    return {
-    };
+    return {};
+}
+
+RobotCommand BallHandlePosControl::sendBackwardsCommand() {
```
mrlukasbos: nice!!

PR #480, opened 2019-05-22
Feature/test2

mrlukasbos:
wait for #494


roboteam_ai/src/bt/Role.cpp
```diff
@@ -40,10 +40,14 @@ Role::Role(std::string name) {
 }
 
 std::vector<Node::Ptr> Role::getChildren() {
-    return std::vector<Node::Ptr>{child};
+    if (child) {
```
rolfvdhulst: Is there a reason you changed this? It was actually kind of nice if this segfaulted because it was an easy way to tell your tree is invalid. Should atleast contain an error message IMO.

roboteam_ai/src/control/positionControllers/BallHandlePosControl.cpp
```diff
@@ -79,7 +79,7 @@ RobotCommand BallHandlePosControl::getRobotCommand(const RobotPtr &r,
```
rolfvdhulst: Why make such a mess of all the print statements here? I liked the way it was before better because atleast there is a clear place where they are. Now you removed some of the switch cases but also kept some. If you change it, change them all.

mrlukasbos: no i did not remove them and i did not make a mess. I made them testable, meaning that now they have a proper input and output

roboteam_ai/src/bt/Role.cpp
mrlukasbos: wel... i got a segfault there

rolfvdhulst: So, there is a tree with a role with no child in it on this branch, which should never be the case (or your test has a role with no child in it), so I'd rather have it there. Our default behavior for tree structures not being right at the moment is AI just segfaulting on you so unless we are going to change that this atleast needs a print.

mrlukasbos:
Ballhandle changed so much I can better start over with this

PR #486, opened 2019-05-23
Feature/real life shotcontroller

Barisimre:
[PR by @mrlukasbos ] 
works beautiful!


roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -47,6 +47,8 @@ ShotData ShotController::getShotData(world::Robot robot, Vector2 shotTarget, boo
     bool validAngle = robotAngleIsGood(robot, lineToDriveOver, precision);
 
     ShotData shotData;
+    std::cout<<" Online: "<<isOnLineToBall <<" behind: "<<isBehindBall<<" valid: "<<validAngle<<" isShooting: " <<isShooting<<std::endl;
```
Barisimre: Remove print or make it like gotopos debugs

PR #511, opened 2019-05-29
Fix free kick and penalty for 8 robots

### LINKED ISSUE #506: Free kick formation does not work with 8 robots
The free kick formation only works with 7 robots or less. otherwise it gives the error: wrong input for hungarian: more robots than positions

Barisimre:
Fixes #506


roboteam_ai/src/Switches.cpp
```diff
@@ -496,7 +498,8 @@ bt::Node::Ptr Switches::tacticSwitch(std::string name, bt::Blackboard::Ptr prope
                     {"line2", robotType::RANDOM},
                     {"line4", robotType::RANDOM},
                     {"line5", robotType::RANDOM},
-                    {"line3", robotType::RANDOM}
+                    {"line3", robotType::RANDOM},
```
ThijsLuttikhuis: why is this not in order?

PR #541, opened 2019-06-03
Fix/keeper commands

### LINKED ISSUE #492: Keeper sometimes not entering defence area (probably because of the ruleset)
Sometimes when a new robot becomes keeper it tries to enter the defence area but it can't.

### LINKED ISSUE #530: Keeper tested and improved
The filed line checks for the penalty also not tested

### LINKED ISSUE #560: Poscontroller's 'customCanMoveInDefenceArea' field overwrites other places unintentionally
Quite difficult to fix for me. I suggest just completely listening to GameStateManager and always listen to the same ruleset and correctly defining that; no Overriding behavior in the goToPosses that disables listening to GameStateManager. Currently also NumTreePosControlTest fails

rolfvdhulst:
fixes some problems for the keeper (tested in Real life)
fixes #492 
fixes #530, but the ballhandling is still meh.
fixes #560

There is some poscontroller problems fixed in here too.


roboteam_ai/src/control/positionControllers/PosController.cpp
```diff
@@ -63,7 +63,9 @@ double PosController::getAvoidBallDistance() const {
 void PosController::setAvoidBallDistance(double ballDistance) {
     customAvoidBallDistance = ballDistance;
 }
-
+void PosController::setListenToInterface(bool listenToInterface) {
```
mrlukasbos: i do not like this. By conditionally listening to the interface, how can you know if editing a pid in the interface will actually work? this obscures PID tuning a lot and you will shoot yourself in the foot eventually with something like this.  (I recall we had this feature a few months ago and I explicitly removed it because of that)

roboteam_ai/src/skills/InterceptBall.cpp
```diff
@@ -15,72 +15,93 @@ InterceptBall::InterceptBall(rtt::string name, bt::Blackboard::Ptr blackboard)
...
 
-    interface::Input::drawData(interface::Visual::INTERCEPT, {ballStartPos, ballEndPos}, Qt::darkCyan, robot->id, interface::Drawing::LINES_CONNECTED);
-    interface::Input::drawData(interface::Visual::INTERCEPT, {interceptPos}, Qt::cyan, robot->id, interface::Drawing::DOTS, 5, 5);
-
+    interface::Input::drawData(interface::Visual::INTERCEPT, {ballStartPos, ballEndPos}, Qt::darkCyan, robot->id,
+            interface::Drawing::LINES_CONNECTED);
+    interface::Input::drawData(interface::Visual::INTERCEPT, {interceptPos}, Qt::cyan, robot->id,
+            interface::Drawing::DOTS, 5, 5);
+    // if we are not already rotating
+    if (currentProgression!=INPOSITION&&deltaPos.length()>0.05) {
```
mrlukasbos: what does this magic value do?

roboteam_ai/src/skills/Keeper.cpp
```diff
@@ -13,12 +13,12 @@ Keeper::Keeper(rtt::string name, bt::Blackboard::Ptr blackboard)
         :Skill(std::move(name), std::move(blackboard)) { }
 
 void Keeper::onInitialize() {
-
     goalPos = world::field->get_our_goal_center();
     goalwidth = world::field->get_field().goal_width;
-
     //Create arc for keeper to drive on
     blockCircle=control::ControlUtils::createKeeperArc();
+    posController.setListenToInterface(false);
```
mrlukasbos: separate poscontroller in keeper? we explicitly removed such functionality only a month ago. Why add it now?

mrlukasbos: when a robot is keeper the robot pid and robot poscontroller should be updated. this shouldn't happen in the skill

roboteam_ai/src/utilities/GameStateManager.cpp
```diff
@@ -39,6 +39,19 @@ void GameStateManager::forceNewGameState(RefCommand cmd) {
     interface::Output::setInterfaceGameState(strategymanager.getRefGameStateForRefCommand(cmd));
     strategymanager.forceCurrentRefGameState(cmd);
 }
-
+bool GameStateManager::canEnterDefenseArea(int robotId) {
```
mrlukasbos: this is nice, i like it

roboteam_ai/src/control/positionControllers/PosController.cpp
ThijsLuttikhuis: I have to agree with Lukas on this.

ThijsLuttikhuis: I think, as you added the keeper pid in interface, this could be removed again?

roboteam_ai/src/skills/InterceptBall.cpp
```diff
@@ -15,72 +15,89 @@ InterceptBall::InterceptBall(rtt::string name, bt::Blackboard::Ptr blackboard)
 //TODO: make prediction for the RobotPtr if it can even intercept the ball at all from the initialization state.
 void InterceptBall::onInitialize() {
     keeper = properties->getBool("Keeper");
-
+    if (keeper) {
+        poscontroller.setListenToInterface(false);
```
ThijsLuttikhuis: hmm

roboteam_ai/src/skills/InterceptBall.cpp
```diff
@@ -163,11 +170,12 @@ Vector2 InterceptBall::computeInterceptPoint(Vector2 startBall, Vector2 endBall)
         else {
             // if the Line does not intercept it usually means the ball is coming from one of the corners-ish to the keeper
             // For now we pick the closest point to the (predicted) line of the ball
-            interceptionPoint = Vector2(robot->pos).project(startBall, endBall); }
+            interceptionPoint = Vector2(robot->pos).project(startBall, endBall);
```
ThijsLuttikhuis: you should probably make startball, endball a line and project the robot->pos onto that

roboteam_ai/src/skills/InterceptBall.h
```diff
@@ -16,16 +16,16 @@ class InterceptBall :public Skill {
...
         const double BALL_DEFLECTION_ANGLE = 30.0/180.0*M_PI;    // Angle at which a ball is considered 'deflected'
-        const double INTERCEPT_POSDIF = 0.04;    // Meters acceptable deviation
-
+        const double INTERCEPT_POSDIF = 0.015;    // Meters acceptable deviation
+        const double TURNING_DISTANCE=0.05;// m distance at which we start turning towards the target angle (if desired)
+        const double TURN_TIME=0.3; //estimated time to make a large rotation (90 to 180 degrees) on the current robot
+        const double GOAL_MARGIN=Constants::BALL_RADIUS()+0.05;// if the ball is shot next to the goal within this margin we still try to intercept
         enum Progression {
-          INTERCEPTING, CLOSETOPOINT, INPOSITION, BALLDEFLECTED, BALLMISSED
+          INTERCEPTING, INPOSITION, BALLDEFLECTED, BALLMISSED
```
ThijsLuttikhuis: use enters for seperating enum variables

roboteam_ai/src/skills/InterceptBall.h
```diff
@@ -35,7 +35,8 @@ class InterceptBall :public Skill {
         Vector2 deltaPos;
         int tickCount, maxTicks;
```
ThijsLuttikhuis: please init them om seperate lines

roboteam_ai/src/skills/Keeper.cpp
```diff
@@ -29,33 +31,33 @@ Keeper::Status Keeper::onUpdate() {
 
     if (ball->pos.x < 0) {
         auto attacker = world::world->getRobotClosestToPoint(ball->pos, world::THEIR_ROBOTS);
-        if ((ball->pos - attacker->pos).length() < 0.3) {
+        if (attacker&&(ball->pos - attacker->pos).length() < 0.3) {
```
ThijsLuttikhuis: ctrl alt L && magic number

roboteam_ai/src/skills/Keeper.cpp
```diff
@@ -29,33 +31,33 @@ Keeper::Status Keeper::onUpdate() {
...
             setGoalPosWithAttacker(attacker);
         }
     }
 
     blockPoint = computeBlockPoint(ballPos);
 
-    if (!world::field->pointIsInField(blockPoint, static_cast<float>(Constants::OUT_OF_FIELD_MARGIN()))) {
+    if (!world::field->pointIsInField(blockPoint, -static_cast<float>(Constants::OUT_OF_FIELD_MARGIN()))) {
```
ThijsLuttikhuis: do " static_cast<float> (- x) " ipv " - static_cast<float(x) "

roboteam_ai/src/skills/Keeper.cpp
```diff
@@ -29,33 +31,33 @@ Keeper::Status Keeper::onUpdate() {
...
         blockPoint=goalPos;
         blockPoint.x+=Constants::KEEPER_CENTREGOAL_MARGIN();
         command.w=0;
     }
     else{
-        command.w=(ballPos-blockPoint).angle();
+        command.w=Angle((ballPos-blockPoint).angle()+M_PI_2).getAngle();
```
ThijsLuttikhuis: command.w = (ballPos-blockPoint).toAngle() + M_PI_2 does the same thing

roboteam_ai/src/utilities/Constants.cpp
```diff
@@ -239,9 +239,11 @@ std::vector<QColor> Constants::TACTIC_COLORS() {
 
 pidVals Constants::standardNumTreePID() { return GRSIM() ? pidVals(4.2, 0.0, 1.4) : pidVals(3.1, 0.0, 0.6); }
 
-pidVals Constants::standardBasicPID() { return GRSIM() ? pidVals(1.6, 0.0, 0.15) : pidVals(2.8, 0.0, 0.0); }
+pidVals Constants::standardBasicPID() { return GRSIM() ? pidVals(3.4, 0.0, 0.4) : pidVals(3.4, 0.0, 0.4); }
```
ThijsLuttikhuis: not sure if the standardBasicPID should be changed

roboteam_ai/src/skills/Keeper.cpp
rolfvdhulst: That is not explicitly limited; this was what I tried first but gives me errors if the first Angle is larger than M_PI_2

roboteam_ai/src/utilities/Constants.cpp
rolfvdhulst: These were the values from testing/ that we used.

roboteam_ai/src/control/positionControllers/PosController.cpp
mrlukasbos: after discussion I understand it. I think a comment should be added here that we should be careful with this function, plus a test that this works properly

PR #547, opened 2019-06-04
Use BallHandle in ShotController

ThijsLuttikhuis:


mrlukasbos:
first we should merge the other shotcontroller branches

mrlukasbos:
deprecated because of #570

PR #549, opened 2019-06-04
shotcontroller upgrade for ball sensor

mrlukasbos:
NEEDS BALLSENSOR
shotcontroller had some duplicate code which was useless anyway. Now it works in both grsim and real life using the same code and the ballsensor because of the ballsensor


mrlukasbos:
I added functionality to make sure it also works when robots have no (working) ballsensors if you indicate it in constants. in that case it uses the old method.

CMakeLists.txt
```diff
@@ -289,11 +289,11 @@ set(CONTROL_SOURCES
         ${PROJECT_SOURCE_DIR}/roboteam_ai/src/control/pid.h
 
         # PositionControllers
-        ${PROJECT_SOURCE_DIR}/roboteam_ai/src/control/positionControllers/RobotCommand.h
-        ${PROJECT_SOURCE_DIR}/roboteam_ai/src/control/positionControllers/BasicPosControl.cpp
-        ${PROJECT_SOURCE_DIR}/roboteam_ai/src/control/positionControllers/BasicPosControl.h
-        ${PROJECT_SOURCE_DIR}/roboteam_ai/src/control/positionControllers/PosController.cpp
-        ${PROJECT_SOURCE_DIR}/roboteam_ai/src/control/positionControllers/PosController.h
+        roboteam_ai/src/control/RobotCommand.h
```
ThijsLuttikhuis: add projectsourcedir

PR #555, opened 2019-06-05
Feature/fix ball placement 2

RobZelluf:
Some stuff in dribble is changed, someone needs to double check whether this is correct.


roboteam_ai/src/skills/ballPlacement/BallPlacementReceive.cpp
```diff
@@ -18,25 +18,21 @@ bt::Node::Status BallPlacementReceive::onUpdate() {
...
     }
 
     if (coach::g_pass.isPassed()) {
-        // Check if the ball was deflected
-//        if (passFailed()) {
-//            publishRobotCommand(); // halt
-//            return Status::Running;
+//        if(ballDeflected()) {
+//            return Status::Failure;
 //        }
```
rolfvdhulst: no commented code please; also why is this if statement empty? This seems like a testing artifact or something else weird.

rolfvdhulst (changes requested):
Seems fine. See small comment.

rolfvdhulst (approved):

PR #557, opened 2019-06-05
Fix kicker and chipper vel

RobZelluf:
Make sure that kicking and chipping works for both GrSim and Serial by making them equal in publishRobotCommand


roboteam_ai/src/interface/widgets/MainControlsWidget.cpp
```diff
@@ -184,6 +184,11 @@ void MainControlsWidget::updatePause() {
         pauseBtn->setStyleSheet("background-color: #cc0000;");
     }
 }
+void MainControlsWidget::keyPressEvent(QKeyEvent *event) {
```
mrlukasbos: what is this doing here? I believe this was in development already...

mrlukasbos (approved):

PR #558, opened 2019-06-05
Feature/ball handle get ball

ThijsLuttikhuis:
Ball handle takes a moving ball into account


roboteam_ai/src/control/ballHandling/DribbleForwards.cpp
```diff
@@ -13,9 +13,10 @@ namespace ai {
 namespace control {
 
 
-RobotCommand DribbleForwards::getRobotCommand(const world::Robot::RobotPtr &r, const Vector2 &targetP,
+RobotCommand DribbleForwards::getRobotCommand(world::Robot::RobotPtr r, const Vector2 &targetP,
         const Angle &targetA) {
-    robot = r;
+
+    robot = std::move(r);
```
rolfvdhulst: I personally don't really like std::move here, as you are editing the robotPtr object which becomeso a nullptr now, so if dribbleForwards tries to access the robotptr afterwards is null now. Is there any reason why you made this change?

rolfvdhulst: also if this is accessing world directly you are erasing robot pointers in world

roboteam_ai/src/control/ballHandling/RotateAroundBall.cpp
```diff
@@ -1,3 +1,5 @@
+#include <utility>
```
rolfvdhulst: not necessary

roboteam_ai/src/control/ballHandling/RotateAroundBall.cpp
```diff
@@ -10,11 +12,10 @@ namespace rtt {
 namespace ai {
 namespace control {
 
-RobotCommand RotateAroundBall::getRobotCommand(const world::Robot::RobotPtr &r, const Vector2 &targetP,
+RobotCommand RotateAroundBall::getRobotCommand(const world::Robot::RobotPtr &robot, const Vector2 &targetP,
```
rolfvdhulst: If you are going to remove const ref from getRobotCommand do it everywhere please.

roboteam_ai/src/control/numTrees/NumTreePosControl.h
```diff
@@ -36,7 +36,7 @@ class NumTreePosControl : public BasicPosControl {
         RobotCommand computeCommand(const Vector2 &exactTargetPos);
 
         // constants
-        const double MAX_CALCULATION_TIME = 20.0;         // Max calculation time in ms
+        static constexpr double MAX_CALCULATION_TIME = 20.0;         // Max calculation time in ms
```
rolfvdhulst: don't we all love CPP

roboteam_ai/src/control/ballHandling/RotateAroundBall.cpp
ThijsLuttikhuis: yes, sorry, fixed

roboteam_ai/src/control/ballHandling/RotateAroundBall.cpp
ThijsLuttikhuis: fixed

roboteam_ai/src/control/ballHandling/DribbleBackwards.cpp
```diff
@@ -1,3 +1,5 @@
+#include <utility>
```
rolfvdhulst: Don't forget this one

ThijsLuttikhuis: yep

PR #559, opened 2019-06-05
Show some more useful info when robots have non-working features

mrlukasbos:
Show some more useful info when robots have non-working features.

image

### DIFF
--- a/roboteam_ai/src/utilities/Constants.cpp
+++ b/roboteam_ai/src/utilities/Constants.cpp
@@ -150,9 +150,7 @@ bool Constants::STD_SHOW_VELOCITIES() { return true; }
 
 bool Constants::STD_SHOW_ANGLES() { return false; }
 
-bool Constants::STD_SHOW_PATHS_ALL() { return false; }
-
-bool Constants::STD_SHOW_PATHS_CURRENT() { return true; }
+bool Constants::STD_SHOW_ROBOT_INVALIDS() { return true; }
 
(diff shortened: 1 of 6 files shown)


roboteam_ai/src/utilities/Constants.cpp
```diff
@@ -186,9 +184,9 @@ std::map<int, bool> Constants::ROBOTS_WITH_WORKING_GENEVA() {
 
 std::map<int, bool> Constants::ROBOTS_WITH_WORKING_DRIBBLER() {
     static std::map<int, bool> workingDribblerRobots;
-    workingDribblerRobots[0] = true;
-    workingDribblerRobots[1] = true;
-    workingDribblerRobots[2] = true;
+    workingDribblerRobots[0] = false;
```
rolfvdhulst: Please revert these :)

roboteam_ai/src/utilities/Constants.cpp
```diff
@@ -166,8 +164,8 @@ std::map<int, bool> Constants::ROBOTS_WITH_WORKING_GENEVA() {
     static std::map<int, bool> workingGenevaRobots;
     workingGenevaRobots[0] = true;
     workingGenevaRobots[1] = true;
-    workingGenevaRobots[2] = true;
-    workingGenevaRobots[3] = true;
+    workingGenevaRobots[2] = false;
```
rolfvdhulst: Testing artifact

PR #575, opened 2019-06-07
Make sure all tactics have 8 roles

### LINKED ISSUE #510: Make sure every tactic in switches and JSONS has 8 roles

mrlukasbos:
fixes #510


roboteam_ai/src/Switches.cpp
```diff
@@ -552,7 +561,8 @@ bt::Node::Ptr Switches::tacticSwitch(std::string name, bt::Blackboard::Ptr prope
                      {"receive2", robotType::RANDOM},
                      {"receive3", robotType::RANDOM},
                      {"receive4", robotType::RANDOM},
-                     {"receive5", robotType::RANDOM}
+                     {"receive5", robotType::RANDOM},
```
ThijsLuttikhuis: probably fine, but why are there not 8 receivers?

ThijsLuttikhuis: nothing

PR #576, opened 2019-06-07
More accurate geneva logging and timing

### LINKED ISSUE #463: geneva state should be checked in publishRobotCommand (if the turn is actually legal) as a last failsafe

### LINKED ISSUE #487: Improve geneva turn timing
We should test the turning of the geneva and timing that

mrlukasbos:
The setGeneva log was triggered when it was not necessary. this should fix that.

it also removes the kick and chip skill because they don't check for geneva states and we have the shotcontroller

fixes #487
fixes #463


roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -26,11 +26,12 @@ RobotCommand ShotController::getRobotCommand(world::Robot robot, const Vector2 &
     bool robotAlreadyVeryClose = robot.pos.dist(ball->pos) < 3.0*Constants::ROBOT_RADIUS();
     int currentDesiredGeneva = robot.getGenevaState();
 
-    if (useAutoGeneva && robot.hasWorkingGeneva() && ! genevaIsTurning && ! robotAlreadyVeryClose) {
+    if (useAutoGeneva && robot.hasWorkingGeneva() && robot.isGenevaReady() && ! robotAlreadyVeryClose) {
         currentDesiredGeneva = determineOptimalGenevaState(robot, aimTarget);
     }
 
     if (chip) {
+        std::cout << "chipping -> simulate shotcontroller to use geneva 3" << std::endl;
```
ThijsLuttikhuis: should this print be here?

roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -50,12 +51,12 @@ RobotCommand ShotController::getRobotCommand(world::Robot robot, const Vector2 &
...
     // std::cout<<" Online: "<<isOnLineToBall <<" behind: "<<isBehindBall<<" valid: "<<validAngle<<" isShooting: " <<isShooting<<std::endl;
     if (isOnLineToBall && isBehindBall && (validAngle || isShooting)) {
-        if (genevaIsTurning) {
+        if (!robot.isGenevaReady() && !chip) {
             isShooting = false;
             // just stand still at the right angle
             shotData.vel = {0.0, 0.0};
             shotData.angle = (lineToDriveOver.second - lineToDriveOver.first).angle();
-            std::cout << "Not shooting because geneva is turning for " << secondsToTurnGeneva << "s" << std::endl;
+            std::cout << "Not shooting because geneva is turning" << std::endl;
```
ThijsLuttikhuis: cout

roboteam_ai/src/skills/Skill.cpp
```diff
@@ -22,14 +22,8 @@ void Skill::publishRobotCommand() {
     if(Constants::GRSIM() && ourSideParam=="right"){
       command=rotateRobotCommand(command);
     }
-    if (std::isnan(command.x_vel) || std::isnan(command.y_vel)) {
```
ThijsLuttikhuis: maybe leave the check in for now.
we are not 100% sure if it is indeed fixed.

PR #584, opened 2019-06-08
Use shootPenalty skill with geneva fixed

RobZelluf:


roboteam_ai/src/control/shotControllers/ShotController.cpp
```diff
@@ -63,7 +68,7 @@ RobotCommand ShotController::getRobotCommand(world::Robot robot, const Vector2 &
             if (robot.hasWorkingBallSensor()) {
                 shotData = shoot(robot, lineToDriveOver, aimTarget, chip, ballspeed);
             } else {
-                shotData = shootWithoutBallSensor(robot, lineToDriveOver, aimTarget, chip, ballspeed);
+                shotData = shoot(robot, lineToDriveOver, aimTarget, chip, ballspeed);
```
ThijsLuttikhuis: should this be shoot as well??

ThijsLuttikhuis (approved):

roboteam_ai/src/control/shotControllers/ShotController.cpp
mrlukasbos: This should indeed be changed back

roboteam_ai/src/skills/gotopos/GoToPos.cpp
```diff
@@ -76,7 +76,7 @@ bt::Node::Status GoToPos::onUpdate() {
             return Status::Success;
         }
     }
-    if (command.x_vel == 0 && command.y_vel == 0 && command.w == 0) {
+    if (command.x_vel == 0 || command.y_vel == 0 || command.w == 0) {
```
mrlukasbos: why is this change?

RobZelluf: Because it otherwise breaks

PR #603, opened 2019-06-14
Hotfix/segfault v2

ThijsLuttikhuis:
Make trees tick one more time when we do not have robots anymore.
_________________
This fixes the 'free' bug:

whenever all robots get out of vision, 'weHaveRobots()' returns false, so the trees stop ticking.
this causes robotdealer to not update for the last robot removed from world and stuff breaks,
so this will tick runOneLoopCycle one more time, so robotdealer has time to remove the robots properly.

### DIFF
--- a/roboteam_ai/src/ApplicationManager.cpp
+++ b/roboteam_ai/src/ApplicationManager.cpp
@@ -60,7 +60,7 @@ void ApplicationManager::loop() {
 }
 
 void ApplicationManager::runOneLoopCycle() {
-    if (ai::world::world->weHaveRobots()) {
+    if (weHaveRobots) {
         ai::analysis::GameAnalyzer::getInstance().start();
 
         // Will do things if this is a demo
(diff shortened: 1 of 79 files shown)


roboteam_ai/src/ApplicationManager.cpp
```diff
@@ -128,6 +128,8 @@ void ApplicationManager::runOneLoopCycle() {
         std::cout <<"NO FIRST WORLD" << std::endl;
         ros::Duration(0.2).sleep();
     }
+
+    weHaveRobots = ai::world::world->weHaveRobots();
```
mrlukasbos: why?

ThijsLuttikhuis: This fixes the 'free' bug:

whenever all robots get out of vision, 'weHaveRobots()' returns false, so the trees stop ticking.
this causes robotdealer to not update for the last robot removed from world and stuff breaks,
so this will tick runOneLoopCycle one more time, so robotdealer has time to remove the robots properly.

ThijsLuttikhuis: my bad though, wasnt clear from the code itselves :)

mrlukasbos: Wow. Impressive. How did you find that?

PR #605, opened 2019-06-14
Feature/fix shared from this

ThijsLuttikhuis:


rolfvdhulst (commented):
Changes look fine but I don't think we should merge this right now, it's really a codebase change but could add unforseen functionality changes since this is one of the more complicated parts in NumTrees. I say; keep this but merge it later when we can test with robots and have time for this. If then everything still works fine, we use it.

PR #609, opened 2019-06-14
GameStates for ShootOut

rolfvdhulst:
Working trees are not added yet but will be added later. Needs to be tested with refbox still.


mrlukasbos:
This definitely needs a test before merging

roboteam_ai/src/utilities/StrategyManager.h
```diff
@@ -52,12 +52,17 @@ class StrategyManager {
          RefGameState(RefCommand::PREPARE_PENALTY_US,   "penalty_us_prepare_strategy",     "keeper_formation_tactic",         "default",  false, RefCommand::DO_PENALTY),
          RefGameState(RefCommand::PREPARE_PENALTY_THEM, "penalty_them_prepare_strategy",   "keeper_penalty_prepare_tactic",   "default",  false, RefCommand::DEFEND_PENALTY),
 
+         // These two prepares are 'custom' because the refbox setup is stupid and should send them to us seperately
```
mrlukasbos: don't blame the refbox in open source code

roboteam_ai/src/utilities/GameStateManager.cpp
```diff
@@ -13,7 +13,21 @@ roboteam_msgs::RefereeData GameStateManager::getRefereeData() {
 
 void GameStateManager::setRefereeData(roboteam_msgs::RefereeData refMsg) {
     GameStateManager::refMsg = refMsg;
-    strategymanager.setCurrentRefGameState(static_cast<RefCommand>(refMsg.command.command));
+    auto cmd = static_cast<RefCommand>(refMsg.command.command);
+    //if we have shootouts, we need different handling
+    if (static_cast<int>(refMsg.stage.stage) == roboteam_msgs::RefereeStage::PENALTY_SHOOTOUT) {
+        if (cmd == RefCommand::PREPARE_PENALTY_US) {
```
ThijsLuttikhuis: why do you need this only for penalty shootouts?

roboteam_ai/src/utilities/GameStateManager.cpp
```diff
@@ -13,7 +13,21 @@ roboteam_msgs::RefereeData GameStateManager::getRefereeData() {
 
 void GameStateManager::setRefereeData(roboteam_msgs::RefereeData refMsg) {
     GameStateManager::refMsg = refMsg;
-    strategymanager.setCurrentRefGameState(static_cast<RefCommand>(refMsg.command.command));
+    auto cmd = static_cast<RefCommand>(refMsg.command.command);
```
mrlukasbos: this should happen in strategymanager, not in gamestate manager, for consistency.

roboteam_ai/src/utilities/GameStateManager.cpp
rolfvdhulst: I am confused, what are you referring to? ShootOuts have different rules than normal penalties.

roboteam_ai/src/utilities/GameStateManager.cpp
rolfvdhulst: What exactly are you referring to with 'this' here? The static casting? I do not follow you

roboteam_ai/src/utilities/GameStateManager.cpp
rolfvdhulst: So even though we get the same refCommand we have to interpret it differently based on the ref stage.

CMakeLists.txt
```diff
@@ -278,7 +278,7 @@ set(CONDITIONS_SOURCES
         ${PROJECT_SOURCE_DIR}/roboteam_ai/src/conditions/RefStateIsNormalPlay.cpp
         ${PROJECT_SOURCE_DIR}/roboteam_ai/src/conditions/RefStateIsNormalPlay.h
         ${PROJECT_SOURCE_DIR}/roboteam_ai/src/conditions/RefBallIsMoving.cpp
-        ${PROJECT_SOURCE_DIR}/roboteam_ai/src/conditions/RefBallIsMoving.h)
+        ${PROJECT_SOURCE_DIR}/roboteam_ai/src/conditions/RefBallIsMoving.h roboteam_ai/src/conditions/ResumePlayAfterPenalty.cpp roboteam_ai/src/conditions/ResumePlayAfterPenalty.h)
```
mrlukasbos: Project source dir

roboteam_ai/src/Switches.cpp
```diff
@@ -297,14 +301,18 @@ bt::Node::Ptr Switches::tacticSwitch(std::string name, bt::Blackboard::Ptr prope
 
     std::map<std::string, std::vector<std::pair<std::string, robotType>>> tactics = {
 
-            // Keeper tactics
+            // Keeper tactics. The rolename Keeper is hardcoded in our system, so use it!
```
mrlukasbos: oh no... this doesn't sound too good

roboteam_ai/src/utilities/StrategyManager.cpp
```diff
@@ -11,7 +11,16 @@ namespace rtt {
 namespace ai {
 
 // process ref commands
-void StrategyManager::setCurrentRefGameState(RefCommand command) {
+void StrategyManager::setCurrentRefGameState(RefCommand command, roboteam_msgs::RefereeStage stage) {
+    // if the stage is shootout, we interpret penalty commands as shootOut penalty commands
```
mrlukasbos: nice!

roboteam_ai/test/UtilTests/StrategyManagerTest.cpp
```diff
@@ -10,34 +10,54 @@
...
     // prepare command followed up by something else (i.e. command a) than normal start should trigger that command (command a).
-    strategyManager.setCurrentRefGameState(RefCommand::PREPARE_KICKOFF_US);
+    strategyManager.setCurrentRefGameState(RefCommand::PREPARE_KICKOFF_US,stage);
     EXPECT_EQ(strategyManager.getCurrentRefGameState().strategyName, "kickoff_us_formation_strategy");
 
-    strategyManager.setCurrentRefGameState(RefCommand::HALT);
+    strategyManager.setCurrentRefGameState(RefCommand::HALT,stage);
     EXPECT_EQ(strategyManager.getCurrentRefGameState().strategyName, "halt_strategy");
 
+    stage.stage=roboteam_msgs::RefereeStage::PENALTY_SHOOTOUT;
```
mrlukasbos: very nice

roboteam_ai/src/Switches.cpp
mrlukasbos: Okay after further reviewing I did not see the string 'keeper' anywhere, so i presume that there has been no behavior change or hardcoded dependency on the rolenome keeper. I presume you mean that the name 'keeper' needs to be that because it also needs to be that in trees? (like with all our roles)

PR #610, opened 2019-06-14
Feedback integration

mrlukasbos:
~Needs https://github.com/RoboTeamTwente/roboteam_msgs/pull/6 to compile~


roboteam_ai/src/io/IOManager.cpp
```diff
@@ -107,6 +107,15 @@ void IOManager::handleGeometryData(const roboteam_msgs::GeometryDataConstPtr &ge
 void IOManager::handleRobotFeedback(const roboteam_msgs::RobotFeedbackConstPtr &robotfeedback) {
```
rolfvdhulst: Is this a special type made for constPtrs from Ros or are you making a const ptr of a const ptr? No harm in making it explicit in c++ but why not just const ptr of RobotFeedback?

mrlukasbos: Its autogenerated from ROS

rolfvdhulst: I am just interested in what the difference between the following two would be:
const roboteam_msgs::RobotFeedbackConstPtr &robotfeedback
const roboteam_msgs::RobotFeedback &robotfeedback

PR #619, opened 2019-06-17
fixing some shootOut problems

rolfvdhulst:
Fixes some shootOut Problems found during testing.


roboteam_ai/src/skills/KickTo.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <roboteam_ai/src/control/ControlUtils.h>
+
+namespace rtt {
+namespace ai {
+
+KickTo::KickTo(string name, bt::Blackboard::Ptr blackboard)
+        :Skill(std::move(name), std::move(blackboard)) {
+}
+void KickTo::onInitialize() {
+    shootPos=properties->getVector2("shootPos");
```
ThijsLuttikhuis: never initialize an absolute position in the field.

this changes for fieldsize

roboteam_ai/src/skills/KickTo.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+/// Get an update on the skill
+bt::Node::Status KickTo::onUpdate() {
+    if (! robot) return Status::Running;
```
ThijsLuttikhuis: not needed. and it should not return running

ThijsLuttikhuis: this is done in skill.cpp

roboteam_ai/src/skills/KickTo.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+    if (world::field->pointIsInDefenceArea(ball->pos, false)) {
+        command.w = 0;
```
ThijsLuttikhuis: why 0?

rolfvdhulst: it defaults to that anyways.

PR #621, opened 2019-06-18
Feature/dont look at ball in ball handle

mrlukasbos:


roboteam_ai/src/world/Field.cpp
```diff
@@ -38,23 +38,20 @@ Vector2 Field::get_their_goal_center() {
...
-    float halfLength = _field.field_length*0.5f;
-    float halfWidth = _field.field_width*0.5f;
+bool Field::pointIsInField(const Vector2 &point, double margin) {
+    auto field = get_field();
 
-    return (point.x <= halfLength - margin &&
-            point.x >= - halfLength + margin &&
-            point.y <= halfWidth - margin &&
-            point.y >= - halfWidth + margin);
+    double halfLength = field.field_length*0.5f - margin;
```
mrlukasbos: the margin is now divided by two since it is only applied once! This thus looks misleading... The test does fail as well on the (literal) edge cases.

mrlukasbos:
@ThijsLuttikhuis what does your latest commit do? It breaks the tests while we should've actually merged this PR

roboteam_ai/src/control/ballHandling/DribbleBackwards.cpp
```diff
@@ -166,40 +166,58 @@ RobotCommand DribbleBackwards::sendTurnCommand() {
...
-    command.dribbler = 28;
+    command.dribbler = 31;
     command.angle = lockedAngle;
     command.vel = lockedAngle.toVector2(- maxVel);
 
+
     // check if the robot is still on the virtual line from ball->pos to the target
     if (control::ControlUtils::distanceToLine(robot->pos,
-            backwardsDribbleLine.first, backwardsDribbleLine.second) > errorMargin*2.5) {
+            backwardsDribbleLine.first, backwardsDribbleLine.second) > errorMargin*5) {
```
mrlukasbos: 5.0

PR #636, opened 2019-07-02
Feature/geneva state feedback

mrlukasbos:


roboteam_ai/src/io/IOManager.cpp
```diff
@@ -157,11 +164,14 @@ void IOManager::publishRobotCommand(roboteam_msgs::RobotCommand cmd) {
             auto robot = world::world->getRobotForId(cmd.id, true);
             if (robot) {
 
-                // this is a failcheck; the geneva state will only be turned if it is possible (which is checked in robot)
-                robot->setGenevaState(cmd.geneva_state);
-                cmd.geneva_state = robot->getGenevaState();
+                if (!robot->genevaStateIsDifferent(cmd.geneva_state) || !robot->genevaStateIsValid(cmd.geneva_state)) {
```
Barisimre: not sure if this is needed, but it is in demo, so probably not used

roboteam_ai/src/world/Robot.cpp
```diff
@@ -253,6 +278,20 @@ void Robot::setBatteryLow(bool batteryLow) {
     Robot::batteryLow = batteryLow;
 }
 
+/*
+ * Returns true when a feedback packet has been detected at most 2 second ago.
+ */
+bool Robot::hasRecentFeedback() {
+    return (lastReceivedFeedbackMoment > world->getTime() - 1.0);
```
Barisimre: magic number

Barisimre: Preferably internal system clock time is used, since this is safer on the network. This should work, but it's safer.

Barisimre:
Approved by @ThijsLuttikhuis and @rolfvdhulst

PR #674, opened 2019-07-27
Robocup/mrl

ThijsLuttikhuis:
All changes from the RoboCup!


mrlukasbos:
I think we should merge this branch to master, although the tests are failing. This branch represents the final version of the software which we used during the Robocup and most of it worked, giving the new team a proper version to fall back to.

PR #679, opened 2019-09-01
Experimental/replace ros

rolfvdhulst:
This PR replaced ROS with ZMQ using self-written publishers/subscribers defined in the repository 'roboteam_proto'. 

**Features**
- No ROS dependency anymore. Faster compile, more intuitive CLion support, cleaner CMakeLists
- MacOS support
- Increased performance
- More flexible messaging standard using Google Protobuffers

**Changes:**
- The structure of publish/subscribe remains the same. (some differences: each topic has its own TCP channel, no need anymore for a ros::spinOnce(), we just keep polling continuously in a separate thread, for more details see roboteam_proto) 
- RoboteamAI displays settings in the interface which it publishes to a 'settings' topic
- Roboteam Vision is deprecated. Field inversion/message conversion does not happen. (so as consequence also no inverses world in roboteam_world)
- Another consequence of removing vision: RoboteamWorld does not output us and them anymore, but outputs 'yellow' and 'blue'.
- Roboteam_utils has nice utilities for rotating a field. This is used in AI where we actually rotate the field if necessary.
- ros::time has been replace with std::chrono
- The performance of the interface has improved a lot
- Roboteam input is broken, together with demo functionality. 

**Dependencies**
(See roboteam_proto readme) 
zeromq, zmqpp, protobuf


src/world/WorldManager.cpp
```diff
@@ -0,0 +1,23 @@
...
+namespace rtt {
+namespace ai {
+namespace world {
+
+void WorldManager::setup() {
+}
+
+// spin ROS at 4x our tick rate
+void WorldManager::loop() {
+//    ros::Rate rate(4.0 * ai::Constants::TICK_RATE());
```
rolfvdhulst: Wait was this actual code we had? omfg.

src/world/WorldManager.cpp
```diff
@@ -0,0 +1,23 @@
+#include <include/roboteam_ai/utilities/Constants.h>
+#include "include/roboteam_ai/world/WorldManager.h"
```
rolfvdhulst: Might want to mark this file as not used/outdated?

src/utilities/GameStateManager.cpp
```diff
@@ -58,5 +68,10 @@ bool GameStateManager::canMoveOutsideField(int robotId) {
     }
     return true;
 }
+
+    Vector2 GameStateManager::getRefereeDesignatedPosition() {
+        auto designatedPos = rtt::ai::GameStateManager::getRefereeData().designated_position();
+        return Vector2(designatedPos.x()/1000, designatedPos.x()/1000);
```
rolfvdhulst: This is definitely not the way to do it; we should store it correctly within AI IMO.

src/utilities/GameStateManager.cpp
```diff
@@ -1,20 +1,24 @@
-#include <roboteam_ai/src/interface/api/Output.h>
-#include "GameStateManager.hpp"
+#include <include/roboteam_ai/Settings/Settings.h>
```
rolfvdhulst: Why do we need a lock here now?

src/treeinterp/JsonReader.cpp
```diff
@@ -24,9 +24,13 @@ std::string JsonReader::getFilePath(std::string name) {
         }
         smallPath.append(word + "/");
     }
-    //std::cout << "Smallpath: " << smallPath << std::endl;
+
+    std::cout << "looking for jsons at smallpath: " << smallPath << std::endl;
     // should be at /home/[user]/roboteamtwente/workspace/src/roboteam_ai/ ish right now
-    smallPath.append("roboteam_ai/src/jsons/" + name + ".json");
+
+    smallPath.append("../../roboteam_ai/src/jsons/" + name + ".json");
```
rolfvdhulst: This is a bit problematic, shouldn't have to backtrack the path but find a way to do it immediately.

src/world/FieldMessage.cpp
```diff
@@ -0,0 +1,215 @@
...
+  goalWidth = mm_to_m(sslFieldSize.goal_width());
+  goalDepth = mm_to_m(sslFieldSize.goal_depth());
+  boundaryWidth = mm_to_m(sslFieldSize.boundary_width());
+
+  for (roboteam_proto::SSL_FieldLineSegment line : sslFieldSize.field_lines()) {
+    FieldLineSegment newLine;
+    newLine.name = std::string(name_map[line.name()]);
+    newLine.begin = mm_to_m(line.p1());
+    newLine.end = mm_to_m(line.p2());
+    newLine.thickness = mm_to_m(line.thickness());
```
rolfvdhulst: Make these 5 lines a function.

src/world/FieldMessage.cpp
```diff
@@ -0,0 +1,215 @@
...
+      bottom_right_penalty_stretch = newLine;
+    }
+}
+
+float FieldMessage::mm_to_m(float scalar) {
+  return scalar/1000;
+}
+
+Vector2 FieldMessage::mm_to_m(Vector2 vector) {
+  return {vector.x/1000, vector.y/1000};
```
rolfvdhulst: Make these all 1000.0 to ensure they are floats.

src/world/FieldMessage.cpp
```diff
@@ -0,0 +1,215 @@
...
+}
+
+std::vector<FieldLineSegment> FieldMessage::getField_lines(){
+  return field_lines;
+}
+std::vector<FieldArc> FieldMessage::getField_arcs(){
+  return field_arcs;
+}
+
+void FieldMessage::invert() {
```
rolfvdhulst: You need to swap the left and right lines since they literally 'take eachother's place' when yous wap the field for example.

src/roboteam_ai.cpp
```diff
@@ -15,12 +16,6 @@ void runBehaviourTrees() {
     app.checkForShutdown();
 }
 
-void runWorld() {
```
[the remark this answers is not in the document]
mrlukasbos: IT does run in a separate thread still. This loop previously Made sure the ROS callbacks are called. With roboteam proto this works different: (every subscriber has its own polling thread)

src/world/FieldMessage.cpp
mrlukasbos: Yeah I made an issue about this

PR #685, opened 2019-09-20
Interface FPS logging

mrlukasbos:
You will need the QT Charts addon for this PR to work.

Linux: 
`sudo apt install libqt5charts5-dev`

macOS: It should already be pre-installed by having installed qt5 through brew


CMakeLists.txt
```diff
@@ -275,7 +275,7 @@ set(INTERFACE_SOURCES
         ${PROJECT_SOURCE_DIR}/src/interface/widgets/MainControlsWidget.cpp
         ${PROJECT_SOURCE_DIR}/src/interface/widgets/VisualizationSettingsWidget.cpp
         ${PROJECT_SOURCE_DIR}/src/interface/api/Toggles.cpp
-        ${PROJECT_SOURCE_DIR}/src/interface/widgets/RuleSetWidget.cpp)
+        ${PROJECT_SOURCE_DIR}/src/interface/widgets/RuleSetWidget.cpp src/interface/widgets/GraphWidget.cpp include/roboteam_ai/interface/widgets/GraphWidget.h)
```
rolfvdhulst: enter.

include/roboteam_ai/interface/widgets/GraphWidget.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace rtt {
+namespace ai {
+namespace interface {
+
+class GraphWidget : public QWidget {
+    Q_OBJECT
+private:
+    float seriesIndex = 0;
+    float fpsyMax = 0;
```
rolfvdhulst: I find the prefix fps confusing, since it's not clear what fps you're talking about (roboteam AI? the widget itself?). Maybe just use graphYMax or so?

src/ApplicationManager.cpp
```diff
@@ -35,32 +35,44 @@ void ApplicationManager::loop() {
...
-
             this->runOneLoopCycle();
+
+            std::chrono::milliseconds afterOneCycleTime = std::chrono::duration_cast< std::chrono::milliseconds >(
+                    std::chrono::system_clock::now().time_since_epoch());
+
+            cyclesInThisSecond++;
+
+            auto fps_diff = now_time - last_fps_count_time;
+            auto fps_timestep =std::chrono::milliseconds(200); // one second
```
rolfvdhulst: this code is a bit wonky on the first cycle making it slightly hard to understand. Not a huge issue though.

src/ApplicationManager.cpp
```diff
@@ -35,32 +35,44 @@ void ApplicationManager::loop() {
...
+            if(fps_diff > fps_timestep) {
+
+                std::cout << "FPS: " << cyclesInThisSecond << std::endl;
+                lastFPS = cyclesInThisSecond;
+                cyclesInThisSecond=0;
+                last_fps_count_time = now_time;
+            }
+            ai::interface::Input::setFps(lastFPS*5);
```
rolfvdhulst: If would only set it in the interface everytime we actually update the value (so within the if Statement). Now you're sending data that's really generated at 5hz at 60hz to the interface.

src/coach/defence/DefenceDealer.cpp
```diff
@@ -63,7 +63,7 @@ void DefenceDealer::updateDefenderLocations() {
     auto foundDefenders=g_defensivePositionCoach.decidePositions(lockedDefenders, freeDefenders);
     availableIDs.clear();
     if (foundDefenders.size()<freeDefenders.size()+lockedDefenders.size()){
-        std::cout<<"WTFF"<<std::endl;
+        std::cout<<"[DefenceDealer] Warning: There is no opponent to defend against"<<std::endl;
```
rolfvdhulst: This should probably be something like 'DefenceDealer could not find a valid position on the field for one or more robots", again not hugely important

src/interface/widgets/MainControlsWidget.cpp
```diff
@@ -11,19 +11,23 @@
...
 #include <utilities/GameStateManager.hpp>
 #include <include/roboteam_ai/Settings/Settings.h>
+#include <include/roboteam_ai/interface/api/Input.h>
 #include "utilities/GameState.h"
 
 namespace rtt {
 namespace ai {
 namespace interface {
 
+
```
rolfvdhulst: these spaces are not necessary.

src/interface/api/Input.cpp
```diff
@@ -44,14 +49,24 @@ const std::vector<Drawing> Input::getDrawings() {
...
-    std::lock_guard<std::mutex> lock(drawingMutex);
+    std::lock_guard<std::mutex> drawingLock(drawingMutex);
     drawings = {};
 }
 
 Input::~Input() {
     clearDrawings();
 }
 
+int Input::getFps() {
```
rolfvdhulst: This function name seems very contradictory... not sure what the class Input does exactly at the moment but this is definitely an output ;)

src/interface/widgets/mainWindow.cpp
```diff
@@ -142,8 +149,15 @@ MainWindow::MainWindow(QWidget* parent) : QMainWindow(parent) {
     connect(robotsTimer, SIGNAL(timeout()), this, SLOT(updateRobotsWidget())); // we need to pass the visualizer so thats why a seperate function is used
     connect(robotsTimer, SIGNAL(timeout()), mainControlsWidget, SLOT(updatePause()));
     connect(robotsTimer, SIGNAL(timeout()), mainControlsWidget, SLOT(updateContents()));
+    robotsTimer->start(200); // 5fps
+
+    auto * graphTimer = new QTimer(this);
+    connect(graphTimer, SIGNAL(timeout()), graphWidget, SLOT(updateContents()));
+    graphTimer->start(200); // 5fps
```
rolfvdhulst: You could for updating also trigger updates when a new FPS data is set or when the internal clock is ticking, that'd be a bit more clean I think. This is fine though.

src/ApplicationManager.cpp
rolfvdhulst: I would maybe make sure that this value (200 ms) is documented clearly somewhere, since we had issues with not detecting problems with 1 second intervals. I wouldn't mind this dropping to 100 ms or even 50 ms so problems are easier to see.

PR #686, opened 2019-09-23
Move world and field to local in Node.tick()

mrlukasbos:
Instead of calling the global variables world::world and world::field, all skills and conditions now relay on world and field that are passed in as parameters for the Tick() function.


rolfvdhulst:
Is this still WIP? You requested my review but there's no discription

include/roboteam_ai/bt/Node.hpp
```diff
@@ -4,8 +4,22 @@
 #include <vector>
 #include <cstdio>
 #include <cstdarg>
+#include <include/roboteam_ai/world/World.h>
```
rolfvdhulst: This include shouldn't be necessary right? Since every source file can use it if it has the include for world already.

src/skills/formations/StopFormation.cpp
```diff
@@ -199,25 +197,25 @@ std::vector<std::vector<Vector2>> StopFormation::getStopPositions() {
 }
 
 bool StopFormation::positionShouldBeAvoided(Vector2 pos) {
-    return (pos.dist(ball->pos) < 0.9 || !world::field->pointIsInField(pos, 0.0));
+    return (pos.dist(ball->pos) < 0.9 || !field->pointIsInField(pos, 0.0));
 }
 
 std::vector<Vector2> StopFormation::getProperPositions(int amount) {
     std::vector<Vector2> properPositions;
-    auto field = world::field->get_field();
```
rolfvdhulst: This function definitely shows why field needs a refactor (look at the amount of unnecesarry copying here) but that's outside of this pr's scope.

mrlukasbos:
> "there would be a static pointer which is just shared by all nodes which is only called for the root for example." 
I don't think I agree with that. Passing through the parameters through the functions gives more clarity about the origin of the variables. using globals or statics can give unexpected behavior. 

I do agree that it 'feels wrong' to give the 'tick()' functions inputs. For now I don't have a better solution than this.

rolfvdhulst:
I think in this context we can solve the passing on pointers at the same time we make AI's objects that work as stand-alones. Then you simply pass the AI object a world object (not a reference) with the current robot positions every frame and just do access internally through a static member value/pointers/inheritance, or something like that. For now, this is fine.

mrlukasbos:
@RobotJesse and @HaicoDorenbos I let it over to you to decide whether you want this or not (its quite an architectural decision)

src/skills/Keeper.cpp
```diff
@@ -133,13 +133,25 @@ void Keeper::setGoalPosWithAttacker(RobotPtr attacker) {
...
-    Vector2 endGoal = {- field.field_length()/2, field.goal_width()/2};
+    auto fieldMsg = field->get_field();
+    Vector2 startGoal = {- fieldMsg.field_length()/2, - fieldMsg.goal_width()/2};
+    Vector2 endGoal = {- fieldMsg.field_length()/2, fieldMsg.goal_width()/2};
     if (control::ControlUtils::lineSegmentsIntersect(start, end, startGoal, endGoal)) {
         goalPos = control::ControlUtils::twoLineIntersection(start, end, startGoal, endGoal);
     }
 }
 
+rtt::Arc Keeper::createKeeperArc() {
```
HaicoDorenbos: The static reference to world and field is not removed inside this newly added code

src/skills/Keeper.cpp
```diff
@@ -133,13 +133,25 @@ void Keeper::setGoalPosWithAttacker(RobotPtr attacker) {
...
+    double goalwidth = rtt::ai::world::field->get_field().goal_width();
```
HaicoDorenbos: The static reference to world and field is not removed inside this newly added code

RobotJesse:
Before we merge this, we will test it with grsim

PR #687, opened 2019-09-23
Synchronize world in behaviourtreeThread

mrlukasbos:
This pr is an attempt to get some more performance by properly synchronizing world inputs and decision making. It's not very stable yet but some the logs from the profiler look promising.

needs #685 
needs #686


CMakeLists.txt
```diff
@@ -36,9 +35,10 @@ endif()
 # for Linux, BSD, Solaris, Minix
 if(UNIX AND NOT APPLE)
     SET(Qt5Widgets_DIR "/usr/include/x86_64-linux-gnu/qt5/QtWidgets")
+    SET(Qt5Charts_DIR "/usr/include/x86_64-linux-gnu/qt5/QtCharts")
 endif()
 
-
+find_package(Qt5Charts REQUIRED)
```
RobotJesse: Does this require a new dependency? Do we need to update the installation guide?

mrlukasbos: No this dependency is already in development. As you can probably see, this PR still has conflicts with the base branch. (it needs another PR to be merged first, and i suspect most conflicts will be gone by then)

mrlukasbos:
I'll close this one since it is deprecated

PR #689, opened 2019-10-01
Refactor/control methods

Ratoone:
Refactored the control methods to inherit from a base controller. Classes that depend on the controllers still reference PID control - will be changed in the future, when integrating other controls.

As per @mrlukasbos suggestion, the branch has been cleared of old commits, resulting in a new branch (while deleting the old one). Also, the name of the new branch respects the branching name convention.

As per @HaicoDorenbos observation, javadoc has been added to the new classes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,7 +265,7 @@ set(TEST_SOURCES
 #
 #        ${PROJECT_SOURCE_DIR}/test/ControlTests/PosControllerTests/NumTreePosControlTest.cpp
 #        ${PROJECT_SOURCE_DIR}/test/ControlTests/PosControllerTests/RobotCommandTest.cpp
-        )
+          )
 
 set(INTERFACE_SOURCES
         ${PROJECT_SOURCE_DIR}/src/interface/widgets/mainWindow.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,17 +17,10 @@ set(CMAKE_AUTOUIC OFF)
 
 # for MacOS X or iOS, watchOS, tvOS (since 3.10.3)
 if(APPLE)
-    #Qt5Test_DIR=/usr/local/opt/qt/lib/cmake/Qt5Test
-    #Qt5Core_DIR=/usr/local/opt/qt/lib/cmake/Qt5Core
-    #Qt5_DIR=/usr/local/opt/qt/lib/cmake
-    #Qt5Concurrent_DIR=/usr/local/opt/qt/lib/cmake/Qt5Concurrent
-    #Qt5Gui_DIR=/usr/local/opt/qt/lib/cmake/Qt5Gui
-#    set(Qt5Gui_DIR "/usr/local/opt/qt/lib/cmake/Qt5Gui")
-#    set(Qt5_DIR "/usr/local/opt/qt/lib/cmake")
     set(Qt5Widgets_DIR "/usr/local/opt/qt5/lib/cmake/Qt5Widgets")
-
     set(Qt5Core_DIR "/usr/local/opt/qt/lib/cmake/Qt5Core")
     set(Qt5Gui_DIR "/usr/local/opt/qt/lib/cmake/Qt5Gui")
+    SET(Qt5Charts_DIR "/usr/local/opt/qt/lib/cmake/Qt5Charts")
 
     find_package(Qt5Core REQUIRED)
     find_package(Qt5Gui REQUIRED)
@@ -36,9 +29,10 @@ endif()
 # for Linux, BSD, Solaris, Minix
 if(UNIX AND NOT APPLE)
     SET(Qt5Widgets_DIR "/usr/include/x86_64-linux-gnu/qt5/QtWidgets")
+    SET(Qt5Charts_DIR "/usr/include/x86_64-linux-gnu/qt5/QtCharts")
 endif()
 
-
+find_package(Qt5Charts REQUIRED)
 find_package(Qt5Widgets REQUIRED)
 
 set(JSON_SOURCES
@@ -282,6 +276,8 @@ set(INTERFACE_SOURCES
         ${PROJECT_SOURCE_DIR}/src/interface/widgets/VisualizationSettingsWidget.cpp
         ${PROJECT_SOURCE_DIR}/src/interface/api/Toggles.cpp
         ${PROJECT_SOURCE_DIR}/src/interface/widgets/RuleSetWidget.cpp
+        ${PROJECT_SOURCE_DIR}/src/interface/widgets/GraphWidget.cpp
+        ${PROJECT_SOURCE_DIR}/src/interface/widgets/SettingsWidget.cpp
 
         # QT wants to know about these headers
         ${PROJECT_SOURCE_DIR}/include/roboteam_ai/interface/widgets/PidBox.h
@@ -293,6 +289,7 @@ set(INTERFACE_SOURCES
         ${PROJECT_SOURCE_DIR}/include/roboteam_ai/interface/widgets/TreeVisualizerWidget.h
         ${PROJECT_SOURCE_DIR}/include/roboteam_ai/interface/widgets/VisualizationSettingsWidget.h
         ${PROJECT_SOURCE_DIR}/include/roboteam_ai/interface/widgets/widget.h
+        ${PROJECT_SOURCE_DIR}/include/roboteam_ai/interface/widgets/SettingsWidget.h
         )
 
 set(WORLD_SOURCES
@@ -320,7 +317,7 @@ set(COACH_SOURCES
         ${PROJECT_SOURCE_DIR}/src/coach/defence/DefenceDealer.cpp
         ${PROJECT_SOURCE_DIR}/src/coach/defence/PossiblePass.cpp
         ${PROJECT_SOURCE_DIR}/src/coach/midField/MidFieldCoach.cpp
-        ${PROJECT_SOURCE_DIR}/src/coach/GetBallCoach.cpp src/interface/widgets/SettingsWidget.cpp include/roboteam_ai/interface/widgets/SettingsWidget.h)
+        ${PROJECT_SOURCE_DIR}/src/coach/GetBallCoach.cpp)
 
 # Main Executable
 add_executable(roboteam_ai
@@ -346,6 +343,7 @@ add_executable(roboteam_ai
 target_link_libraries(roboteam_ai
         PRIVATE Qt5::Widgets
         PRIVATE Qt5::Gui
+        PRIVATE Qt5::Charts
         PRIVATE roboteam_utils)
 
 target_include_directories(roboteam_ai
@@ -379,19 +377,20 @@ if(APPLE)
             PRIVATE roboteam_utils
             PRIVATE Qt5::Widgets
             PRIVATE Qt5::Gui
+            PRIVATE Qt5::Charts
             /usr/local/lib/libgtest.a
             /usr/local/lib/libgtest_main.a)
 endif()
 
 if(UNIX AND NOT APPLE)
 
-target_link_libraries(ai_tests
+  target_link_libraries(ai_tests
         PRIVATE roboteam_utils
         PRIVATE Qt5::Widgets
         PRIVATE Qt5::Gui
+        PRIVATE Qt5::Charts
         PRIVATE gcov)
 endif()
-
 
 target_include_directories(ai_tests
         INTERFACE include
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,7 +178,10 @@ set(CONDITIONS_SOURCES
 
 set(CONTROL_SOURCES
         ${PROJECT_SOURCE_DIR}/src/control/ControlUtils.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/pid.cpp
+
+        # Controllers
+        ${PROJECT_SOURCE_DIR}/src/control/controllers/PidController.cpp
+        ${PROJECT_SOURCE_DIR}/src/control/controllers/PidTwoAxesController.cpp
 
         # PositionControllers
         ${PROJECT_SOURCE_DIR}/src/control/BasicPosControl.cpp
@@ -197,8 +200,7 @@ set(CONTROL_SOURCES
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/DribbleBackwards.cpp
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/DribbleForwards.cpp
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateAroundBall.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateWithBall.cpp
-        )
+        ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateWithBall.cpp)
 
 set(TEST_SOURCES
         ${PROJECT_SOURCE_DIR}/test/main.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,7 +174,8 @@ set(CONDITIONS_SOURCES
         ${PROJECT_SOURCE_DIR}/src/conditions/RobotOutside.cpp
         ${PROJECT_SOURCE_DIR}/src/conditions/RefStateIsNormalPlay.cpp
         ${PROJECT_SOURCE_DIR}/src/conditions/RefBallIsMoving.cpp
-        ${PROJECT_SOURCE_DIR}/src/conditions/ResumePlayAfterPenalty.cpp)
+        ${PROJECT_SOURCE_DIR}/src/conditions/ResumePlayAfterPenalty.cpp
+        )
 
 set(CONTROL_SOURCES
         ${PROJECT_SOURCE_DIR}/src/control/ControlUtils.cpp
@@ -200,7 +201,8 @@ set(CONTROL_SOURCES
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/DribbleBackwards.cpp
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/DribbleForwards.cpp
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateAroundBall.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateWithBall.cpp)
+        ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateWithBall.cpp
+        )
 
 set(TEST_SOURCES
         ${PROJECT_SOURCE_DIR}/test/main.cpp
--- a/include/roboteam_ai/control/PosController.h
+++ b/include/roboteam_ai/control/PosController.h
@@ -7,7 +7,8 @@
 
 #include "RobotCommand.h"
 #include <utilities/Constants.h>
-#include <control/pid.h>
+#include <include/roboteam_ai/control/controllers/PidController.h>
+#include <include/roboteam_ai/control/controllers/PidTwoAxesController.h>
 
 namespace rtt {
 namespace ai {
@@ -32,13 +33,10 @@ class PosController {
         bool customCanMoveInDefenseArea = true;
 
         // PID functions
-        PID xpid = PID(0.0, 0.0, 0.0);
-        PID ypid = PID(0.0, 0.0, 0.0);
+        PidTwoAxesController pid = PidTwoAxesController();
         bool getPIDFromInterface = true;
-        RobotCommand controlWithPID(const RobotPtr &robot, RobotCommand target);
+        RobotCommand controlWithPID(const RobotPtr &robot, const RobotCommand& target);
         virtual void checkInterfacePID() = 0;
-
-        virtual Vector2 calculatePIDs(const RobotPtr &robot, RobotCommand &target);
 
     public:
         PosController() = default;
--- a/include/roboteam_ai/control/PosController.h
+++ b/include/roboteam_ai/control/PosController.h
@@ -7,8 +7,8 @@
 
 #include "RobotCommand.h"
 #include <utilities/Constants.h>
-#include <include/roboteam_ai/control/controllers/PidController.h>
-#include <include/roboteam_ai/control/controllers/PidTwoAxesController.h>
+#include <control/controllers/PidController.h>
+#include <control/controllers/PidTwoAxesController.h>
 
 namespace rtt {
 namespace ai {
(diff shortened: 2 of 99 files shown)


CMakeLists.txt
```diff
@@ -197,8 +200,7 @@ set(CONTROL_SOURCES
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/DribbleBackwards.cpp
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/DribbleForwards.cpp
         ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateAroundBall.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateWithBall.cpp
-        )
+        ${PROJECT_SOURCE_DIR}/src/control/ballHandling/RotateWithBall.cpp)
```
rolfvdhulst: Not huge, but please put keep the ) on the next line. This prevents merge conflicts when multiple people are working on different branches.

include/roboteam_ai/control/PosController.h
```diff
@@ -7,7 +7,8 @@
 
 #include "RobotCommand.h"
 #include <utilities/Constants.h>
-#include <control/pid.h>
+#include <include/roboteam_ai/control/controllers/PidController.h>
```
rolfvdhulst: You should never have to write include/roboteam_ai in your include's, you can remove it here (not that important but just convention)

src/control/PosController.cpp
```diff
@@ -80,8 +71,8 @@ void PosController::setAutoListenToInterface(bool listenToInterface) {
 
 void PosController::updatePid(pidVals pid) {
     if (lastPid != pid) {
-        xpid.setPID(pid);
-        ypid.setPID(pid);
+        this->pid = PidTwoAxesController(std::tuple_cat(pid,std::make_tuple(0)));
```
rolfvdhulst: This probably needs a comment that explains what the 0 and std::tuple_cat are doing. (Something short like //we always set F to 0)

include/roboteam_ai/control/controllers/PidTwoAxesController.h
```diff
@@ -0,0 +1,26 @@
...
+#define RTT_PIDTWOAXESCONTROLLER_H
+
+#include <include/roboteam_ai/control/ControlUtils.h>
+#include "BaseController.h"
+#include "PidController.h"
+
+class PidTwoAxesController : BaseController<Vector2,Vector2>{
+public:
+    PidController pidXAxis = PidController(0, 0, 0, 0);
+    PidController pidYAxis = PidController(0, 0, 0, 0);
```
rolfvdhulst: These probably do not need to be public.

include/roboteam_ai/control/controllers/PidTwoAxesController.h
```diff
@@ -0,0 +1,26 @@
...
+
+    PidTwoAxesController(std::tuple<double, double, double, double> pidXParams, std::tuple<double, double, double, double> pidYParams);
+
+    explicit PidTwoAxesController(std::tuple<double, double, double, double> pidParams);
+
+    Vector2 computeOutput(Vector2 setPointInput, Vector2 actualInput) override;
+
+    PidTwoAxesController() = default;
```
rolfvdhulst: Another style/opinion thing but for most files we put the constructors/destructors as one of the first functions

include/roboteam_ai/control/PosController.h
```diff
@@ -32,14 +33,11 @@ class PosController {
         bool customCanMoveInDefenseArea = true;
 
         // PID functions
-        PID xpid = PID(0.0, 0.0, 0.0);
-        PID ypid = PID(0.0, 0.0, 0.0);
+        PidTwoAxesController pid = PidTwoAxesController();
```
HaicoDorenbos: What is the relation between the PosController and PidTwoAxesController? It is unclear for me, so this variable needs a comment which explains this relation.

include/roboteam_ai/control/controllers/BaseController.h
```diff
@@ -0,0 +1,25 @@
...
+
+#ifndef RTT_BASECONTROLLER_H
+#define RTT_BASECONTROLLER_H
+
+template <typename Input, typename Output>
+/**
+ * The base abstract class for the controllers. Any control method should inherit from this
+ * @tparam Input
+ * @tparam Output
+ */
```
HaicoDorenbos: @author and @since are missing here

include/roboteam_ai/control/controllers/BaseController.h
```diff
@@ -0,0 +1,25 @@
+//
+// Created by ratoone on 25-09-19.
+//
+
+#ifndef RTT_BASECONTROLLER_H
+#define RTT_BASECONTROLLER_H
+
+template <typename Input, typename Output>
+/**
+ * The base abstract class for the controllers. Any control method should inherit from this
```
HaicoDorenbos: What is a controller and control method? This is still unclear for me when I read the documentation.

include/roboteam_ai/control/controllers/BaseController.h
```diff
@@ -0,0 +1,25 @@
...
+ * @tparam Input
+ * @tparam Output
```
HaicoDorenbos: The @param tag should not be used here, because the class does not use parameters only the function uses parameter.

include/roboteam_ai/control/controllers/PidTwoAxesController.h
```diff
@@ -0,0 +1,26 @@
...
+//
+
+#ifndef RTT_PIDTWOAXESCONTROLLER_H
+#define RTT_PIDTWOAXESCONTROLLER_H
+
+#include <include/roboteam_ai/control/ControlUtils.h>
+#include "BaseController.h"
+#include "PidController.h"
+
+class PidTwoAxesController : BaseController<Vector2,Vector2>{
```
HaicoDorenbos: The PidTwoAxesController class and the public functions do not have any documentations. Also it is not clear for me what the Pid means. Does it mean "Pelvic inflammatory disease"? This should also be clear from the documentation or you should use the full name instead of the abbreviation Pid.

include/roboteam_ai/control/controllers/PidTwoAxesController.h
```diff
@@ -0,0 +1,26 @@
...
+public:
+    PidController pidXAxis = PidController(0, 0, 0, 0);
```
HaicoDorenbos: Also not clear for me what these variables mean. But if you document the class then this might be better understandable.

src/control/PosController.cpp
mrlukasbos: You can maybe also check for the existence of a fourth variable in the tuple in the constructor and put it to 0 there.

include/roboteam_ai/control/controllers/PidTwoAxesController.h
HaicoDorenbos: I agree, however it is not a major issue.

HaicoDorenbos: These functions are public, so they should have documentation added as well.

src/control/controllers/PidTwoAxesController.cpp
```diff
@@ -0,0 +1,36 @@
+//
```
HaicoDorenbos: Documentation should be added in the header file not in the source file.

src/control/controllers/PidTwoAxesController.cpp
```diff
@@ -0,0 +1,36 @@
...
+// Created by ratoone on 25-09-19.
+//
+
+#include <control/controllers/PidTwoAxesController.h>
+
+/**
+ * Create a PID controller for two axes, with different parameters
+ * @param pidXParams a tuple with 4 parameters for X dir PID - proportional, integrative, derivative and feedforward gain
+ * @param pidYParams a tuple with 4 parameters for Y dir PID - proportional, integrative, derivative and feedforward gain
+ */
```
HaicoDorenbos: Documentation should be added in the header file not in the source file

src/control/controllers/PidTwoAxesController.cpp
```diff
@@ -0,0 +1,36 @@
...
+PidTwoAxesController::PidTwoAxesController(std::tuple<double, double, double, double> pidXParams, std::tuple<double, double, double, double> pidYParams){
+    pidXAxis = PidController(pidXParams);
+    pidYAxis = PidController(pidYParams);
+}
+
+/**
+ * Computes the PID output for the two axes
+ * @param setPointInput the setpoint expressed as a Vector2
```
HaicoDorenbos: This param documentation does not give me any more information. I indeed know that it is a setpoint, because the variable is called a setPoint and I know that it is a Vector2, because the type of this variable is Vector2

src/control/controllers/PidTwoAxesController.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * @param pidXParams a tuple with 4 parameters for X dir PID - proportional, integrative, derivative and feedforward gain
+ * @param pidYParams a tuple with 4 parameters for Y dir PID - proportional, integrative, derivative and feedforward gain
+ */
+PidTwoAxesController::PidTwoAxesController(std::tuple<double, double, double, double> pidXParams, std::tuple<double, double, double, double> pidYParams){
+    pidXAxis = PidController(pidXParams);
+    pidYAxis = PidController(pidYParams);
+}
+
+/**
+ * Computes the PID output for the two axes
```
HaicoDorenbos: What is PID?

src/control/controllers/PidTwoAxesController.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * @param setPointInput the setpoint expressed as a Vector2
+ * @param actualInput the real value expressed as a Vector2
```
HaicoDorenbos: Same as above, this param documentation does not make anything more clear.

src/control/shotControllers/ShotController.cpp
```diff
@@ -164,7 +164,7 @@ RobotCommand ShotController::moveStraightToBall(world::Robot robot,
     //auto newPid = interface::Output::getNumTreePid();
     auto newPidValues = interface::Output::getShotControllerPID();
     updatePid(newPidValues);
-    double pidOutput = pid.getOutput(err.length()*sign, 0);
+    double pidOutput = pid.computeOutput(err.length() * sign, 0);
```
HaicoDorenbos: Indeed there should be a space between the multiplication operator and after the multiplication operator :+1:

include/roboteam_ai/control/ballHandling/BallHandlePosControl.h
```diff
@@ -50,16 +49,14 @@ class BallHandlePosControl : public NumTreePosControl {
         int ticksNotMoving = 0;
 
         pidfVals pidfGoToBall = std::make_tuple(0.0, 0.0, 0.0, 1.5);
-        PID xGoToBallPID = PID(pidfGoToBall);
-        PID yGoToBallPID = PID(pidfGoToBall);
+        PidTwoAxesController goToBallPid = PidTwoAxesController(pidfGoToBall, pidfGoToBall);
```
Ratoone: @mrlukasbos The idea was to use dependency injection for the actual controller. So I have an interface (BaseController) which will be used instead of the implementation. For instance, we have here the PidTwoAxesController. Instead of it, maybe I want to use some other controller. Then I will just declare a BaseController<Vector2,Vector2>, and through some setter / in the constructor I can set which type of controller we want

mrlukasbos: I understand. Why not set the controller you want immediately? Now you are forced to use generics making your code significantly harder.

Btw, this is very much opinion based. Its your project now and if you want it like this then don't let me stop you from doing so :-)

include/roboteam_ai/control/controllers/PidTwoAxesController.h
rolfvdhulst: I disagree with this sentiment to be honest. It is googleable and anyone working on this code is very likely to know what it means (and actually does need to know in order to fully understand/edit the code), so this should be common knowledge.

src/control/controllers/PidTwoAxesController.cpp
rolfvdhulst: https://en.m.wikipedia.org/wiki/PID_controller

src/control/PosController.cpp
rolfvdhulst: That would be even prettier.

include/roboteam_ai/control/controllers/PidTwoAxesController.h
HaicoDorenbos: I think it is still better to add class documentation, because of the following reasons:
1. When I search for Pid Two Axes Controller I cannot find anything that directly explains how it works. 
2. Documentation of code is a major issue of the RoboTeam.
3. When you add documentation to code we can quickly using CTRL-Q check how a function works. 

If there is some online source which describes how something works then the URL to this source should be included in the documentation.

src/control/controllers/PidTwoAxesController.cpp
HaicoDorenbos: If this URL clearly explains how it works then I propose to add this URL to the documentation.

include/roboteam_ai/control/controllers/BaseController.h
RobotJesse: I don't think we need this much detail here, everyone in the control team should know what a controller is

Ratoone:
Updated with the requested changes

include/roboteam_ai/control/controllers/BaseController.h
```diff
@@ -0,0 +1,29 @@
...
+#define RTT_BASECONTROLLER_H
+
+template <typename Input, typename Output>
+/**
+ * The base abstract class for the controllers (e.g. velocity control, position control). Any control
+ * method (e.g. PID control) should inherit from this.
+ * @tparam Input the type of the input of the controller (reference type)
+ * @tparam Output the type of the output of the controller (plant input)
+ * @author Paul Vacariu
+ * @SInCE 25.09.2019
```
HaicoDorenbos: @since should be written in lowercase

HaicoDorenbos:
Level of documentation seems to me to be quite satisfactory, but I will still do a full check...

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+//
+
+#ifndef ROBOTEAM_AI_PID_H
+#define ROBOTEAM_AI_PID_H
+
+#include <tuple>
+#include "BaseController.h"
+
+/**
+ * The basic implementation of a PIDF (proportional - integrative - derivative
```
HaicoDorenbos: Why do you put - between the words and why do you have a line break here?

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+ * - feedback) controller. The actual controller accepts a float reference
+ * and outputs a float controller action. For more info check https://en.wikipedia.org/wiki/PID_controller
```
HaicoDorenbos: It is now more clear for me, because I know this class is not about the Pelvic Inflammatory Disease. Reference URL to the Wikipedia page which explains more is quite useful.

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+     * @param d Derivative gain. Responds quickly to large changes in error. Small values prevents P and I terms from causing overshoot.
+     */
+    PidController(double p, double i, double d);
+    /** Create a new PID object.
+     * @param p Proportional gain. Large if large difference between setpoint and target.
+     * @param i Integral gain.	Becomes large if setpoint cannot reach target quickly.
+     * @param d Derivative gain. Responds quickly to large changes in error. Small values prevents P and I terms from causing overshoot.
+     * @param f Feedforward gain. The value that will be multiplied with the input and added to the output
+     */
+    PidController(double p, double i, double d, double f);
```
HaicoDorenbos: Good job :+1: we should indeed know what the parameters are used for in the constructor

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+    /** Create a new PID object
+     * @oaram pid tuple with the three parameters of the PID
```
HaicoDorenbos: What is an oaram?

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+     */
+    explicit PidController(std::tuple<double, double, double> pid);
+    /** Create a new PID object
+     * @oaram pid tuple with the three parameters of the PID, plus feedforward gain
```
HaicoDorenbos: Like before, this is misspelled

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+     * @param max Maximum value of range
+     * @return
+     */
+    bool bounded(double, double, double);
+    /**
+     * To operate correctly, all PID parameters require the same sign,
+     * with that sign depending on the {@literal}reversed value
+     */
+    void checkSigns();
+    double P;
```
HaicoDorenbos: It is better to use the full names instead, like proportionalGain to make these variables more meaningfull

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+    double maxIOutput;
+    double maxError;
+    double errorSum;
+
+    double maxOutput;
+    double minOutput;
+
+    double setpoint;
+
+    double lastActual;
```
HaicoDorenbos: Some of these variables can be const

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+    /** Create a new PID object
+     * @oaram pid tuple with the three parameters of the PID
+     */
+    explicit PidController(std::tuple<double, double, double> pid);
+    /** Create a new PID object
+     * @oaram pid tuple with the three parameters of the PID, plus feedforward gain
+     */
+    explicit PidController(std::tuple<double, double, double, double> pidf);
+
+    void setPID(double, double, double);
```
HaicoDorenbos: Huh?! Where are the parameter names?

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+     * @param maximum. Units are the same as the expected output value
+     */
+    void setMaxIOutput(double maximum);
+    /**Specify a maximum output. If a single parameter is specified, the minimum is
+     * set to (-maximum).
+     * @param output
+     */
+    void setOutputLimits(double output);
+    /**
+     * Specify a maximum output.
```
HaicoDorenbos: I guess you also specify minimum output, so this should be included in the summary as well

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+     * @param minimum possible output value
+     * @param maximum possible output value
+     */
+    void setOutputLimits(double minimum, double maximum);
+    /** Set the operating direction of the PID controller
+     * @param reversed Set true to reverse PID output
+     */
+    void setDirection(bool reversed);
```
HaicoDorenbos: So if you do twice setDirection(true) you will move in the same direction as you did before?

include/roboteam_ai/control/controllers/PidController.h
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Resets the controller. this->erases the I term buildup, and removes D gain on the next loop.
+     */
+    void reset();
+    /**Set the maximum rate the output can increase per cycle.
+     * @param rate
```
HaicoDorenbos: Missing explanation

include/roboteam_ai/control/controllers/PidTwoAxesController.h
```diff
@@ -0,0 +1,96 @@
...
+     * @param pidYParams a tuple with 4 parameters for Y dir PID - proportional, integrative, derivative and feedforward gain
+     */
+    PidTwoAxesController(std::tuple<double, double, double, double> pidXParams, std::tuple<double, double, double, double> pidYParams);
+
+    /**
+     * Create a PID controller for two axes, with different parameters and zero feedforward
+     * @param pidXParams a tuple with 3 parameters for X dir PID - proportional, integrative, derivative
+     * @param pidYParams a tuple with 3 parameters for Y dir PID - proportional, integrative, derivative
+     */
+    PidTwoAxesController(std::tuple<double, double, double> pidXParams, std::tuple<double, double, double> pidYParams);
```
HaicoDorenbos: It is also possible in C++ to use default arguments and then you can remove this overloaded constructor and use the previous constructor for this case

src/control/controllers/PidController.cpp
```diff
@@ -0,0 +1,225 @@
...
+PidController::PidController(std::tuple<double, double, double> pid)
+        : PidController(std::get<0>(pid), std::get<1>(pid), std::get<2>(pid)) {}
+
+PidController::PidController(std::tuple<double, double, double, double> pidf)
+        : PidController(std::get<0>(pidf), std::get<1>(pidf), std::get<2>(pidf), std::get<3>(pidf)) {}
+
+
+//**********************************
+//Configuration functions
+//**********************************
```
HaicoDorenbos: In the original pid.cpp file we had more documentation here, why did you remove it?

src/control/controllers/PidController.cpp
```diff
@@ -0,0 +1,225 @@
...
+    setpoint=0;
+    lastActual=0;
+    firstRun=true;
+    reversed=false;
+    outputRampRate=0;
+    lastOutput=0;
+    outputFilter=0;
+    setpointRange=0;
+
+    P=p; I=i; D=d; F=f;
```
HaicoDorenbos: According to C++ policies you should never have multiple code lines at the same line, after a ; there should be a new line.

src/control/controllers/PidTwoAxesController.cpp
```diff
@@ -0,0 +1,46 @@
...
+        : PidTwoAxesController(pidParams, pidParams){}
+
+Vector2 PidTwoAxesController::computeOutput(Vector2 setPointInput, Vector2 actualInput) {
+    Vector2 outputCommand;
+    outputCommand.x = pidXAxis.computeOutput(setPointInput.x, actualInput.x);
+    outputCommand.y = pidYAxis.computeOutput(setPointInput.y, actualInput.y);
+    return outputCommand;
+}
+
+void PidTwoAxesController::setOutputLimits(double minimumX, double maximumX, double minimumY, double maximumY) {
```
HaicoDorenbos: Nice Facade pattern

include/roboteam_ai/control/controllers/PidController.h
Ratoone: Added an explanation to the summary taken from the code. Apparently the function is never used though

PR #692, opened 2019-10-16
Refactored & Documented the Ball and BallPossession classes

HaicoDorenbos:
Summarized, the main changes I made are:
1. Added Javadoc documentation to every class and public/private functions.
2. Added comments for every global variable and for unclear parts of the code
3. Turned public variables into private and added getters and setters
4. Improved the naming of variables
5. Removed out-commented code
6. Used constants instead of magic values
7. Split code in functions with single responsibility

Also after having discussions with Rolf, we decided to remove the following functions from the Ball class:
- updateDribbling
- updateBallModel

When reviewing this code it is important to look at:
- Whether the behavior of the code changed (which should not be the case, I also used grSim and everything seems to be unchanged)
- If the code is more readable for you, appropriately documented, did the technical debt decreased, are the Ball and BallPossession classes better maintainable and did the design of these classes improve, etc...

### DIFF
--- a/include/roboteam_ai/world/Ball.h
+++ b/include/roboteam_ai/world/Ball.h
@@ -78,7 +78,6 @@ class Ball {
         void updateBallAtRobotPosition(const Ball &oldBall, const WorldData &worldData);
 
     public:
-        Vector2 pos = Vector2();
         Vector2 vel = Vector2();
         static bool exists;
         bool visible = false;
--- a/include/roboteam_ai/world/Ball.h
+++ b/include/roboteam_ai/world/Ball.h
@@ -101,19 +101,19 @@ class Ball {
         void updateBall(const BallPtr &oldBall, const WorldData &worldData);
 
         /**
-         * Get the expected position of the ball.
+         * Get the current expected position of the ball.
          * @return A vector which represents this position.
          */
         const Vector2 &getPos() const;
 
         /**
-         * Set the new expected position of the ball.
+         * Set the new current expected position of the ball.
          * @param new_pos The new expected position of the ball.
          */
         void setPos(const Vector2 &new_pos);
 
         /**
-         * Get the expected velocity of the ball.
+         * Get the current expected velocity of the ball.
          * @return A vector which represents the direction and the magnitude of this velocity.
          */
         const Vector2 &getVel() const;
--- a/include/roboteam_ai/world/Ball.h
+++ b/include/roboteam_ai/world/Ball.h
@@ -78,9 +78,7 @@ class Ball {
         void updateBallAtRobotPosition(const Ball &oldBall, const WorldData &worldData);
 
     public:
-        Vector2 vel = Vector2();
         static bool exists;
-        bool visible = false;
 
         /**
          * Create an uninitialized ball (where the expected position and expected velocity are not set yet).
--- a/include/roboteam_ai/world/Ball.h
+++ b/include/roboteam_ai/world/Ball.h
@@ -1,7 +1,3 @@
-//
-// Created by thijs on 1-4-19.
-//
-
 #ifndef ROBOTEAM_AI_BALL_H
 #define ROBOTEAM_AI_BALL_H
 
@@ -15,41 +11,133 @@ namespace world {
 
 class WorldData;
 class Robot;
+
+/**
+ * Represents and simulates the expected state of the Ball.
+ * @author Created by: Thijs Luttikhuis <br>
+ *         Documented by: Haico Dorenbos
+ * @since 2019-04-01
+ */
 class Ball {
+    private:
+        /* If the closest distance between the ball and our robots is smaller than this value then the ball is
+         * considered to be close to that robot. */
+        const float THRESHOLD_ROBOT_CLOSE_TO_BALL = 0.5;
+
+        /* The movement friction during simulation and real life are different, because the simulation does not model
+         * everything. So the movement friction has to be adjusted to compensate for this difference. */
+        const float SIMULATION_FRICTION = 1.22; //The expected movement friction of the ball during simulation.
+        const float REAL_FRICTION = 0.61; //The expected movement friction during real life performance.
+
+        // The maximum possible factor used for the Kalman filter which is used when filtering the velocity.
+        const float FILTER_MAX_FACTOR_FOR_VELOCITY = 0.8;
+        /* The smallest velocity used for the Kalman filter for which we have that the factor is equal
+         * THRESHOLD_ROBOT_CLOSE_TO_BALL. Larger velocities will also have THRESHOLD_ROBOT_CLOSE_TO_BALL as factor. */
+        const float FILTER_VELOCITY_WITH_MAX_FACTOR = 8.0;
+        /* The threshold used for the filtered velocity, if it exceeds this value then we use the expectedVelocity
+        instead as estimation for the filtered velocity. More information can be found in the comment in the
+        filterBallVelocity method of the Ball.cpp file. */
+        const float MAXIMUM_FILTER_VELOCITY = 100.0;
+
     public:
         using BallPtr = std::shared_ptr<Ball>;
         using RobotPtr = std::shared_ptr<Robot>;
 
     private:
-        bool ballInAir = false;
-        bool collidesNow = false;
-        bool kickedNow = false;
-        bool dribbledNow = false;
-        double lastKickVel = 0;
-        int ballStraightTicks = 0;
-        Vector2 ballStillPosition = Vector2();
+        Vector2 expectedPosition = Vector2(); // The expected position where this ball is.
+        const Vector2 expectedVelocity = Vector2(); // The expected velocity of the ball.
+        /* If the ball is visible by any camera. This value is true if the ball is visible by any camera and false
+         * otherwise */
+        bool visibleByAnyCamera = false;
+        // The position where the ball is expected to end (lay still) after following his path.
+        Vector2 expectedBallEndPosition = Vector2();
+        // The velocity but then adjusted to determine a more realistic expectedBallEndPosition
+        Vector2 filteredVelocity = Vector2();
 
-        void updateDribbling(const Ball &oldBall, const WorldData &worldData);
-        Robot* getDribblingRobot(const std::vector<RobotPtr> &robots, double maxDribbleRange);
-        void updateBallModel(const Ball &oldBall, const WorldData &worldData);
-        void updateBallPosition(const Ball &oldBall, const WorldData &worldData);
-        void updateExpectedPositionWhereBallIsStill(const Ball &oldBall, const WorldData &worldData);
+        /**
+         * When the position of the ball is not yet initialized, but was previously near to a robot then initialize
+         * the position of the ball to a position close to that robot.
+         */
+        void initBallAtRobotPosition(const Ball &oldBall, const WorldData &worldData);
+
+        /**
+         * Adds a moving average over the Kalman filter to make the ball-velocity more stable (this method could be
+         * moved to rtt_world).
+        */
+        void filterBallVelocity(Ball &oldBall, const WorldData &worldData);
+
+        /**
+         * Update the expected position where the ball will end (lay still) after following his path.
+         */
+        void updateExpectedBallEndPosition(const Ball &oldBall, const WorldData &worldData);
+
+        /**
+         * When the position of the ball is close to a robot and not visible by camera then update the ball according
+         * to the movement and rotation of that robot (which happens during dribbling).
+         */
+        void updateBallAtRobotPosition(const Ball &oldBall, const WorldData &worldData);
 
     public:
-        Ball();
-        // Ball(const Ball &copy) = default;
-        explicit Ball(const roboteam_proto::WorldBall &copy);
-        void updateBall(const BallPtr &oldBall, const WorldData &worldData);
-
-        const Vector2 &getBallStillPosition() const;
         Vector2 pos = Vector2();
         Vector2 vel = Vector2();
-        Vector2 acc = Vector2();
-        Vector2 filteredVel = Vector2();
-        double spin = 0.0;
         static bool exists;
         bool visible = false;
-        void filterBallVelocity(Ball &oldBall, const WorldData &worldData);
+
+        /**
+         * Create an uninitialized ball (where the expected position and expected velocity are not set yet).
+         */
+        Ball();
+
+        /**
+         * Create a Ball object with the current data about the ball.
+         * @param copy The current data about the ball
+         */
+        explicit Ball(const roboteam_proto::WorldBall &copy);
+
+        /**
+         * Update the expected end (lay still) position of the ball after the ball has followed his path. Moreover
+         * this method updates the expected position of the ball if the ball is close to a robot (which happens during
+         * dribbling).
+         * @param oldBall The old state of the ball
+         * @param worldData The current data about the world
+         */
+        void updateBall(const BallPtr &oldBall, const WorldData &worldData);
+
+        /**
+         * Get the expected position of the ball.
+         * @return A vector which represents this position.
+         */
+        const Vector2 &getPos() const;
+
+        /**
+         * Set the new expected position of the ball.
+         * @param new_pos The new expected position of the ball.
+         */
+        void setPos(const Vector2 &new_pos);
+
+        /**
+         * Get the expected velocity of the ball.
+         * @return A vector which represents the direction and the magnitude of this velocity.
+         */
+        const Vector2 &getVel() const;
+
+        /**
+         * Check if the ball is visible by any camera.
+         * @return True if the ball is visible by any camera, false otherwise.
+         */
+        bool getVisible();
+
+        /**
+         * Set the new visibility of the ball by any camera.
+         * @param visible True if the ball is visible by any camera, false otherwise.
+         */
+        void setVisible(bool visible);
+
+        /**
+         * Get the expected position where the ball will end (lay still) after following his path.
+         * @return A vector which represents this position.
+         */
+        const Vector2 &getExpectedBallEndPosition() const;
 };
 
 }
--- a/include/roboteam_ai/world/Ball.h
+++ b/include/roboteam_ai/world/Ball.h
@@ -34,7 +34,7 @@ class Ball {
         /* The smallest velocity used for the Kalman filter for which we have that the factor is equal
          * THRESHOLD_ROBOT_CLOSE_TO_BALL. Larger velocities will also have THRESHOLD_ROBOT_CLOSE_TO_BALL as factor. */
         const float FILTER_VELOCITY_WITH_MAX_FACTOR = 8.0;
-        /* The threshold used for the filtered velocity, if it exceeds this value then we use the expectedVelocity
+        /* The threshold used for the filtered velocity, if it exceeds this value then we use the velocity
         instead as estimation for the filtered velocity. More information can be found in the comment in the
         filterBallVelocity method of the Ball.cpp file. */
         const float MAXIMUM_FILTER_VELOCITY = 100.0;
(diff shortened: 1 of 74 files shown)


src/world/BallPossession.cpp
```diff
@@ -69,19 +70,18 @@ bool BallPossession::teamCloseToBall(const world::WorldData &world, bool ourTeam
...
-/// return true if given team is relatively far from ball
 bool BallPossession::teamFarFromBall(const world::WorldData &world, bool ourTeam) {
     if (world.ball) {
-        double farThreshHoldDist = 0.4;
+        double farThreshHoldDist = STANDARD_FAR_THRESHOLD;
 
-        if (!ourTeam) {
-            // if the ball is on our side, go more defensive.
-            farThreshHoldDist = world.ball->pos.x < 0.0 ? 0.9 : 0.4;
+        if (!ourTeam && world.ball->getPos().x < this->MIDDLE_LINE_X) {
```
rolfvdhulst: Personally I think having a constant for 0 is a bit overkill, but it does no harm.

src/world/FieldMessage.cpp
```diff
@@ -115,6 +116,10 @@ FieldLineSegment FieldMessage::getBottom_line(){
 FieldLineSegment FieldMessage::getLeft_line(){
   return left_line;
 }
+double FieldMessage::getLeftLineX() {
```
rolfvdhulst: Functionality like this is probably better to do in the field class itself.

src/world/Ball.cpp
```diff
@@ -16,138 +13,72 @@ namespace world {
...
+          visibleByAnyCamera(false) {
+}
 
 Ball::Ball(const roboteam_proto::WorldBall &copy)
-        :pos(copy.pos()), vel(copy.vel()), filteredVel(copy.vel()),
-         visible(copy.visible()) {
+        : expectedPosition(copy.pos()), expectedVelocity(copy.vel()), filteredVelocity(copy.vel()),
+          visibleByAnyCamera(copy.visible()) {
     exists = exists || copy.area() || Vector2(copy.pos()).isNotNaN();
     if (! exists) std::cout << "BallPtr message has existence = 0!!" << std::endl;
```
rolfvdhulst: This should be refactored (together with existance as a concept)

rolfvdhulst: For now it's ok, since this probably requires changes in roboteam_proto and roboteam_world too

src/control/ballHandling/DribbleForwards.cpp
```diff
@@ -162,12 +162,6 @@ RobotCommand DribbleForwards::sendDribbleForwardsCommand() {
     interface::Input::drawData(interface::Visual::BALL_HANDLING, {forwardsDribbleLine.first, forwardsDribbleLine.second},
             Qt::white, robot->id, interface::Drawing::LINES_CONNECTED);
 
-    // check if the ball is not too far right or too far left of the robot, and try to compensate for that
-    if (ball->visible && false) {
-        Angle ballAngleRelativeToRobot = (ball->pos - robot->pos).toAngle() - robot->angle;
-        command.vel += (robot->angle + M_PI_2).toVector2(ballAngleRelativeToRobot*0.23);
-    }
-
     // limit velocity close to the target
```
rolfvdhulst: Did you talk to Thijs about this change? I'm fine with it because this seems like a hack, but it's definitely a functionality change.

src/control/ballHandling/DribbleBackwards.cpp
```diff
@@ -218,12 +218,6 @@ RobotCommand DribbleBackwards::sendDribbleBackwardsCommand() {
     interface::Input::drawData(interface::Visual::BALL_HANDLING, {backwardsDribbleLine.first, backwardsDribbleLine.second},
             Qt::white, robot->id, interface::Drawing::LINES_CONNECTED);
 
-    // check if the ball is not too far right or too far left of the robot, and try to compensate for that
-    if (ball->visible && false) {
-        Angle ballAngleRelativeToRobot = (ball->pos - robot->pos).toAngle() - robot->angle;
-        command.vel += (robot->angle + M_PI_2).toVector2(ballAngleRelativeToRobot);
-    }
-
```
rolfvdhulst: Same thing, discussed this with thijs? I do not know the reason for this code, or why you removed it.

include/roboteam_ai/world/BallPossession.h
```diff
@@ -11,33 +7,89 @@
...
+     * their possession. */
+    const double THEIR_POSSESSION_RELATIVE_WIDTH_THRESHOLD = 0.375; // 3/8 of the field
+
 public:
     enum Possession {
-        LOOSEBALL,
-        OURBALL,
-        THEIRBALL,
-        CONTENDEDBALL
+        LOOSEBALL,      // If neither us nor the opponent possesses the ball.
```
rolfvdhulst: it's spelled possess in all cases here.

rolfvdhulst (changes requested):
Generally looks good, I like the const reference access functions especially. There's however some mistakes and odd changes you made that change functionality (see comments). I have some opinions about style, but those are a matter of opinion indeed.

include/roboteam_ai/world/BallPossession.h
rolfvdhulst: (many places in the file)

include/roboteam_ai/world/Ball.h
```diff
@@ -15,41 +11,130 @@ namespace world {
...
-        Vector2 filteredVel = Vector2();
-        double spin = 0.0;
-        static bool exists;
-        bool visible = false;
-        void filterBallVelocity(Ball &oldBall, const WorldData &worldData);
+        /**
+         * Get the expected position of the ball.
+         * @return A vector which represents this position.
+         */
+        const Vector2 &getPos() const;
```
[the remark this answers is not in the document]
RobotJesse: I think when something is performing the functionality of getting a variable, the method name should be getThing(). This is helpful because if you want to "get" a property, then you can just type "get" and you autocomplete will show you all of the get methods in once place.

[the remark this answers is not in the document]
mrlukasbos: Agree with @RobotJesse on this one.

src/world/BallPossession.cpp
```diff
@@ -29,10 +28,13 @@ void BallPossession::recomputeState() {
     bool weAreFar = farFromUsTime > FAR_TIME_TRESHOLD;
     bool theyAreFar = farFromThemTime > FAR_TIME_TRESHOLD;
 
-    if ((weAreClose && ! theyAreClose) || (world::world->getBall()->pos.x > world::field->get_field().field_length()/4.0)) {
+    FieldMessage field = world::field->get_field();
+    double ourPossessionX = field.getLeftLineX() + OUR_POSSESSION_RELATIVE_WIDTH_THRESHOLD * field.field_width();
```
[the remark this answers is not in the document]
mrlukasbos: Very shady indeed

include/roboteam_ai/world/BallPossession.h
```diff
@@ -11,33 +7,89 @@
...
+ * @author Created by: Rolf van der Hulst <br>
+ *         Documented by: Haico Dorenbos
+ * @since 2019-04-15
+ */
 class BallPossession {
     FRIEND_TEST(BallPossessionTest, team_far_or_close_to_ball);
     FRIEND_TEST(BallPossessionTest, it_properly_computes);
+
+private:
+    const double MIDDLE_LINE_X = 0.0; // The x coordinate that corresponds to the middle of the field.
```
mrlukasbos: I think this goes a little far

src/control/ballHandling/DribbleForwards.cpp
mrlukasbos: Its definitely not a functionality change since if(false) does not do anything ;-)

src/control/numTrees/Collision.cpp
```diff
@@ -72,7 +72,7 @@ void Collision::setGoalCollision(const Vector2 &collisionPos, double distance) {
 
 const Vector2 Collision::collisionPosition() const {
     if (collisionRobot->id != - 1) return collisionRobot->pos;
-    else if (collisionBall->visible) return collisionBall->pos;
+    else if (collisionBall->getVisible()) return collisionBall->getPos();
```
mrlukasbos: getVisible now gets the value of 'visiblyByAnyCamera' right? maybe keep the names consistent?

mrlukasbos (commented):
I generally don't see the use of documenting getters and setters, because if you give proper variable names ever programmer knows exactly what it means. 

@rolfvdhulst  already pointed out a major functionality issue.

Furthermore I think this is a very good refactor! Well done :)

src/world/BallPossession.cpp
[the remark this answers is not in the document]
HaicoDorenbos: I now changed field_width to field_length (it was being updated in a newer version of the code which I did not commit), but it should not be 5/8th the field but 6/8th the field. Since the field length is measured from the left part of the field (negative x coordinate) to the right part of the field (positive x coordinate). So field_length/4.0 is half in between the middle line and the left line and you measure this difference from the middle line which is at 1/2 of the field_length. Therefore it should be 6/8 = 3/4 = 1/2 + 1/4

include/roboteam_ai/world/Ball.h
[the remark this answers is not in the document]
Ratoone: I also agree with the getter naming. However, why not use the full name (getPosition, getVelocity)?

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -122,18 +122,18 @@ world::WorldData DefencePositionCoach::removeBotFromWorld(world::WorldData world
     return world;
 }
 Vector2 DefencePositionCoach::getMostDangerousPos(const world::WorldData &world) {
-    if (world.ball->vel.length()>0.5){
-        return world.ball->pos+world.ball->vel*0.3;
+    if (world.ball->getVel().length() > 0.5){
+        return world.ball->getPos() + world.ball->getVel() * 0.3;
```
Ratoone: What does 0.5 and 0.3 mean?

src/world/BallPossession.cpp
Ratoone: In this case zero actually means something. What if somebody wants sometime to change the reference system to start from the corner of the field instead of the center? I would even go one step further and put this into the constants, while also looking for the places where it is used.

rolfvdhulst: Whilst I don´t mind the variable for reading clarity, I think the argument you give is weak. There is virtually no reason for ever changing the reference system within our AI (which has some nice unique properties, like being symmetric around zero and being consistent with the rest of SSL software). It would probably be more advised to simply change a different reference system back to the one our AI accepts, if we need to use one at some point.

src/conditions/TwoRobotBallPlacement.cpp
```diff
@@ -22,7 +22,7 @@ bt::Node::Status TwoRobotBallPlacement::onUpdate() {
     int minimumRequiredRobotsInField = robotDealer::RobotDealer::keeperExistsInWorld() ? 3 : 2;
     bool weHaveEnoughRobots = us.size() >= minimumRequiredRobotsInField;
     //TODO: THIS REMOVES TWOROBOTBALLPLACEMENT (15.1)
-    bool ballIsCloseToBallPlacementPos = ballPlacementPos.dist(ball->pos) < 15.1;
+    bool ballIsCloseToBallPlacementPos = ballPlacementPos.dist(ball->getPos()) < 15.1;
```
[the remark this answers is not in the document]
rolfvdhulst: Although these are issues I would argue they are not relevant to this refactor (Haico did not make any changes except for ball class related things here), and should not be done here since they are clearly seperate issues. 

(Note this specific line is a robocup hack, see the TODO).

include/roboteam_ai/world/BallPossession.h
```diff
@@ -11,33 +7,89 @@
...
+    /* If the ball has been uninterruptedly close to a team for this number of seconds then this team is considered to
+     * be long enough close to the ball. */
+    const double CLOSE_TIME_TRESHOLD = 0.05;
+    /* If the ball has been uninterruptedly far from a team for this number of seconds then this team is considered
+     * to be long enough far from the ball. */
+    const double FAR_TIME_TRESHOLD = 1.5;
+    /* If the ball x coordinate is larger than relatively OUR_POSSESSION_RELATIVE_WIDTH_THRESHOLD of the width of the
+     * field (with respect to the leftmost x-coordinate of the field) then the ball is considered to be always in our
+     * possession. */
+    const double OUR_POSSESSION_RELATIVE_WIDTH_THRESHOLD = 0.75; // 6/8 of the field
```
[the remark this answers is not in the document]
rolfvdhulst: Nvm, it's actually correct

rolfvdhulst (approved):

include/roboteam_ai/world/Ball.h
```diff
@@ -15,41 +11,130 @@ namespace world {
...
+        void updateExpectedBallEndPosition(const Ball &oldBall, const WorldData &worldData);
+
+        /**
+         * When the position of the ball is close to a robot and not visible by camera then update the ball according
+         * to the movement and rotation of that robot (which happens during dribbling).
+         */
+        void updateBallAtRobotPosition(const Ball &oldBall, const WorldData &worldData);
 
     public:
+        static bool exists;
```
Ratoone: Should this be static?

include/roboteam_ai/world/Ball.h
```diff
@@ -15,41 +11,130 @@ namespace world {
...
+        /**
+         * Update the expected position where the ball will end (lay still) after following his path.
+         */
+        void updateExpectedBallEndPosition(const Ball &oldBall, const WorldData &worldData);
+
+        /**
+         * When the position of the ball is close to a robot and not visible by camera then update the ball according
+         * to the movement and rotation of that robot (which happens during dribbling).
+         */
+        void updateBallAtRobotPosition(const Ball &oldBall, const WorldData &worldData);
```
Ratoone: These methods depend on another Ball, but we are already in the Ball class. Is the current Ball the oldBall or not?

src/world/Ball.cpp
```diff
@@ -16,138 +13,72 @@ namespace world {
...
 Ball::Ball()
-        :pos(Vector2()), vel(Vector2()), filteredVel(Vector2()), visible(false) { }
+        : expectedPosition(Vector2()), expectedVelocity(Vector2()), filteredVelocity(Vector2()),
+          visibleByAnyCamera(false) {
+}
 
 Ball::Ball(const roboteam_proto::WorldBall &copy)
-        :pos(copy.pos()), vel(copy.vel()), filteredVel(copy.vel()),
-         visible(copy.visible()) {
+        : expectedPosition(copy.pos()), expectedVelocity(copy.vel()), filteredVelocity(copy.vel()),
```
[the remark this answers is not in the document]
Ratoone: That is the method from the WorldBall class, now the Ball class

include/roboteam_ai/world/BallPossession.h
HaicoDorenbos: The middle line position is a magic value, we could for example also decide that our left line is at x = 0.0. I indeed agree that we still have some kind of code cloning in this case, but it is much better than in the previous case.

src/coach/defence/DefencePositionCoach.cpp
HaicoDorenbos: When reviewing a pull request you should look at the changes made and at the changes that should be made.  My pull request is only about refactoring of the Ball and BallPossession classes, not about the DefencePositionCoach. And because I did not introduce/create these magic values you should not have a remark about them in this pull request.

include/roboteam_ai/world/Ball.h
rolfvdhulst: Yes. Don't ask. As I mentioned above, this is an ugly hack that should be remove, but it requires changes in roboteam_world/roboteam_proto to fix aswell.

include/roboteam_ai/world/Ball.h
```diff
@@ -15,41 +11,130 @@ namespace world {
...
-        double lastKickVel = 0;
-        int ballStraightTicks = 0;
-        Vector2 ballStillPosition = Vector2();
-
-        void updateDribbling(const Ball &oldBall, const WorldData &worldData);
-        Robot* getDribblingRobot(const std::vector<RobotPtr> &robots, double maxDribbleRange);
-        void updateBallModel(const Ball &oldBall, const WorldData &worldData);
-        void updateBallPosition(const Ball &oldBall, const WorldData &worldData);
-        void updateExpectedPositionWhereBallIsStill(const Ball &oldBall, const WorldData &worldData);
+        Vector2 expectedPosition = Vector2(); // The expected position where this ball is.
```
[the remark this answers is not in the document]
HaicoDorenbos: The expectedPosition likewise the other values are all set in the constructor, so this is not an issue.

include/roboteam_ai/world/BallPossession.h
```diff
@@ -11,33 +7,89 @@
...
 public:
     enum Possession {
-        LOOSEBALL,
-        OURBALL,
-        THEIRBALL,
-        CONTENDEDBALL
+        LOOSEBALL,      // If neither us nor the opponent possesses the ball.
+        OURBALL,        // If we possesses the ball.
+        THEIRBALL,      // If the opponent possesses the ball.
+        CONTENDEDBALL   // If we and the opponent both possesses the ball.
```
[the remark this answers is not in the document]
HaicoDorenbos: I do not agree: contended means challenged according to the Cambridge dictionary. And because the meaning is well documented this should not be a problem.

src/world/BallPossession.cpp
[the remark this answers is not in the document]
HaicoDorenbos: FieldMessages are passed by reference which is quite efficient. Also the number of calls to get_field() is now reduced by 1, which should have increased the speed of our code.

src/world/FieldMessage.cpp
HaicoDorenbos: I do not agree, I want to use the FieldMessage class to compute direct things about the Field (such as sizes, line positions and coordinates) and the Field class itself to compute indirect things about the Field (whether a point is left of a line).

include/roboteam_ai/world/BallPossession.h
HaicoDorenbos: This should now be fixed

src/control/numTrees/Collision.cpp
HaicoDorenbos: getVisibleByAnyCamera() is a very long method name. I decided to make the function names of the getters and setters short and the actual variable names long, to keep the right balance between the size of the code and the understandability of the code.

include/roboteam_ai/world/Ball.h
HaicoDorenbos: updateBall() is a function that is once called for every new ball instance where it gets updated with respect to the state of the old ball. Thus the current ball is not the old ball.

HaicoDorenbos:
Every remark is either fixed, or I put down a comment why I did not agree with your changes.

src/world/BallPossession.cpp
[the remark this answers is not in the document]
rolfvdhulst: I'm 99.9% sure that what happens here is that the FieldMessage is being copied from the Field class (which has it's own FieldMessage type, look at the getter), which is quite inefficient and copies a lot of data. It's definitely not passed by reference (look at the code, the global instance that stores the FieldMessage is accessed by reference). The accessing of FieldMessage itself in line 32 could be pass by reference, but that's not the major issue I was talking about.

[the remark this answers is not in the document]
rolfvdhulst: But this issue is not relevant for now in any way, so I suggest it's dropped and refactored in a later PR.

Ratoone (approved):

RobotJente (approved):
Good job!

PR #694, opened 2019-10-24
Use macOS as VM for travis (repairs CI) 

mrlukasbos:
This PR fixes our problems with Continuous Integration (CI). After we moved to protobuf this was not working properly on the Linux VM. But I noticed that on a macOS installation this was way easier because we can just use homebrew for almost everything. 

Also, with roboteam_suite our caching is even better and with small changes the full build takes less than 4.5 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ script:
   - git clone https://github.com/RoboTeamTwente/roboteam_suite roboteam_suite --recurse-submodules
   - cd ~/roboteam_suite
   - rm -rf roboteam_ai
-  - rm -rf roboteam_world
   - git submodule foreach git checkout master
   - mv ${TRAVIS_BUILD_DIR} .
   - mkdir cmake-build-debug && cd cmake-build-debug;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,66 @@
-sudo: required
-services:
-- xvfb
+language: cpp
+os: osx
+osx_image: xcode11.2
+
+addons:
+  homebrew:
+    packages:
+      - zmq
+      - zmqpp
+      - protobuf
+      - armadillo
+      - ccache
+      - gtest
+
 notifications:
   email: false
+
 cache:
-- ccache
+  directories:
+    - $HOME/Library/Caches/Homebrew
 
-language: cpp
-matrix:
-  include:
-  - name: "Xenial"
-    dist: xenial
-
-before_install:
-- sudo apt-get clean
-- sudo apt update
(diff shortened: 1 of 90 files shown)


RobotJesse:
So if I understand correctly we want to switchTravis from Linux to mac because protobuff has better osx support?

mrlukasbos:
Yes. Travis just builds the whole thing on a virtual machine and runs tests. Usually we tested on Ubuntu Xenial, but somehow protobuf took ages to build. We can probably use Docker, but for now I thought it was easy to switch to a macOS VM

RobotJesse:
Yeah Docker is probably a lot more work, though it sounds quite cool. Why do we need to build protobuf from scratch every time we run travis? I thought we could cache on travis as well

.travis.yml
```diff
@@ -1,43 +1,70 @@
...
+  - sudo make install #installs Google Test
+
+script:
+  - cd ~
+  - export CC="ccache gcc"
+  - export CXX="ccache g++"
+  - export CCACHE_SLOPPINESS=pch_defines,time_macros
+  - git clone https://github.com/RoboTeamTwente/roboteam_suite roboteam_suite --recurse-submodules
+  - cd ~/roboteam_suite
+  - rm -rf roboteam_ai
```
RobotJesse: Why is the AI folder removed?

mrlukasbos: The ai repo is cloned twice and we don’t need this one. In the rule below we replace roboteam-ai with the other one

mrlukasbos:
Yeah I don’t know. Protobuf took long and I remember zmq giving errors after that. But with build times over 20 minutes I did not want to debug that

.travis.yml
RobotJesse: Hahaha thanks, that documentation also answered about 6 other questions I had 😆🤦‍♂

.travis.yml
```diff
@@ -1,43 +1,62 @@
...
-  include:
-  - name: "Xenial"
-    dist: xenial
+os: osx
+osx_image: xcode11.2
 
-before_install:
-- sudo apt-get clean
-- sudo apt update
-- sudo apt-get install -y ccache ninja-build python3 python3-pip autoconf automake libtool make g++ unzip lcov gcovr qtbase5-dev libzmq3-dev libzmqpp-dev libarmadillo-dev
```
RobotJesse: Not all dependencies seem to be transferred from ubuntu to osx, but most of them are standard on the travis osx version as well, except possibly the code coverage ones?

RobotJesse: But as you mention, we should probably skip code cov for now

PR #706, opened 2019-10-31
Feature/behaviour tree no json

### LINKED ISSUE #693: Trees without json editor
We want to experiment with removing the json for the strategies/tactics/roles since this introduces a lot of extra bugs and complexity.

RobotJente:
- [x] The tests are passing

### Summary
Give a short summary of what the changes are and why you made them. If applicable, apply screenshots as well. 

### Related issues
If there are related issues, reference them here.

fixes #693   

### Comments for the reviewer
If there are comments for the reviewer, leave them here

**functionality**  
- Is the code is properly tested?
- Is the code working as expected
- Is the code is understandable and easy to read
- Are proper abstractions made? (no duplicate code, nice use of functions/objects)

**style**  
- Are smart pointers used?
- unused variables?
- magic numbers?
- well documented?
- No unused imports?
- Proper indentation
- Proper use of namespaces


RobotJesse:
If anyone would like to, please review the changes I made to the ProtoTypeTree class. I am mostly interested in the design of this class. I am aware that outside of this class, there are quite a few issues, which I am fixing now. You can also comment on these if you want, but I am still fixing them anyways.

include/roboteam_ai/treeinterp/TreeProtoType.h
```diff
@@ -0,0 +1,33 @@
...
+namespace bt {
+
+class TreeProtoType {
+ public:
+    std::shared_ptr<bt::DefaultTactic> createDefensiveTactic(std::shared_ptr<Blackboard> bb);
+    std::shared_ptr<bt::BehaviorTree> createNormalPlayStrategy();
+    std::shared_ptr<bt::Role> createDefenderRole(std::string name);
+
+private:
+    std::vector<std::pair<std::string, rtt::ai::robotDealer::RobotType>> robots = {
```
rolfvdhulst: I am generally not a fan of long default arguments, because you would expect this information in the constructor. When you run in release mode the overhead is optimised away anyways.

rolfvdhulst: Also, it keeps the headers small and readable

src/treeinterp/BTFactory.cpp
```diff
@@ -4,12 +4,20 @@
 
 #include "treeinterp/BTFactory.h"
 
+// The tactics variable saves the tactic of the behaviour tree and can be kept
+
+// Changes made by crazy person:
```
rolfvdhulst: Please refer from making comments like this, they do not help at all

src/treeinterp/TreeProtoType.cpp
```diff
@@ -0,0 +1,80 @@
...
+//
+#include <include/roboteam_ai/bt/composites/Selector.hpp>
+#include "bt/RoleDivider.h"
+#include "bt/BehaviorTree.hpp"
+#include "bt/decorators/Repeater.hpp"
+#include "treeinterp/TreeProtoType.h"
+#include "bt/tactics/DefaultTactic.h"
+#include "bt/Role.h"
+#include "skills/gotopos/GoToPos.h"
+#include "skills/Attack.h"
```
rolfvdhulst: Although this does not seem like an issue now, it can be problematic if a lot of includes are in one file because it means on every little change you will have to recompile all the trees. In a testing sense that can be good, but there might be an easier way. Previously, switches also had to be recompiled pretty much every time. There might be some easier way to do it with forward declaration or templates.

Also, do we plan on having all of the trees in one file? If so, that mi

rolfvdhulst (commented):
I just reviewed the relevant (asked for) parts. Generally looks ok and building trees is roughly how I imagined it to be. I think you need to think about how to handle roles; it feels a bit ugly to continuously have to pass them on right now.

src/treeinterp/BTFactory.cpp
RobotJesse: Yes that comment was mostly for me to keep track of where I made changes :grinning:

include/roboteam_ai/treeinterp/TreeProtoType.h
RobotJesse: Putting this in the constructor is a good idea! I think now there is a separate file for each of these robot vectors, but I feel like keeping them with the tactic/tree makes sense now that the jsons are out of the picture. The constructor does seem like a better place to do this than the header file :)

RobotJesse:
As @rolfvdhulst mentioned, creating roles within the strategy class is messy. Haico and I discussed this, and we are thinking about creating both the tactics and the strategy in the tree class, which implemenents a strategy interface. The roles will be separate classes. This is because the roles can be used by different tactics. Also I think it would be nice to make some base roles like AttackerRole etc, and then a variation of this, for example, AgressiveAttackerRole, can inherit from this class and only override the functionalities needed. However, I think these classes need to be made when they are necessary and not pre-engineered.

mrlukasbos (commented):
I added some comments for you to look at. Good to see some things are taking shape.

include/roboteam_ai/treeinterp/TreeProtoType.h
```diff
@@ -0,0 +1,27 @@
+//
+// Created by jesse on 14.10.19.
+//
+
+#include <include/roboteam_ai/bt/tactics/DefaultTactic.h>
+#include "bt/BehaviorTree.hpp"
+#include "bt/Role.h"
+#ifndef RTT_MAKE_TREE_WITHOUT_JSONS_H
```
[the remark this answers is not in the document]
RobotJesse: Will fix this, thank you for the warning :)

src/utilities/RobotDealer.cpp
```diff
@@ -95,7 +95,9 @@ int RobotDealer::claimRobotForTactic(RobotType feature, const std::string &roleN
 
         switch (feature) {
 
-        default:return - 1;
+        default:
+            std::cout << "no match for feature" << std::endl;
```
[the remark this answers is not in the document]
RobotJesse: It's a warning that the robotdealer can't find a robot with those specifications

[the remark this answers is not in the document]
mrlukasbos: I know that, but if i would get the error in my terminal I would not know that. 
Something like "[RobotDealer] could not find a robot with the desired properties" Could be better in this case

Besides, you changed behavior so that no robot is claimed (so you maybe don't take a penalty at all, and you hope you don't segfault).

include/roboteam_ai/bt/BehaviorTree.hpp
```diff
@@ -9,11 +9,13 @@ namespace bt {
 class BehaviorTree : public Node {
         FRIEND_TEST(BehaviorTreeTest, it_sets_blackboards);
     public:
-
+        std::string name;
         using Ptr = std::shared_ptr<BehaviorTree>;
 
         BehaviorTree();
 
+        BehaviorTree(std::string name);
```
HaicoDorenbos: A constructor is a public function so it needs documentation. I don't know if the name string should be a foreign key for example.

include/roboteam_ai/treeinterp/BTFactory.h
```diff
@@ -33,6 +33,8 @@ class BTFactory {
 
         static std::map<std::string, bt::BehaviorTree::Ptr> keeperRepo;
 
+        static std::map<std::string, bt::BehaviorTree::Ptr> codeTrees;
```
HaicoDorenbos: It would be nice if public variables like these have a comment which explains the meaning of it.

include/roboteam_ai/treeinterp/BTFactory.h
```diff
@@ -47,10 +49,17 @@ class BTFactory {
...
+
+
+private:
         static std::string currentTree;
         static std::string keeperTree;
         static bool weMadeTrees;
+
+
+
+        static std::shared_ptr<bt::BehaviorTree> testing_tree;
```
HaicoDorenbos: Also it would be nice if you have a comment which explains the meaning of it. Moreover prevent static variables where possible.

include/roboteam_ai/treeinterp/TreeProtoType.h
```diff
@@ -0,0 +1,49 @@
...
+     */
+    std::shared_ptr<bt::DefaultTactic> createOffensiveTactic(std::shared_ptr<Blackboard> bb);
+
+
+    /**
+     * Create an offender robot with a certain name. The role needs to be given to the tactic (so the tactic has 8 children that are roles)
+     * @param name, the name of the robot
+     * @return an offender role, a subtree containing the nodes necessary to execute this role.
+     */
+    std::shared_ptr<bt::Role> createOffenderRole(std::string name);
```
HaicoDorenbos: Nice that you document these functions appropriately :+1:

src/skills/Attack.cpp
```diff
@@ -30,6 +30,7 @@ bt::Node::Status Attack::onUpdate() {
         return Status::Running;
     }
 
+
```
HaicoDorenbos: I would rather remove this new line, because you already have a new line before it

src/skills/gotopos/GoToPos.cpp
```diff
@@ -21,6 +21,7 @@ GoToPos::GoToType GoToPos::stringToGoToType(const string &gtt) {
 }
 
 void GoToPos::onInitialize() {
+
```
HaicoDorenbos: I would rather remove this new line, because it is not useful to clarify the structure of your code

src/treeinterp/BTFactory.cpp
```diff
@@ -19,6 +28,21 @@ void BTFactory::makeTrees() {
 
     std::cout << "Re-Make Trees From Json" << std::endl;
 
+    // We need this instance of the TreeProtoType Class to use its methods. An alternative would be to make the methods static,
+    // but this essentially makes them global which is dangerous and unnecessary in my opinion. Additionally, some settings
+    // are initialized in the default constructor of TreeProtoType
```
HaicoDorenbos: Comments indeed should explain why you made certain decisions :+1: rather than how things work.

src/treeinterp/TreeProtoType.cpp
```diff
@@ -0,0 +1,104 @@
...
+        defensiveTactic->addChild(temprole);
+
+    }
+
+    return defensiveTactic;
+}
+
+/// This function creates a defender role. This role consists of a repeater with as child an attack skill
+std::shared_ptr<Role> TreeProtoType::createDefenderRole(std::string rolename) {
+    // set the rolename for the current role. This is important because the robotdealer decides how to deal the robots based on their rolenames
```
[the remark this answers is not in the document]
HaicoDorenbos: I indeed agree. However this is not really documentation but rather a comment. The general rule should be to have a lot of documentation in the header file and few comments in the cpp file (if you have a good design and decomposition of functions then everything should be self-explanatory). But since Jesse already added comments I would not recommend Jesse to remove them, although in the future you should not waste your time adding too many comments.

HaicoDorenbos (changes requested):
There are still some places where I expect you have documentation. And some global variables needs comments.

PR #711, opened 2019-11-12
Resolved all compiler warnings and removed some UB + general improvements

Yuhanun:
- [ ] The tests are passing 
Could not do tests due to linker errors

### Summary
Resolved most compiler warnings, resolved some UB

### Related issues
Check commit messages

### Comments for the reviewer
You can check commit messages + code itself for notitions on what I changed

### DIFF
--- a/include/roboteam_ai/interface/widgets/TreeVisualizerWidget.h
+++ b/include/roboteam_ai/interface/widgets/TreeVisualizerWidget.h
@@ -34,7 +34,7 @@ public:
     /**
      * Prevent future linker errors for undefined runtime vtable lookup
      */
-    virtual ~ThreeVisualizerWidget() = default; 
+    virtual ~TreeVisualizerWidget() = default; 
 public slots:
     void invalidateTree();
     void updateContents(bt::BehaviorTree::Ptr tree);
--- a/include/roboteam_ai/interface/widgets/TreeVisualizerWidget.h
+++ b/include/roboteam_ai/interface/widgets/TreeVisualizerWidget.h
@@ -22,6 +22,9 @@ class TreeVisualizerWidget : public QTreeWidget {
 private:
     QColor getColorForStatus(bt::Node::Status status);
     void addRootItem(bt::Node::Ptr parent, QTreeWidgetItem* QParent);
+    /**
+     * unordered_map
+     */
     std::map<QTreeWidgetItem *, bt::Node::Ptr> treeItemMapping;
     bool hasCorrectTree = false;
     MainWindow * parent = nullptr;
--- a/include/roboteam_ai/interface/widgets/TreeVisualizerWidget.h
+++ b/include/roboteam_ai/interface/widgets/TreeVisualizerWidget.h
@@ -30,6 +30,11 @@ private:
 public:
     explicit TreeVisualizerWidget(MainWindow * parent);
     void setHasCorrectTree(bool hasCorrectTree);
+
+    /**
+     * Prevent future linker errors for undefined runtime vtable lookup
+     */
+    virtual ~ThreeVisualizerWidget() = default; 
 public slots:
     void invalidateTree();
     void updateContents(bt::BehaviorTree::Ptr tree);
--- a/src/analysis/GameAnalyzer.cpp
+++ b/src/analysis/GameAnalyzer.cpp
@@ -44,7 +44,13 @@ std::shared_ptr<AnalysisReport> GameAnalyzer::generateReportNow() {
 }
 
 BallPossession GameAnalyzer::convertPossession(rtt::ai::BallPossession::Possession possession) {
+    /**
+     * Default should usually be the bottom one
+     */
     switch (possession) {
+    case (rtt::ai::BallPossession::CONTENDEDBALL): return BallPossession::NEUTRAL;
+    case (rtt::ai::BallPossession::THEIRBALL): return BallPossession::THEY_HAVE_BALL;
+    case (rtt::ai::BallPossession::OURBALL): return BallPossession::WE_HAVE_BALL;
     default:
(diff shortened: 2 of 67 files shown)


include/roboteam_ai/interface/widgets/TreeVisualizerWidget.h
```diff
@@ -30,6 +30,11 @@ class TreeVisualizerWidget : public QTreeWidget {
 public:
     explicit TreeVisualizerWidget(MainWindow * parent);
     void setHasCorrectTree(bool hasCorrectTree);
+
+    /**
+     * Prevent future linker errors for undefined runtime vtable lookup
+     */
+    virtual ~ThreeVisualizerWidget() = default;
```
mrlukasbos: typo

Yuhanun: Okay well I'm basically removing boost:: as a dependency right now as both std::variant and std::optional exist now, which is the only thing you use from boost, so I'll fix that aswel

src/interface/widgets/TreeVisualizerWidget.cpp
```diff
@@ -98,7 +98,9 @@ void TreeVisualizerWidget::populateRow(bt::Node::Ptr node, QTreeWidgetItem* row,
     QString status = QString::fromStdString(node->status_print(node->getStatus()));//statusToString(node->getStatus()));
     if (row->text(1)!=status) {
         row->setText(1, status);
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
mrlukasbos: I personally do not like ignoring these warnings. These functions are deprecated and therefore need replacement instead of being ignored

Yuhanun: Exactly. they need to be but im not familiar with Qt5 and I couldn't quickly find a function that satisfy the requirements for this call, I temporarily suppressed it, I thought i added a comment saying they had to be replaced but I guess not :)

I'm completely against any warnings, if it were up to me I'd add -Wall -Wextra -Werror -pedantic to compiler flags.

Yuhanun:
non-breaking changes, compilation has been tested.

General performance increases due to constexpr and some comments on general C++ stuff

src/analysis/GameAnalyzer.cpp
```diff
@@ -44,7 +44,13 @@ std::shared_ptr<AnalysisReport> GameAnalyzer::generateReportNow() {
 }
 
 BallPossession GameAnalyzer::convertPossession(rtt::ai::BallPossession::Possession possession) {
+    /**
+     * Default should usually be the bottom one
+     */
     switch (possession) {
```
Yuhanun: Discard this change, I'm dumb and made a mistake, my bad

rolfvdhulst: Making mistakes does not make you a dumb person :)

src/bt/Blackboard.cpp
```diff
@@ -12,9 +12,14 @@ void Blackboard::setBool(std::string key, bool value) {
...
-        bools[key] = false;
-    }
+    // if (bools.find(key) == bools.end()) {
+    //     bools[key] = false;
+    // }
+    // bools[key] deafult constructs if no exist
+    // https://en.cppreference.com/w/cpp/container/map/operator_at
+    // If an insertion is performed, the mapped value is value-initialized (default-constructed for class types, zero-initialized otherwise) and a reference to it is returned.
+
+    // https://godbolt.org/z/87GUTV
```
rolfvdhulst: I'm pretty sure in this class it's actually intended behavior to return 0/false/empty string if the value is not found. This is one of our most ugly classes though, so I don't blame you.

Yuhanun: I know, and that's exactly what happens if the value is not found

bools[key] is explicitly set to false however operator[] will already do that for you.

rolfvdhulst: I'm not sure if I like that personally, as it makes readability more difficult. If you change it I think it's at least worth a comment since I don't expect the average programmer in our system to know that the access operator also zero-initializes a field.

Yuhanun: Okay, well the issue is not just that it makes code look more messy / better.

std::map::find() on itself is already O(n log n), then operator[] is another O(n log n) complexity, yes the complexity stays the same but the actual time it takes to perform 2 O(n log n) operations would double, it's not merely a readability thing. But yeah, it definitely could use a comment since looking at it for the first time without any operator[] knowledge could cause some confusion.

rolfvdhulst: I fully agree we should have your proposed changes here with a comment. Readability vs efficiency is an important discussion to have generally. The reason why you see so little unordered maps, constexpr and stuff like that is because most computations and skills do not really take up significant amounts of computing time in our system. 90+% of the computational time costs comes from 4 key algorithms. 

We however spend a lot more time fixing bugs and mistakes caused by difficult to read code and just bad code practices than optimizing our code, because in most places optimization is not really necessary. Only the code that is consistently run very often (these 4 algorithms) needs to be fast and efficient. In places like this, where the map is often very small, the difference between  O log(1) vs O log(n) vs O(n) or 2 O(n log(n)) is minimal, so we typically do not spend that much time worrying about performance, and as a consequence this code is often not very efficient. 

This is merely an explanation (and not a rebuttal or counterargument, I fully agree!), but keep it in mind if you want to go through more code, so you do not get frustrated. For us, human mistakes are a far bigger issue than program performance.

Yuhanun:
@HaicoDorenbos Thank you very much, I try :) took me a lot of effort to get to this level, I tend to pick up new concepts relatively quickly, which is why I'm excited about learning new things.

Thanks a lot :)

PR #714, opened 2019-11-27
WIP: Feature/midfieldharasscpp

RobotJente:
- [ ] The tests are passing

### Summary
In this PR (which is branched from the unmerged TreeProtoType merge request), I create a midfieldharassrole in a CPP file. I've created helper functions for certain common subtrees. I notice the code in the createMidFieldHarassRole is quite long and some parts are a bit hardcoded. If anyone has feedback/new ideas for this, I would be love to hear it!

- Are smart pointers used?
- unused variables?
- magic numbers?
- well documented?
- No unused imports?
- Proper indentation
- Proper use of namespaces


RobotJesse:
I think it is difficult to construct the tree in a way that makes sense. I am currently thinking of just making a simple regex parses that checks for "addChild()" and uses that to make a visualisation of the current tree, because the advantage of the interface is the visualization aspect.

src/treeinterp/BTFactory.cpp
```diff
@@ -19,6 +28,21 @@ void BTFactory::makeTrees() {
 
     std::cout << "Re-Make Trees From Json" << std::endl;
 
+    // We need this instance of the TreeProtoType Class to use its methods. An alternative would be to make the methods static,
+    // but this essentially makes them global which is dangerous and unnecessary in my opinion. Additionally, some settings
+    // are initialized in the default constructor of TreeProtoType
+    bt::TreeProtoType protoTypeTree;
+    std::shared_ptr<bt::BehaviorTree> testingTree;
+    testingTree = protoTypeTree.createOffensiveStrategy();
```
rolfvdhulst: Static methods != global methods if they are properly handled using includes. Global objects are very bad because they state you cannot reason about. " Global" static methods are actually quite fine since they do not have any implicit state.

Also, I think the treeProtoType does not need to be a class here but it can just be a method. I feel the name is very misleading, and it's not really clear what happens when you construct a 'TreeProtoType' .

include/roboteam_ai/treeinterp/TreeProtoType.h
```diff
@@ -0,0 +1,49 @@
...
+         * Creates a Strategy behaviour tree. This tree is created in BTFactory.
+         * @return the behaviour tree that contains this strategy
+         */
+        std::shared_ptr<bt::BehaviorTree> createOffensiveStrategy();
+
+    private:
+        /**
+         * Vector that is used by RobotDealer to determine how to match robots to robot IDs.
+         */
+        std::vector<std::pair<std::string, rtt::ai::robotDealer::RobotType>> robots;
```
rolfvdhulst: So basically the only state this object holds is the robotDealer labels, which are never updated. You might want to consider just making this a function

CMakeLists.txt
```diff
@@ -337,13 +339,14 @@ add_executable(roboteam_ai
         ${DEMO_SOURCES}
         ${INTERFACE_SOURCES}
         ${WORLD_SOURCES}
-        )
+        src/treeinterp/MidFieldHarassRole.cpp include/roboteam_ai/treeinterp/MidFieldHarassRole.h src/TreeHelper/BeingPassedToHelper.cpp include/roboteam_ai/TreeHelper/BeingPassedToHelper.h src/TreeHelper/RobotOutOfFieldHelper.cpp include/roboteam_ai/TreeHelper/RobotOutOfFieldHelper.h)
```
rolfvdhulst: please put this in one of the variables (such as ${BT_SOURES} (or make a new one) and put an enter in between each file.

rolfvdhulst (commented):
Looks good in general so far, I think the idea's are heading the right direction (not approving because it is WIP). I have some minor comments, but nothing important.
Also, I personally think you put a bit too many comments which explain things that are either unimportant or very vague (although that's fine for now, it might get annoying later). Maybe you want to work with //TODO (which CLION can find) so you can easily see what things you still need to fix. You seem a bit chaotic and all over the place now ;)

src/treeinterp/BTFactory.cpp
RobotJesse: I think it can be improved by renaming the treeprototype class to, in this case, OffensiveStrategy, and then having a create method for it, I think this would remedy the confusing name. I see your argument for making the methods static, I think this does remove unnecessary code. So having something like OffensiveStrategy::create() seems ok to me, but I am curious what you think about this.

include/roboteam_ai/treeinterp/TreeProtoType.h
RobotJesse: I'm not sure I understand what you mean. I think making this vector in the constructor is nice, since it means you always have some default robotdealer config, but if you decide later on you want the strategy but with some other flags you can easily build this functionality this way.

RobotJesse: I think I understand what you mean now, making everything static bypasses the constructor so I agree that making a method for this is better

src/treeinterp/BTFactory.cpp
RobotJesse: I've changed my mind about this, as bypassing the constructor means you can't give the OffensiveStrategy the tree it makes since the object itself is never created, and some stuff also needs to be done in the constructor

Issue #537, opened 2019-06-03
Passcoach makes passes fail too easy


Barisimre:
There is something like an error where if the pass is too inaccurate the receiver doesn't work. This should also be dependent on ball speed, robot positions and in general be more okay with less accurate passes

RobZelluf:
Can you elaborate? @Barisimre 

Should the passcoach just return success more often, or should it return passes that are more accurate?

RobZelluf:
Or should the receiver fail less quickly, so just continue trying to receive the ball?

PR #715, opened 2019-11-29
Solve pass strictness problem

HaicoDorenbos:
Solved issue #537 by removing the requirements that the ball can only be passed to robots across the midline and removed the minimum distance requirement for ball passes.  Created a new issue #716 regarding the passes between defenders (which unfortunately still does not happen).


RobotJente (approved):
Nice fix :)

PR #717, opened 2019-11-29
added some nice constructors for selector and sequence

RobotJente:
Constructors for selector and sequence that make it clearer what the children are. Right now the order is set by the way you call "addChild(Node node)" but I think this adds too many lines of code and reduces readability. Therefore I made this constructor, which makes it clear what nodes are in what order. Examle usage would be:
Attack attack = new Attack(bla, blob);
GTP go = new GTP(here, there);
IsARobot isARobot = new IsARobot(this);

Selector select = new Selector(attack, go, isARobot);

### DIFF
--- a/include/roboteam_ai/bt/Node.hpp
+++ b/include/roboteam_ai/bt/Node.hpp
@@ -39,6 +39,11 @@ class Node {
 
         using Ptr = std::shared_ptr<Node>;
 
+        /**
+         * The update function executes the node. This function is overwritten a lot, most interestingly in the Selector/Sequence/Inverter nodes,
+         * since these nodes can have children. Therefore the update function of these nodes ticks their children.
+         * @return Status result of the update (Running, Success, Failure)
+         */
         virtual Status update() = 0;
 
         Status NodeUpdate();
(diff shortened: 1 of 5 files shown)


include/roboteam_ai/bt/composites/Selector.hpp
```diff
@@ -6,6 +6,7 @@ namespace bt {
 
 class Selector : public Composite {
 public:
+    Selector(std::vector<std::shared_ptr<bt::Node>> childs);
```
mrlukasbos: "children" ;-)

mrlukasbos (approved):
Easy but smart improvement, especially in context with the task at hand. Good PR!

include/roboteam_ai/bt/Node.hpp
```diff
@@ -39,6 +39,11 @@ class Node {
 
         using Ptr = std::shared_ptr<Node>;
 
+        /**
+         * The update function executes the node. This function is overwritten a lot, most interestingly in the Selector/Sequence/Inverter nodes,
+         * since these nodes can have children. Therefore the update function of these nodes ticks their children.
+         * @return Status result of the update (Running, Success, Failure)
+         */
```
HaicoDorenbos: :+1:

HaicoDorenbos (approved):
Looks fine to me

PR #720, opened 2019-12-04
Feature/CoachDefenderRole

jordiluong:
- [ ] The tests are passing

### Summary
Created CoachDefenderRole in cpp. Tree is tested and works.


CMakeLists.txt
```diff
@@ -31,7 +31,6 @@ if(UNIX AND NOT APPLE)
     SET(Qt5Widgets_DIR "/usr/include/x86_64-linux-gnu/qt5/QtWidgets")
     SET(Qt5Charts_DIR "/usr/include/x86_64-linux-gnu/qt5/QtCharts")
 endif()
-
```
rolfvdhulst: why? These spaces are actually nice to have for readability

CMakeLists.txt
```diff
@@ -337,13 +338,15 @@ add_executable(roboteam_ai
         ${DEMO_SOURCES}
         ${INTERFACE_SOURCES}
         ${WORLD_SOURCES}
-        )
+        src/treeinterp/MidFieldHarassRole.cpp include/roboteam_ai/treeinterp/MidFieldHarassRole.h src/TreeHelper/BeingPassedToHelper.cpp include/roboteam_ai/TreeHelper/BeingPassedToHelper.h src/TreeHelper/RobotOutOfFieldHelper.cpp include/roboteam_ai/TreeHelper/RobotOutOfFieldHelper.h src/treeinterp/CoachDefenderRole.cpp include/roboteam_ai/treeinterp/CoachDefenderRole.h)
```
rolfvdhulst: Please don't do this, but add them to one of the ${SOURCES} variables such as BT or whereever they could be relevant.

rolfvdhulst: And seperate them onto multiple lines.

CMakeLists.txt
```diff
@@ -382,13 +385,14 @@ if(APPLE)
...
 if(UNIX AND NOT APPLE)
-
   target_link_libraries(ai_tests
         PRIVATE roboteam_utils
         PRIVATE Qt5::Widgets
         PRIVATE Qt5::Gui
         PRIVATE Qt5::Charts
-        PRIVATE gcov)
+        PRIVATE gcov
+          gtest
```
rolfvdhulst: You should probably mark gtest private.

include/roboteam_ai/TreeHelper/RobotOutOfFieldHelper.h
```diff
@@ -0,0 +1,18 @@
+//
+// Created by jessevw on 21.11.19.
+//
+
+#ifndef RTT_ROBOTOUTOFFIELDHELPER_H
+#define RTT_ROBOTOUTOFFIELDHELPER_H
+#include <include/roboteam_ai/bt/Node.hpp>
```
rolfvdhulst: It's minor but please be consistent with include style between doing include/ or doing bt/ and the usage of ""and <> between different files of the same "type" (such as these helpers).

include/roboteam_ai/bt/Node.hpp
```diff
@@ -96,6 +98,7 @@ class Node {
 
         rtt::ai::world::World * world = nullptr;
         rtt::ai::world::Field * field = nullptr;
-};
+
+    };
```
rolfvdhulst: unnecessary.

include/roboteam_ai/bt/Node.hpp
```diff
@@ -87,6 +87,8 @@ class Node {
 
 //        ros::Time getLastTickTime();
 
+        void setRole(std::string roleName);
```
rolfvdhulst: Maybe name this function setRoleString() or so, so that the difference between the class Node and the function is clear for outside users.

include/roboteam_ai/bt/composites/Selector.hpp
```diff
@@ -6,6 +6,7 @@ namespace bt {
 
 class Selector : public Composite {
 public:
+    Selector();
```
rolfvdhulst: Why add all these empty constructors? (Also goes for all other composite.h files) Isn't it better to initialize explicitly so certain nodes MUST have children?

include/roboteam_ai/skills/AvoidBall.h
```diff
@@ -21,7 +21,7 @@ class AvoidBall : public Skill {
 
     public:
         explicit AvoidBall(std::string name, bt::Blackboard::Ptr blackboard = nullptr);
-        void onInitialize() override;
+        void onInitialize(std::string type);
```
rolfvdhulst: Ooh, I like this way of doing it.

include/roboteam_ai/treeinterp/BTFactory.h
```diff
@@ -47,10 +49,17 @@ class BTFactory {
...
-    private:
+
+
+private:
         static std::string currentTree;
         static std::string keeperTree;
         static bool weMadeTrees;
+
+
+
```
rolfvdhulst: I know you need somewhere to breath clean air after reading this code but you won't find it by adding extra space to our code.

include/roboteam_ai/treeinterp/CoachDefenderRole.h
```diff
@@ -0,0 +1,32 @@
+//
+// Created by jordi on 03-12-19.
+//
+
+#ifndef RTT_COACHDEFENDERROLE_H
+#define RTT_COACHDEFENDERROLE_H
+#include <include/roboteam_ai/bt/tactics/DefaultTactic.h>
+#include "bt/Role.h"
```
rolfvdhulst: Aargh consistent includes

include/roboteam_ai/treeinterp/CoachDefenderRole.h
```diff
@@ -0,0 +1,32 @@
...
+         * Creates a Strategy behaviour tree. This tree is created in BTFactory.
+         * @return the behaviour tree that contains this strategy
+         */
+        std::shared_ptr<Role> createCoachDefenderRole(std::string rolename);
+
+    private:
+        /**
+         * Vector that is used by RobotDealer to determine how to match robots to robot IDs.
+         */
+        std::vector<std::pair<std::string, rtt::ai::robotDealer::RobotType>> robots;
```
rolfvdhulst: Although I already told you, I am personally not a fan of this being a class but would rather see something like a static function or so. Also, the name of the class implies it in itself is a Node, which it's not (since Role's ARE nodes). You would looking at the name expect this to be a superclass of Role.

include/roboteam_ai/treeinterp/MidFieldHarassRole.h
```diff
@@ -0,0 +1,33 @@
+//
+// Created by jessevw on 19.11.19.
+//
+
+#ifndef RTT_MIDFIELDHARASSROLE_H
+#define RTT_MIDFIELDHARASSROLE_H
+#include <include/roboteam_ai/bt/tactics/DefaultTactic.h>
+#include "bt/BehaviorTree.hpp"
```
rolfvdhulst: .

rolfvdhulst: Same comments as on other "Role"

include/roboteam_ai/treeinterp/OffensiveStrategy.h
```diff
@@ -0,0 +1,57 @@
...
+         * Vector that is used by RobotDealer to determine how to match robots to robot IDs.
+         */
+        std::vector<std::pair<std::string, rtt::ai::robotDealer::RobotType>> robots;
+
+        /**
+         * The tree that is given to the offensive strategy
+         */
+        std::shared_ptr<bt::BehaviorTree> tree;
+
+        /**
```
rolfvdhulst: So what kind of design pattern are you really choosing now. Are you going to make everything like this with ownership of the tree or not? This is inconsistent with what you did in both of the new "Role" files.

rolfvdhulst: I'mtalking about the std::shared_ptr<bt::BehaviorTree tree> bit.

src/TreeHelper/BeingPassedToHelper.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace bt {
+    BeingPassedToHelper::BeingPassedToHelper() {}
+
+    std::shared_ptr<bt::Node> BeingPassedToHelper::createBeingPassedToChecker() {
+        auto localbb = std::make_shared<bt::Blackboard>();
+
+        std::shared_ptr<bt::Sequence> seque = std::make_shared<bt::Sequence>();
+        std::shared_ptr<rtt::ai::Receive> receive = std::make_shared<rtt::ai::Receive>("receivernode", localbb);
+        std::shared_ptr<rtt::ai::IsInDefenseArea> defenseArea = std::make_shared<rtt::ai::IsInDefenseArea>("defendy", localbb);
```
rolfvdhulst: I don't like the string naming, but not a huge problem.

rolfvdhulst: As in, use names you can recognize 3 months down the line (especially 'defendy' is very vague)

src/bt/BehaviorTree.cpp
```diff
@@ -18,6 +21,7 @@ BehaviorTree::BehaviorTree(const Blackboard::Ptr &shared)
     globalBB = shared;
 }
 
+
```
rolfvdhulst: .

src/skills/Attack.cpp
```diff
@@ -30,6 +30,7 @@ bt::Node::Status Attack::onUpdate() {
         return Status::Running;
     }
 
+
```
rolfvdhulst: .

src/skills/gotopos/GoAroundPos.cpp
```diff
@@ -12,7 +12,9 @@ namespace ai {
 GoAroundPos::GoAroundPos(string name, bt::Blackboard::Ptr blackboard)
         :GoToPos(name, blackboard) { }
 
-void GoAroundPos::gtpInitialize() {
+
+
```
rolfvdhulst: .

src/treeinterp/BTFactory.cpp
```diff
@@ -3,13 +3,22 @@
...
+std::map<std::string, bt::BehaviorTree::Ptr>BTFactory::codeTrees;
+
 std::string BTFactory::currentTree = "NaN";
 std::string BTFactory::keeperTree;
 std::mutex BTFactory::keeperTreeMutex;
+
+// C++ code trees
+
+
+
```
rolfvdhulst: .

src/treeinterp/MidFieldHarassRole.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+        auto memSelector = std::make_shared<bt::MemSelector>();
+
+        auto clearShotSeque = std::make_shared<bt::Sequence>();
+        auto hasClearShot = std::make_shared<rtt::ai::HasClearShot>("hasclearshot", localbb);
+        auto attack = std::make_shared<rtt::ai::Attack>(name, localbb);
+
+        auto passSkill = std::make_shared<rtt::ai::Pass>("Pass", localbb);
+        auto attackPassSkill = std::make_shared<rtt::ai::Attack>("Pass attack", localbb);
+        auto passAttackSeque = std::make_shared<bt::Sequence>();
```
rolfvdhulst: If we are going to give these all default names like name Attack node Attack, we should probably default it in the constructor of the node or class definition, and only override when necessary.

src/treeinterp/CoachDefenderRole.cpp
```diff
@@ -0,0 +1,97 @@
...
+        ballOutOfFieldBb->setDouble("secondsAhead", 0.4);
+        auto ballOutOfFieldNode = std::make_shared<rtt::ai::BallOutOfField>("ballOutOfField", ballOutOfFieldBb);
+        auto inverterBallOutOfField = std::make_shared<bt::Inverter>();
+        inverterBallOutOfField->addChild(ballOutOfFieldNode);
+
+        auto isInDefenseAreaBb = std::make_shared<bt::Blackboard>();
+        isInDefenseAreaBb->setString("ROLE", rolename);
+        isInDefenseAreaBb->setBool("ourDefenseArea", true);
+        isInDefenseAreaBb->setBool("outsideField", true);
+        isInDefenseAreaBb->setDouble("secondsAhead", 0.4);
```
rolfvdhulst: In a case like this I would put a constant like const double SECONDSAHEAD=0.4 and call it in both line 51 and 60.

rolfvdhulst (changes requested):
I think the design choice is not very clear; either store the trees as data members or not, but don't mix them. A lot of comments about naming, unnecessary spaces and include style. Functionally most seems fineThere are also some things you can do to make your life easier, like initializing a skill's name in the constructor so it's not necessary to type everytime you create a new node of that skill.

include/roboteam_ai/TreeHelper/BeingPassedToHelper.h
```diff
@@ -0,0 +1,23 @@
...
+#ifndef RTT_BEINGPASSEDTOHELPER_H
+#define RTT_BEINGPASSEDTOHELPER_H
+#include "bt/BehaviorTree.hpp"
+
+/**
+ * Wonder if this should be a class in itself as it is now, or part of a class and as a function.
+ */
+
+namespace bt {
+    class BeingPassedToHelper {
```
Yuhanun: Shouldn't this inherit from Node?

include/roboteam_ai/bt/BehaviorTree.hpp
```diff
@@ -9,11 +9,13 @@ namespace bt {
 class BehaviorTree : public Node {
         FRIEND_TEST(BehaviorTreeTest, it_sets_blackboards);
     public:
-
+        std::string name;
```
Yuhanun: Don't we usually return the name from node_name as a literal? Allowing it to get inlined?

include/roboteam_ai/bt/Node.hpp
Yuhanun: Well yes but we usually just have a literal in node_name

include/roboteam_ai/bt/composites/Selector.hpp
Yuhanun: Because std::make_unique<selector> requires a default constructor, we construct a map somewhere of it std::string to std::shared_ptr<Node> iirc, which requires a default ctor, stl things.

include/roboteam_ai/skills/AvoidBall.h
Yuhanun: isn't onInitialize() literally a constructor? By definition a constructor is called when you initialize.

include/roboteam_ai/treeinterp/MidFieldHarassRole.h
```diff
@@ -0,0 +1,33 @@
...
+namespace bt {
+
+    class MidFieldHarassRole {
+    public:
+        MidFieldHarassRole();
+        /**
+         * Creates a Strategy behaviour tree. This tree is created in BTFactory.
+         * @return the behaviour tree that contains this strategy
+         */
+        std::shared_ptr<Role> createMidFieldHarassRole(std::string rolename);
```
Yuhanun: I'm liking the documentation :) however you sometimes forget to do @param

src/Switches.cpp
```diff
@@ -597,16 +597,17 @@ bt::Node::Ptr Switches::tacticSwitch(std::string name, bt::Blackboard::Ptr prope
...
-                    {"c1", robotType::CLOSE_TO_BALL},
-                    {"c2", robotType::RANDOM},
-                    {"c3", robotType::RANDOM},
-                    {"c4", robotType::RANDOM},
-                    {"c5", robotType::RANDOM},
-                    {"c6", robotType::RANDOM},
-                    {"c7", robotType::RANDOM},
-                    {"c8", robotType::RANDOM}
-            }
+                {"c1", robotType::CLOSE_TO_BALL},
```
Yuhanun: Honestly this whole switches.cpp file is kinda questionable, but whatever, there might not be a better way.

src/TreeHelper/RobotOutOfFieldHelper.cpp
```diff
@@ -0,0 +1,35 @@
...
+//
+#include <include/roboteam_ai/TreeHelper/RobotOutOfFieldHelper.h>
+#include <include/roboteam_ai/bt/composites/Sequence.hpp>
+#include <include/roboteam_ai/conditions/RobotOutside.h>
+#include <include/roboteam_ai/skills/gotopos/GTPSpecial.h>
+#include "bt/RoleDivider.h"
+#include "bt/Role.h"
+#include "skills/Attack.h"
+
+// Reference wrapper
```
Yuhanun: yeeaaah! std::reference_wrapper instead of pointers could go a long way :eyes: however, reference_wrapper is owning, so not sure.

src/bt/Node.cpp
```diff
@@ -114,4 +114,17 @@ std::string Node::status_print(Node::Status s) {
     }
 }
 
+void Node::setRole(std::string roleName) {
+    auto children = this->getChildren();
+    if (properties) {
```
Yuhanun: ?... Isn't properties a default constructed shared map in every case? why bother checking whether the pointer is null when it (should) never be?

src/bt/Node.cpp
```diff
@@ -114,4 +114,17 @@ std::string Node::status_print(Node::Status s) {
...
+        this->properties->setString("ROLE", roleName);
+    }
+
+    for (auto child : this->getChildren()) {
+        if (child) {
```
Yuhanun: Are children sometimes null?... how

Yuhanun: They're shared pointers, their whole point is to always be valid until you dont use it anymore.

src/bt/Role.cpp
```diff
@@ -23,18 +24,17 @@ Node::Status Role::update() {
...
-
 }
+
 std::string Role::node_name() {
     return name;
 
 }
-Role::Role(std::string name) {
-    this->name = std::move(name);
+Role::Role(std::string name) : name(name) {
```
Yuhanun: string_view :eyes:

src/bt/composites/Selector.cpp
```diff
@@ -34,4 +34,8 @@ Node::Status Selector::update() {
     return Status::Failure;
 }
 
+    Selector::Selector() {
```
Yuhanun: = default;

src/bt/composites/Sequence.cpp
```diff
@@ -37,4 +37,8 @@ Node::Status Sequence::update() {
     return Status::Success;
 }
 
+    Sequence::Sequence() {
```
Yuhanun: = default;

src/skills/AvoidBall.cpp
```diff
@@ -20,12 +20,14 @@ AvoidBall::AvoidBall(std::string name, bt::Blackboard::Ptr blackboard)
 : Skill(std::move(name), std::move(blackboard)) {
 }
 
-void AvoidBall::onInitialize() {
+void AvoidBall::onInitialize(std::string type) {
```
Yuhanun: string_view, or const&

src/treeinterp/BTFactory.cpp
```diff
@@ -3,13 +3,22 @@
...
 #include "treeinterp/BTFactory.h"
+#include "treeinterp/OffensiveStrategy.h"
+
 
 std::map<std::string, bt::BehaviorTree::Ptr> BTFactory::strategyRepo;
 std::map<std::string, bt::Node::Ptr>BTFactory::tacticsRepo;
 std::map<std::string, bt::BehaviorTree::Ptr>BTFactory::keeperRepo;
+std::map<std::string, bt::BehaviorTree::Ptr>BTFactory::codeTrees;
+
 std::string BTFactory::currentTree = "NaN";
```
Yuhanun: lol, nice

src/treeinterp/BTFactory.cpp
```diff
@@ -19,6 +28,14 @@ void BTFactory::makeTrees() {
 
     std::cout << "Re-Make Trees From Json" << std::endl;
 
+    /*
+     * Here we store the C++ trees in a map, key = treename, val = cpp tree.
+     * In order to do this in a cleaner way, maybe build trees automatically by going through directory
+     */
+    auto Off = new bt::OffensiveStrategy();
```
Yuhanun: Memory leak.

Change to 

auto Off = bt::OffensiveStrategy();

codeTrees[...] = Off.createOffensiveStrategy();

src/treeinterp/BTFactory.cpp
```diff
@@ -42,9 +59,20 @@ void BTFactory::makeTrees() {
...
+ * @param treeName the name of the behaviour tree you are requesting.
+ * @return The behaviourtree corresponding to that treename (currently hardcoded)
+ */
 bt::BehaviorTree::Ptr BTFactory::getTree(std::string treeName) {
     std::lock_guard<std::mutex> lock(keeperTreeMutex);
 
+    // Un-kill the code below by commenting the return statement to restore json functionality
+    auto treefound = codeTrees.find("attackertree");
+    return treefound->second;
+    return codeTrees["attackertree"];
```
Yuhanun: two return statements.

src/treeinterp/OffensiveStrategy.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "skills/Attack.h"
+
+#include "include/roboteam_ai/treeinterp/CoachDefenderRole.h"
+
+
+namespace bt {
+
+    OffensiveStrategy::OffensiveStrategy() {
+        // Set the robottypes for the robot so the robotdealer can decide which robot should do what
+        this->robots = {
```
Yuhanun: you can do this in the class definition already

std::map... Robots = {
   { "o1", ... },
   ...
} 

Allows for more optimizations + a default ctor

src/treeinterp/OffensiveStrategy.cpp
```diff
@@ -0,0 +1,85 @@
...
+            auto temphelper = bt::CoachDefenderRole();
+            std::shared_ptr<Role> temprole = temphelper.createCoachDefenderRole(name);
+            temprole->setRole(name);
+            offensiveTactic->addChild(temprole);
+        }
+
+        return offensiveTactic;
+    }
+
+    std::shared_ptr<Role> OffensiveStrategy::createOffenderRole(std::string name) {
```
Yuhanun: so many shared_ptr's not your fault tho <3

src/utilities/RobotDealer.cpp
```diff
@@ -95,7 +95,9 @@ int RobotDealer::claimRobotForTactic(RobotType feature, const std::string &roleN
 
         switch (feature) {
 
-        default:return - 1;
+        default:
+            std::cout << "[Robotdealer] could not find a match for this feature " << std::endl;
```
Yuhanun: std::cerr

PR #735, opened 2019-12-17
fix segfault in interface

### LINKED ISSUE #702: Segfault in interface
**Describe the bug**
Segfault in the visual interface. (interface/widgets/widget.cpp), mousepressEvent function has world()->getworld().us. This pointer is dead on startup.

**To Reproduce**
Open an AI without an available world state. 
Click multiple times on the field

**Expected behavior**
Don't crash?

mrlukasbos:
fixes #702

### DIFF
--- a/src/world/World.cpp
+++ b/src/world/World.cpp
@@ -105,7 +105,7 @@ bool World::weHaveRobots() {
 const WorldData World::getWorld() {
     std::lock_guard<std::mutex> lock(worldMutex);
     auto thing = WorldData(*worldDataPtr);
-    return thing;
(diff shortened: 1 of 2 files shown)


src/world/World.cpp
```diff
@@ -104,8 +104,7 @@ bool World::weHaveRobots() {
 
 const WorldData World::getWorld() {
     std::lock_guard<std::mutex> lock(worldMutex);
-    auto thing = WorldData(*worldDataPtr);
-    return thing;
+    return WorldData(*worldDataPtr);
```
mrlukasbos: couldnt resist

PR #740, opened 2019-12-18
FieldMessage and Field seperation

HaicoDorenbos:
This PR is a Christmas present for the RoboTeam :christmas_tree:. In this PR I have basically seperated the FieldMessage class and Field class, which means that:
- The previous FieldMessage class now takes the role of the Field class (contains the current state of the field)
- The previous Field class now takes the role of FieldComputation class (it takes the previous FieldMessage class as argument in all its functions as pointer and does computations on it).
- Static references to the previous Field class have been removed and are added to the previous FieldMessage class instead (like a Singleton pattern) and are used only by classes which did not have a reference to the previous Field class.
- All methods in the previous Field class have been made static. (In the future I might create a separate interface for this)

And because of this seperation we have a major performance increase, because the FieldComputation is no longer responsible for accessing/copying the Field class. The classes that call the FieldComputation class are now responsible for accessing/copying the Field class which results in less copying of the Field, since the FieldComputation class was often called multiple times in the same function which results in a new copy everytime a function in the FieldComputation class was called.

When reviewing this PR you should check whether the behavior of the system did not accidentally changed (which should not have happened). Also it is good to check whether these changes are good for the performance. And last of all you should check whether the main design has improved or not (do not focus too much on this because in future PR this will be more improved).


rolfvdhulst:
Can you explain why you made field a singleton?

HaicoDorenbos:
> Can you explain why you made field a singleton? I feel like most of your changes will only cause further problems with multithreading, performance and synchronization. I don't understand why field is not just a simple data format which is managed by a tiny class that wraps it. Singletons in general are a really, really bad idea, especially for a data format like this which you want to be copyable. If you are going to change it I suggest using Lukas's solution and just passing it by reference or pointer everywhere,

Because the previous Field class also had a static variable which was accessed everywhere in the code. With the current state of the code it is not very easy to remove all those static accesses to the Field class. That is the reason why I made this decision. I indeed agree that we should try to remove all static accesses, but that will take much longer to achieve.

mrlukasbos:
As you requested my review, I have some comments. 

> And because of this seperation we have a major performance increase, because the FieldComputation is no longer responsible for accessing/copying the Field class

I very much doubt this. Previously we were accessing pointers and had one function indirection, you have a map lookup. Both are not significant to the performance, and I am quite sure it is not a major performance increase. Have you run a profiler for comparison? 

> With the current state of the code it is not very easy to remove all those static accesses to the Field class

Not harder than with your implementation, you just need the field object everywhere, that's all. 

> And last of all you should check whether the main design has improved or not

Honestly, I think it got worse. field->getTheirGoal() is way more intuitive than `(*field)[THEIR_GOAL]`. Everyone loses autofill in their IDE, it is totally invisible what properties the `field` class has for developers, which is bad for maintainability. I mean, you make something quite complicated while it could be created by just making some methods in field, which is the obvious and easy solution. I wonder what others think about this, and it's really the call of the your software team.

I think replacing things like `field->getDistanceToOurGoal(...)` with `fieldComputation::getDistanceToOurGoal(*field, ...)` is fine though. Our controlUtils functions work just like that and are super useful.

HaicoDorenbos:
> As you requested my review, I have some comments.
> 
> > And because of this seperation we have a major performance increase, because the FieldComputation is no longer responsible for accessing/copying the Field class
> 
> I very much doubt this. Previously we were accessing pointers and had one function indirection, you have a map lookup. Both are not significant to the performance, and I am quite sure it is not a major performance increase. Have you run a profiler for comparison?
 
Good point :+1:, I will check the running time with the profiler next time to see the difference in running time. 

> > With the current state of the code it is not very easy to remove all those static accesses to the Field class
> 
> Not harder than with your implementation, you just need the field object everywhere, that's all.

Indeed but FieldComputation has now been separated for Field, so instead of 2 classes being linked statically we now only have 1 class linked statically.

> > And last of all you should check whether the main design has improved or not
> 
> Sadly, I think it got a little bit worse. field->getTheirGoal() is, for me, more intuitive than `(*field)[THEIR_GOAL]`. Developers lose autofill in their IDE, it is totally invisible what properties the `field` class has for developers, which is bad for maintainability. I mean, you make something quite complicated (we even need a custom hash function) while it could be created by just making some methods in field, which is the obvious and easy solution. I wonder what others think about this, and it's really the call of the your software team.
> 

- On the first point we already had a discussion with Jesse, Rolf and Paul and we decided that this approach with just 4 getters is much better than having 20 getters. We mainly decided this, because an error in code is easily made and with just 4 getters it is easier to check if these getters work or not. Also to make the access fast we used an unordered_map instead of a map, so there is only a little bit more overhead. But my vision is that in general it is good to sacrifice a slightly bit performance to increase readability of the code (e.g. it would be the best to write all your code in 1 function for performance, because function calls can also give some time overhead, however it is terrible for readability and maintenance). Moreover the array access operator overloading idea was from Paul and I agreed with him that it was much better than using get function calls. However I will check if a pass by reference instead of pointer will make it more readable and otherwise I write an additional line like Field field = *field; to make it more readable, because I indeed agree that (*field) is not very readable. :+1:
- Regarding the second point that the code is now more complicated because of the hash function: my vision is that Travis should there be for us, we should not be there for Travis. In general I do not think that the hash thing is that ugly (and I appreciate it that you fixed it), but if Travis continues to force us to make the code less readable/maintainable then we should stop using Travis. Travis was meant to make the code more maintainable, not the other way around.

> I think replacing things like `field->getDistanceToOurGoal(...)` with `fieldComputation::getDistanceToOurGoal(*field, ...)` is fine though. Our controlUtils functions work just like that and are super useful.

But if the field computation class get too much functionality then I will of course split it up in an interface instead of a large class with a lot of static functions.

rolfvdhulst:
Aside from other points: I'd like to point out you have duplicate header guards in Field.h and FieldComputation.h so at least one of the two classes you did a ton of work on is not even being compiled.

rolfvdhulst:
I mostly agree with lukas here.

The way I would envision we should make this class work is simply a class/almost POD with many members that are only const ref accessible. Then we have a small 'fieldmanager' that simply keeps track of the latest field, and passes it to our AI. By having a global object/singleton with a pointer we are making lives really difficult for ourselves. Ideally, we should simply pass this to the behavior trees instead of pointers, in my opinion.  I feel like with the amount of changes you did here you might as well have done that. The pointer indirects (*field)[ENUM ACCESS] calls are really ugly IMO, also because it is not clear from the code whether or not the object you're passing is implicitly copied or not in many cases.

rolfvdhulst:
The FieldComputations class is nice though.

HaicoDorenbos:
> Aside from other points: I'd like to point out you have duplicate header guards in Field.h and FieldComputation.h so at least one of the two classes you did a ton of work on is not even being compiled.

I looked at it, it was not a duplicate guard, however I still changed it to avoid confusion

Yuhanun:
Right, I really don't want to repeat what was already said, but I completely agree with most of the comments on this. I personally do think that a class with only static functions should be a namespace but that doesn't really matter, not in performance, nor in practical usage. I also think singletons are questionable, since they're just glorified global variables, even if they're in scott meyer's pattern.

About the performance however, I was reading through the comments and noticed that you wanted to run a profiler next time for optimization purposes, that's good, because just claiming something is faster isn't exactly proof.

I've gone ahead and taken a look at the classes you created, and I do not agree with your decisions, you decided to change the FieldMessage from a class with regular members, to a class with hashmaps, not only is this a huge performance decrease, this will decrease the readability and code completion, you simply do not get code completion for these fields, as they are not readable by most conventional IDE's, including CLion, VSCode and visual studio.

I've gone ahead and benchmarked a simple case, 

http://quick-bench.com/EG4wGGcCfG-7qDUwPY-0TuhQYno 

as you can see, the allocation of the hashmap and lookup times of the hashmap are way bigger than anything that a struct could ever cause with data like strings, vector2's and doubles. The way structs are located in memory is fairly simple, you reserve a block of memory and say "ok here are sizeof(mystruct) bytes which are reserved by me", structures dont exist in memory, which is the reason why things like python operate with hashmaps, and C++ simply has compiletime offsets for members, greatly improving performance over something like LUA, python or javascript.

Another reason why this is bad, is because we are currently working towards objects with an immutable state, meaning that every time you get a new FieldMessage you have to copy the current field to history, and then have to re-construct another instance of your hashmap, resulting in huge performance hits, runtime. I don't think I have to tell you that since this occurs 60 times per second, at least, this is not a smart idea.

I get that you want optional states for the enum values, hence I suggest a structure that simply has optional fields (std::optional).

We can talk about this after the break if you'd like, if you'd like me to elaborate on why I think that this is a bad idea, and why this shouldn't be performed, and maybe I can give another alternative or two.

include/roboteam_ai/world/Field.h
```diff
@@ -1,58 +1,253 @@
...
+
+    // Used to convert field arc name, in string format, to the corresponding FieldArcName enum value
+    std::unordered_map<std::string, FieldArcName> CONVERT_TO_FIELD_ARC_NAME = {
+        {"center_circle", CENTER_CIRCLE},
+    };
 
-} // world
-} // ai
-} // rtt
+private:
```
RobotJente: Double private keyword

src/interface/widgets/widget.cpp
```diff
@@ -152,11 +152,13 @@ void Visualizer::drawFieldLines(QPainter &painter) {
     painter.setPen(Constants::FIELD_LINE_COLOR());
     painter.setBrush(Qt::transparent);
     // draw lines
-    for (auto &item : field.getField_lines()) {
-        auto &line = item.second;
-        rtt::Vector2 start = toScreenPosition(line.begin);
-        rtt::Vector2 end = toScreenPosition(line.end);
-        painter.drawLine(start.x, start.y, end.x, end.y);
+    auto fls = field.getField_lines();
```
Yuhanun: What's up with this naming convention?...

include/roboteam_ai/world/Field.h
```diff
@@ -214,7 +225,7 @@ class Field {
      * Get all the lines of the field
      * @return A map which contains all field lines
      */
-    std::unordered_map<FieldLineName, FieldLineSegment> getField_lines();
+    std::optional<FieldLineSegment>* getField_lines();
```
Yuhanun: this LineSegment shouldn't be mutated, return a constant version

include/roboteam_ai/bt/Node.hpp
```diff
@@ -4,6 +4,7 @@
 #include <vector>
 #include <cstdio>
 #include <cstdarg>
+#include <include/roboteam_ai/world/Field.h>
```
mrlukasbos: This is very bad for compile time: everytime you make a change in field (and the dependencies of field) you have to recompile every skill, condition and tree.

mrlukasbos: As you do a fwd declaration you don't have to include it here.
you can safely include it in de .cpp file.

include/roboteam_ai/bt/Node.hpp
```diff
@@ -12,10 +13,8 @@ namespace rtt {
     namespace ai {
     namespace world {
         class World;
-
-        class Field;
-
     }
+    class FieldComputations;
```
mrlukasbos: fieldComputations is not namespaced?

include/roboteam_ai/world/Field.h
```diff
@@ -1,58 +1,264 @@
...
+    THEIR_GOAL_CENTER, // The middle point of the opponents goal (this point is on the right line).
+    LEFT_PENALTY_POINT, // The penalty point from which penalties are made towards our goal.
+    RIGHT_PENALTY_POINT, // The penalty point from which penalties are made towards the opponents goal.
+    OUR_BOTTOM_GOAL_SIDE, // The bottom most point of our goal (this point is on the left line).
+    OUR_TOP_GOAL_SIDE, // The top most point of our goal (this point is on the left line).
+    THEIR_BOTTOM_GOAL_SIDE, // The bottom most point of the opponents goal (this point is on the right line).
+    THEIR_TOP_GOAL_SIDE, // The top most point of the opponents goal (this point is on the right line).
+
+    // Enum item used to indicate the number of Field Vector Names (make sure that this always is the last item of this enum)
+    NUMBER_FIELD_VECTOR_NAMES
```
mrlukasbos: No! Why are you doing this!? By evading the getters and setters, which were fine, you now have to go to these kinds of solutions? These things are really indications that you are overcomplicating things.

include/roboteam_ai/world/Field.h
```diff
@@ -1,58 +1,264 @@
...
+     * @return A map which contains all field lines
+     */
+    const std::optional<FieldLineSegment>* getField_lines();
+
+private:
+    /**
+     * Convert a float measured in millimeters to meters (is needed, because proto message contains values measured in
+     * millimeters).
+     */
+    static float mm_to_m(float scalar);
```
mrlukasbos, line 267: the mm_to_m is also available in utils I believe

mrlukasbos (changes requested):
I just cannot approve this. Honestly, this PR, and also the previous field PR, makes things only worse for everyone. 

- There is no performance gain 
- There is no autocomplete --> The contents of field become invisible while they were fine.
- You have to resort to doing shady hacks with enums
- It does not solve a problem! Therefore, it only creates them.

include/roboteam_ai/world/Field.h
HaicoDorenbos: I will do this in later pull request. This clone was not introduced by this pull request, it was already inside the code.

include/roboteam_ai/bt/Node.hpp
HaicoDorenbos: It is, but it is not namespaced in world. It is only namespaced in rtt::ai

HaicoDorenbos:
The most notorious pull request is now ready to be merged. It has caused a lot of discussion, but hopefully we now all agree on this final version. The map things has been reverted and replaced by a lot of getters. Moreover there are no more static references/singletons in the Field or FieldComputation class :innocent:. It has been replaced by calling a static getter in the IOManager a few times (per main component).

src/world/Field.cpp
```diff
@@ -143,7 +130,7 @@ Vector2 Field::operator[](FieldVectorName vectorName) const {
     }
 }
 
-const std::optional<FieldLineSegment>* Field::getField_lines(){
+const std::optional<FieldLineSegment>* Field::getField_lines() const {
```
Yuhanun: Naming convention s weird again
getField_lines -> getFieldLines() or get_field_lines(), Roboteam sticks to camelCase, so yeee

src/io/IOManager.cpp
```diff
@@ -82,6 +81,10 @@ void IOManager::handleReferee(proto::SSL_Referee & refData) {
   }
 }
 
+const Field &IOManager::getField() {
+    std::lock_guard<std::mutex> lock(geometryMutex);
```
Yuhanun: Unnessecary lock (?)

src/io/IOManager.cpp
```diff
@@ -82,6 +81,10 @@ void IOManager::handleReferee(proto::SSL_Referee & refData) {
...
+    return field;
+}
 
 const proto::World &IOManager::getWorldState() {
     std::lock_guard<std::mutex> lock(worldStateMutex);
```
Yuhanun: Same here I think, you don't have to lock it just to return the reference, unless you're expecting the location to change during returning, which doesn't happen in field iirc

src/interface/widgets/RobotsWidget.cpp
```diff
@@ -29,6 +29,7 @@ RobotsWidget::RobotsWidget(QWidget* parent) : QWidget(parent){
 }
 
 void RobotsWidget::updateContents(Visualizer* visualizer) {
+    const Field &field = io::io.getField();
     auto us = rtt::ai::world::world->getUs();
```
Yuhanun, line 31: auto const&

src/coach/midField/MidFieldCoach.cpp
```diff
@@ -171,10 +169,10 @@ bool MidFieldCoach::isRobotAlreadyBeingHarassed(const world::World::RobotPtr &op
     return alreadyBeingHarassed;
 }
 
-MidFieldCoach::Target MidFieldCoach::standFree(const RobotPtr &thisRobot) {
+MidFieldCoach::Target MidFieldCoach::standFree(const Field &field, const RobotPtr &thisRobot) {
     Target target;
```
Yuhanun: default initialize this to prevent UB values

src/coach/heuristics/CoachHeuristics.cpp
```diff
@@ -84,10 +83,10 @@ double CoachHeuristics::calculateBehindBallScore(const Vector2 &position, const
     }
 }
 
-double CoachHeuristics::calculatePassDistanceToBallScore(const Vector2 &position,
+double CoachHeuristics::calculatePassDistanceToBallScore(const Field &field, const Vector2 &position,
         const CoachHeuristics::WorldData &world) {
     auto ball = world.ball;
```
Yuhanun, line 79: const&

Yuhanun: Also, use the getter, not the value itself.

src/coach/PassCoach.cpp
```diff
@@ -143,7 +143,7 @@ bool PassCoach::validReceiver(const RobotPtr& passer, const RobotPtr& receiver,
         if (receiver->pos.x < -RECEIVER_MAX_DISTANCE_INTO_OUR_SIDE) {
             return false;
         }
-        auto MIN_PASS_DISTANCE = std::max((double) Field::get_field()[GOAL_WIDTH] / 2, SMALLEST_MIN_PASS_DISTANCE);
+        auto MIN_PASS_DISTANCE = std::max((double) field[GOAL_WIDTH] / 2, SMALLEST_MIN_PASS_DISTANCE);
```
Yuhanun: I thought you resolved these, I thought you changed this kinda stuff to 
field.getGoalWidth() etc, this still causes issues with code suggestion etc

src/coach/PassCoach.cpp
```diff
@@ -72,14 +72,14 @@ bool PassCoach::isPassed() {
...
-int PassCoach::determineReceiver(int passerID) {
+int PassCoach::determineReceiver(const Field &field, int passerID) {
     coach::PassScore passScore;
     double bestScore = 0;
     int bestRobotID = -1;
     auto passer = world::world->getRobotForId(passerID, true);
     for(auto &robot : world::world->getUs()) {
-        if (!validReceiver(passer, robot)) continue;
-        double score = passScore.calculatePassScore(robot->pos);
+        if (!validReceiver(field, passer, robot)) continue;
```
Yuhanun, line 68: if (!valid...)
    continue;
... 


For readability sake

src/analysis/GameAnalyzer.cpp
```diff
@@ -212,23 +176,24 @@ void GameAnalyzer::stop() {
...
     for (auto robot : robots) {
-        robotDangers.emplace_back(robot, evaluateRobotDangerScore(robot, ourTeam));
+        robotDangers.emplace_back(robot, evaluateRobotDangerScore(field, robot, ourTeam));
     }
 
     std::sort(robotDangers.begin(), robotDangers.end(),
-            [](std::pair<RobotPtr, RobotDanger> a,
-                    std::pair<RobotPtr, RobotDanger> b) {
-              return a.second.getTotalDanger() > b.second.getTotalDanger();
+            [field](std::pair<RobotPtr, RobotDanger> a, std::pair<RobotPtr, RobotDanger> b) {
```
Yuhanun: This creates a copy of field

[&field](...) {} 

Would capture it by reference instead of by value

src/analysis/GameAnalyzer.cpp
```diff
@@ -212,23 +176,24 @@ void GameAnalyzer::stop() {
...
+        const Field &field = io::io.getField();
+        generateReportNow(field);
         std::this_thread::sleep_for(delay);
     }
 }
 
-std::vector<std::pair<GameAnalyzer::RobotPtr, RobotDanger>> GameAnalyzer::getRobotsSortedOnDanger(bool ourTeam) {
+std::vector<std::pair<GameAnalyzer::RobotPtr, RobotDanger>> GameAnalyzer::getRobotsSortedOnDanger(const Field &field,
+        bool ourTeam) {
     auto robots = ourTeam ? world::world->getUs() : world::world->getThem();
```
Yuhanun, line 183: auto const&

src/analysis/GameAnalyzer.cpp
```diff
@@ -212,23 +176,24 @@ void GameAnalyzer::stop() {
 void GameAnalyzer::loop(unsigned delayMillis) {
     std::chrono::milliseconds delay(delayMillis);
     while (! stopping) {
-        generateReportNow();
+        const Field &field = io::io.getField();
```
Yuhanun, line 176: Isn't this still kind of a glorified global

include/roboteam_ai/utilities/RobotDealer.h
```diff
@@ -10,6 +10,7 @@
 #include <set>
 #include <mutex>
 #include <vector>
+#include <include/roboteam_ai/world/Field.h>
```
Yuhanun: consistency in includes

<include/roboteam_ai/world/Field.h>
vs
"world/Field.h"

Yuhanun: Can we decide on one include convention within roboteam? :eyes: Not sure how enforcable it is but clang-format is already a good step in the good direction (Thanks emiel <3 )

src/bt/tactics/DefaultTactic.cpp
```diff
@@ -57,7 +57,7 @@ void DefaultTactic::claimRobots(int amount) {
     // for the amount of robots we still need
     for (int i = 0; i < amount; i++) {
         auto toClaim = getNextClaim();
-        robotIDs.insert(dealer::claimRobotForTactic(toClaim.second, toClaim.first, name));
+        robotIDs.insert(dealer::claimRobotForTactic(*field, toClaim.second, toClaim.first, name));
```
Yuhanun, line 58: I'd assert field for nullptr tbh, but not really nessecary

src/analysis/GameAnalyzer.cpp
```diff
@@ -62,58 +63,28 @@ BallPossession GameAnalyzer::convertPossession(rtt::ai::BallPossession::Possessi
...
     auto robots = ourTeam ? simulatedWorld.us : simulatedWorld.them;
     double total = 0.0;
     for (auto robot : robots) {
-        total += world::field->getDistanceToGoal(ourTeam, robot->pos);
-    }
-    return (total/robots.size());
-}
-
-/// return the attackers of a given team sorted on their vision on their opponents goal
-std::vector<std::pair<GameAnalyzer::RobotPtr, double>> GameAnalyzer::getAttackersSortedOnGoalVision(bool ourTeam,
```
RobotJente: Why did you remove this function?

src/io/IOManager.cpp
mrlukasbos: That does happen.

src/io/IOManager.cpp
mrlukasbos: No, iomanager stores the most recently received field in that variable, which is a mutation obviously. Which can happen during a read

include/roboteam_ai/world/Field.h
```diff
@@ -1,58 +1,298 @@
...
+    double getFieldWidth() const;
+    double getFieldLength() const;
+    double getGoalWidth() const;
+    double getGoalDepth() const;
+    double getBoundaryWidth() const;
+    double getCenterY() const;
+    double getLeftmostX() const;
+    double getRightmostX() const;
+    double getBottommostY() const;
+    double getTopmostY() const;
```
RobotJente, line 213: Glad these are back!

include/roboteam_ai/utilities/RobotDealer.h
RobotJente: Agreed. There's probably something for this. I'll make it an issue.

src/coach/OffensiveCoach.cpp
```diff
@@ -186,61 +182,38 @@ Vector2 OffensiveCoach::getShootAtGoalPoint(const Vector2 &fromPoint) {
...
-    if (leftDif<=rightDif){
-        if((leftPoint - keeperPos).length() >= 0.4){
-            return std::make_pair(leftPoint,true);
-        }
-        return std::make_pair(rightPoint,false);
-    }
-    if(( rightPoint - keeperPos).length() >= 0.4){
-        return std::make_pair(rightPoint,true);
-    }
-    return std::make_pair(leftPoint,false);
```
RobotJente: Why did you remove this function?

src/coach/PassCoach.cpp
RobotJente: Agreed

RobotJente (changes requested):
I agree with the direction this PR is going. I like the statics and the fact that field is now pass by reference. The extra documentation is super helpful. The return of the getters is saving the roboteam galaxy. The only reason I'm not approving is because you removed some functions. If you resolve those comments I will approve

src/coach/OffensiveCoach.cpp
HaicoDorenbos: The same as above. This code is dead.

src/analysis/GameAnalyzer.cpp
HaicoDorenbos: This function is not called (dead code). Dead code needs to be removed, since it is being reviewed/read by developers when it is not used.

src/analysis/GameAnalyzer.cpp
HaicoDorenbos: No

src/world/Field.cpp
HaicoDorenbos: This was already present in the previous version of the code

src/coach/PassCoach.cpp
HaicoDorenbos: This was not an error by this commit....

src/conditions/BallKickedToOurGoal.cpp
```diff
@@ -21,8 +21,8 @@ bt::Node::Status BallKickedToOurGoal::onUpdate() {
     }
 
     // determine the goalsides
-    Vector2 goalCentre = field->get_field().get(OUR_GOAL_CENTER);
-    double goalWidth = field->get_field().get(GOAL_WIDTH);
+    Vector2 goalCentre = (*field).getOurGoalCenter();
```
Yuhanun, line 21: field->getOurGoalCenter()

src/conditions/BallKickedToOurGoal.cpp
```diff
@@ -21,8 +21,8 @@ bt::Node::Status BallKickedToOurGoal::onUpdate() {
...
+    double goalWidth = (*field).getGoalWidth();
```
Yuhanun, line 22: field->getGoalWidth()

src/conditions/BallKickedToOurGoal.cpp
```diff
@@ -21,8 +21,8 @@ bt::Node::Status BallKickedToOurGoal::onUpdate() {
...
     double margin = BALL_TO_GOAL_MARGIN;
```
Yuhanun, line 23: why not use BALL_TO_GOAL_MARGIN direct? allows for inlining.

mrlukasbos (commented):
Why are there so many formatting changes, changing everything from 4 spaces to 2 spaces?

Yuhanun:
Right, so, I haven't looked at the full code but I have a question or two.

After all these changes, comments, reverts etc. etc. I have a question.

What actually changed, what improved, what got removed, what got ditched?

HaicoDorenbos:
One major important thing: the static reference to the Field class got removed and the FieldComputation class is now a seperate module (independent of the field). Also the Field is now better documented. And we can see if Field variables have been initialized. Moreover some clones have been removed and performance optimizations were made which made things a lot more clearer.

src/conditions/BallKickedToOurGoal.cpp
HaicoDorenbos: Will all be solved in the next PR

src/conditions/BallKickedToOurGoal.cpp
HaicoDorenbos: Will be solved in the next PR

HaicoDorenbos:
This pull request is merged, because delaying it would have caused a huge additional maintenance cost, since it is involved with a lot of different files and delaying it will cause a lot of merge conflicts. And for me it is harder to solve the merge conflicts in all your pull requests whereas for you it is more easier to solve the merge conflicts on your branch (the burden is spread over more developers).

HaicoDorenbos:
> Why are there so many formatting changes, changing everything from 4 spaces to 2 spaces?

Because I noticed that linter has not yet been applied on the Roboteam_AI branch and we decided to use linter.

PR #744, opened 2020-01-13
Refactor/roboteam world ownership

Yuhanun:
I'm working on refactoring the ownership models. 

It's not done yet (by far), but im opening this to get feedback already <3


emielsteerneman:
How did commits from 2017 end up in here??

Yuhanun:
> How did commits from 2017 end up in here??

No clue.

Yuhanun:
However, to me this is ready to be merged. Better merge this ASAP to get used to using it, it's fairly intuitive (std::unique_ptr style), accessing data that's directly on the robot / world / whatever would be done using 


RobotView view{ &robot }->getSomething();

Accessing members functions would be for example

World::instance()->getWorld().getRobotClosestToBall();

Summary of this pullrequest:

- World state now assumes a constant state
- World state does not get updated once constructed, however it will get pushed to history
- Any get() and set() operations on the POD types, such as getWorld(), is O(1), meaning it's a constant time complexity, unless indicated otherwise
- Accessing history is O(1)
- Everything is documented to an extent where it should be fairly easy to tell what's new and what's different
- The POD types are no longer meant to be accessed directly, however there's now Views over the POD types, allowing you to keep a "reference" to a POD type, which you cannot (without hacks) modify in any way. 

If this does end up getting merged, then please start using it ASAP so we can completely get rid of the old world, I'll be starting a new branch for doing that myself as soon as this gets accepted.

Old world files have been moved to world_old to not yet break backwards compatibility and to allow for quick merging, which would allow us to get used to (and start using) the new classes / views before the robocup.

I'm expecting a relatively large performance gain as soon as we're fully transferred, due to the ommitance of std::shared_ptr's copy semantics, which has a volatile atomic counting operation in it.

I'm aware that Travis does not yet build, not sure why.

src/world/views/robot_view.cpp
```diff
@@ -0,0 +1,50 @@
+//
+// Created by john on 1/14/20.
+//
+
+#include <include/roboteam_ai/utilities/Constants.h>
+#include "include/roboteam_ai/world/views/robot_view.hpp"
```
Ratoone: Include does not need the whole path

PR #747, opened 2020-01-16
Re-added deprecated methods, because the new method gave segfaults

rttwente:
axisY() and axisX() are deprecated
A non-deprecated method can be found here https://stackoverflow.com/questions/53812267/qtcharts-axisx-depreciated
However, for some reason it gave segfaults the last time I tried this (jan 16 2020)


RobotJente (approved):
Looks fine to me. I think we should try the fix again though, but fixing development first is a good idea 😊

PR #750, opened 2020-01-20
Travis/docker fix

Yuhanun:
Changed the travis builds to now use a docker image with pre-built dependencies.


Yuhanun:
Ready for merge.

RobotJesse:
It looks like this branch is ahead of development. Is it possible you added * to git? I think the only file that is meant to be changed is the Travis.yml file right?

PR #756, opened 2020-01-23
Refactor/roboteam world ownership

Yuhanun:
Accidentally made a mistake in the new world code that was causing segfaults for some reason.

I hope that's resolved now.


Yuhanun:
Why is there this much shit again...

PR #758, opened 2020-01-23
Fix/roboteam world segfaults

Yuhanun:
Forgot to uncomment locks

### DIFF
--- a/src/world_new/World.cpp
+++ b/src/world_new/World.cpp
@@ -20,9 +20,8 @@ WorldData const &World::setWorld(WorldData &newWorld) noexcept {
 }
 
 void World::toHistory(WorldData &world) noexcept {
-    std::lock_guard mtx{ updateMutex };
     updateTickTime();
-    if (history.size() < HISTORY_SIZE && currentIndex < history.size()) {
+    if (history.size() < HISTORY_SIZE) {
         history.emplace_back(std::move(world));
     } else {
         history[currentIndex] = std::move(world);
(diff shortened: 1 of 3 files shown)


src/world_new/World.cpp
```diff
@@ -20,9 +20,8 @@ WorldData const &World::setWorld(WorldData &newWorld) noexcept {
 }
 
 void World::toHistory(WorldData &world) noexcept {
-//    std::lock_guard mtx{ updateMutex };
     updateTickTime();
-    if (history.size() < HISTORY_SIZE && currentIndex < history.size()) {
+    if (history.size() < HISTORY_SIZE) {
         history.emplace_back(std::move(world));
     } else {
         history[currentIndex] = std::move(world);
```
mrlukasbos, line 27: is it guaranteed that currentIndex has a proper value? 

use `history.at(currentIndex)` instead to make sure you crash instead of UB in such a case?

Yuhanun: currentIndex++;
currentIndex %= HISTORY_SIZE;

So i'd assume it's a proper value.

I can do .at() but we don't do exceptions

mrlukasbos: OK. still would be nice to have this class tested. Are there plans of testing this class anytime?

Yuhanun: Yup, I'm going to write tests myself + we can test using grsim n stuff, it works fine for me actually

PR #760, opened 2020-01-23
remove json from AI and use apt/brew json instead

mrlukasbos:
The json library was both in utils and AI. This PR removes the one from AI.


mrlukasbos (commented):
Looks good to me!

mrlukasbos:
I have to merge this with admin rights, as roboteam_utils is already updated and does not contain the json dependency anymore.

PR #761, opened 2020-01-23
simplify directory structure and cleanup cmake lists 

mrlukasbos:
The directory structure became quite confusing. This is a proposal to structure it a bit. 

- I changed the naming of some directories to be more consistent. 
- I moved the Settings and IO class to `utilities`
- I cleaned up the CMake 

This could be more rigorous but I did not want change too much


RobotJente:
I think it looks good, I like the renaming of PlaysObjects to play-utilties. Do you have a proposal to change more? If not, then I think this is good to merge

mrlukasbos:
I do have some proposals to change more, but PR's should be kept small, so i'd rather do such changes in a different PR.

RobotJente (approved):
Looks good!

PR #765, opened 2020-01-24
Refactor/num trees

Ratoone:
A new structure for the position control. This is supposed to completely replace the NumTreesPosControl. Right now, the file was split into a path generation, path tracking, collision detector, and a container for all of them. For testing purposes, this was connected to the manual drive, the elimination of the old structure being work in progress.

Apart from the old implementation, this PR contains an implementation of a Voronoi path generation and a simple basic path tracking.


test/ControlTests/PosControllerTests/VoronoiPathPlanningTest.cpp
```diff
@@ -0,0 +1,74 @@
...
+    VoronoiPathPlanning posControl = VoronoiPathPlanning(100, 100, robotPositions);
+    robots[0]->pos = {9, 2};
+    robots[1]->pos = {1, 9};
+    robots[2]->pos = {1, 3};
+    robots[3]->pos = {5, 5};
+    robots[4]->pos = {0, 0};
+    posControl.computePath({0,0},{10,10});
+    auto graph = posControl.getGraphAdjacencyList();
+    ASSERT_EQ(graph.size(), 9);
+    ASSERT_FALSE(graph.find(rtt::Vector2(6,10)) == graph.end());
```
rolfvdhulst: Minor but when doing tests we typically prefer EXPECT as it does not exit the test immediately, so that we can catch more than one bug using one test run. The only reason it exists is to prevent the test from e.g. deadlocking/segfaulting.

src/control/positionControl/pathTracking/BasicPathTracking.cpp
```diff
@@ -0,0 +1,21 @@
+//
+// Created by ratoone on 12-11-19.
+//
+
+#include <control/positionControl/pathTracking/BasicPathTracking.h>
+
+void BasicPathTracking::trackPath(const rtt::Vector2 &currentPosition, const rtt::Vector2 &currentVelocity,
+        std::list<rtt::Vector2> &pathPoints, rtt::Vector2 &outputVelocity, double &outputAngle) {
```
rolfvdhulst: Probably better to use std::vector<> which gets inlined more easily and is a LOT faster. Since Vector2 is <64 bytes std::vector<> is much much faster

rolfvdhulst: Typically we like return's better than passing elements by reference which are mutable. You could use the Position class (look at roboteam_utils) for this, which is basically a Vector2+angle struct

src/control/positionControl/pathTracking/NumTreesTracking.cpp
```diff
@@ -0,0 +1,23 @@
...
+//
+
+#include "control/positionControl/pathTracking/NumTreesTracking.h"
+
+namespace rtt::ai::control{
+
+void
+NumTreesTracking::trackPath(const rtt::Vector2 &currentPosition, const rtt::Vector2 &currentVelocity,
+                                              std::list<rtt::Vector2> &pathPoints, rtt::Vector2 &outputVelocity,
+                                              double &outputAngle) {
```
rolfvdhulst: Again personally I would prefer this returning rtt::Position rather than being void.

src/control/positionControl/pathTracking/NumTreesTracking.cpp
```diff
@@ -0,0 +1,23 @@
...
+    int lookAhead = std::min(pathPoints.size()-1, stepsAhead);
+    Vector2 currentTarget = *std::next(pathPoints.begin(), lookAhead);
+    if (pathPoints.size() > 1 && (currentTarget - currentPosition).length() < minimumDistance){
+        pathPoints.erase(pathPoints.begin(), std::next(pathPoints.begin(),lookAhead));
+    }
```
rolfvdhulst: This definitely needs a comment explaining what/why you are doing this.

src/control/positionControl/pathTracking/PidTracking.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+PidTracking::PidTracking(){
+    xPid.setMaxIOutput(maxVelocity);
+    yPid.setMaxIOutput(maxVelocity);
+}
+
+void PidTracking::trackPath(const rtt::Vector2 &currentPosition, const rtt::Vector2 &currentVelocity,
+                                              std::list<rtt::Vector2> &pathPoints, rtt::Vector2 &outputVelocity,
+                                              double &outputAngle) {
+    if (pathPoints.size() > 1 && (pathPoints.front() - currentPosition).length() < minimumDistance){
```
rolfvdhulst: This is the second time I'm reading this exact line. You might want to make it a function on pathPoints and make pathPoints a struct/class?

src/control/positionControl/pathTracking/PidTracking.cpp
```diff
@@ -0,0 +1,33 @@
...
+namespace rtt::ai::control{
+
+PidTracking::PidTracking(){
+    xPid.setMaxIOutput(maxVelocity);
+    yPid.setMaxIOutput(maxVelocity);
+}
+
+void PidTracking::trackPath(const rtt::Vector2 &currentPosition, const rtt::Vector2 &currentVelocity,
+                                              std::list<rtt::Vector2> &pathPoints, rtt::Vector2 &outputVelocity,
+                                              double &outputAngle) {
```
rolfvdhulst: Again I personally dislike references being used to return values.

src/skills/DriveWithInterface.cpp
```diff
@@ -5,23 +5,26 @@
...
+namespace rtt::ai {
 DriveWithInterface::DriveWithInterface(string name, bt::Blackboard::Ptr blackboard) : Skill(name, blackboard) {}
 Skill::Status DriveWithInterface::onUpdate() {
     if (interface::Output::usesRefereeCommands()) {
         return Status::Failure;
     }
+    //TODO: workaround until world ownership changes
+    if (numTreeGtp == nullptr){
+        numTreeGtp = world_new::World::instance()->getRobotPositionController();
+    }
```
rolfvdhulst: Make this is an issue when we merge this PR (or do it now).

src/world_new/World.cpp
```diff
@@ -86,4 +86,12 @@ uint64_t World::getTimeDifference() const noexcept {
 robot::RobotControllers &World::getControllersForRobot(uint8_t id) noexcept {
     return robotControllers[id];
 }
+
+ai::control::PositionControl* World::getRobotPositionController() noexcept {
+    if (positionControl == nullptr){
+        positionControl = new ai::control::PositionControl(this->currentWorld->getRobots());
+    }
+    return positionControl;
```
rolfvdhulst: This will only create positioncontrol once with all the set of seen robots but afterwards if new robots/other robots enter the field how does this communicate to it?

This does not make very much sense to me.

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    std::list<rtt::Vector2> pathPoints;
+    for (rtt::Vector2 backtrack = targetPosition; parentVector.find(backtrack) != parentVector.end(); backtrack = parentVector[backtrack]){
+        pathPoints.push_back(backtrack);
+    }
+    std::reverse(pathPoints.begin(), pathPoints.end());
+    return pathPoints;
+}
+
+std::list<rtt::Vector2> VoronoiPathPlanning::computePath(const rtt::Vector2 &robotPosition, const rtt::Vector2 &targetPosition) {
```
rolfvdhulst: std::vector

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+        graphAdjacencyList[closestTargetPoint].push_back({targetPosition, minTargetDist});
+    }
+}
+
+const std::unordered_map<rtt::Vector2, std::list<GraphNode>, hashPoint> &
+VoronoiPathPlanning::getGraphAdjacencyList() const {
+    return graphAdjacencyList;
+}
+
+std::list<rtt::Vector2> VoronoiPathPlanning::generatePathDijkstra(const rtt::Vector2& initialPosition, const rtt::Vector2& targetPosition) {
```
rolfvdhulst: std::vector

include/roboteam_ai/world_new/World.hpp
```diff
@@ -180,6 +188,11 @@ namespace rtt::world_new {
          * Duration between ticks
          */
         uint64_t tickDuration{};
+
+        /**
+         * The position controller, initially null
+         */
+        ai::control::PositionControl *positionControl = nullptr;
```
rolfvdhulst: Although this seems fine I personally am not a fan of controllers being in 'World'. World should be a place where data lives, not where execution of control lives.

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+                continue;
+            }
+            if(distanceVector[adjacentNode.nextNodePosition] > distanceVector[currentNode] + adjacentNode.distance){
+                distanceVector[adjacentNode.nextNodePosition] = distanceVector[currentNode] + adjacentNode.distance;
+                parentVector[adjacentNode.nextNodePosition] = currentNode;
+            }
+        }
+    }
+
+    std::list<rtt::Vector2> pathPoints;
```
rolfvdhulst: std::vector

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+    for (rtt::Vector2 backtrack = targetPosition; parentVector.find(backtrack) != parentVector.end(); backtrack = parentVector[backtrack]){
+        pathPoints.push_back(backtrack);
+    }
+    std::reverse(pathPoints.begin(), pathPoints.end());
```
rolfvdhulst: This needs a comment.

include/roboteam_ai/control/positionControl/CollisionDetector.h
```diff
@@ -0,0 +1,60 @@
...
+     * @param nextPoint
+     * @return
+     */
+    bool isRobotCollisionBetweenPoints(const Vector2& initialPoint, const Vector2& nextPoint);
+
+    bool isPointInsideField(const Vector2 &point);
+
+    bool isPointInDefenseArea(const Vector2 &point);
+
+    std::vector<const Vector2 *> getRobotPositions();
```
rolfvdhulst: Copying for small data types like Vector2 is probably equally fast and much easier to work with + it keeps data aligned for autovectorization.

include/roboteam_ai/control/positionControl/CollisionDetector.h
```diff
@@ -0,0 +1,60 @@
...
+#include "control/ControlUtils.h"
+
+
+namespace rtt::ai::control{
+
+/**
+ * Checks for collision between points and different components of the field: robots, defence area, etc.
+ * Check isCollisionBetweenPoints method for more details
+ */
+class CollisionDetector {
```
rolfvdhulst, line 19: I really like this abstraction/premise of this class. This is a step in the right direction. However:
Personally would split it into two classes based on experience of things getting messy down the road (since you did not implement e.g. referee rules in here yet)

One 'generator' which sole task is to generate obstacles from a worldstate,field, referee rules etc. 
And one 'checker' which then takes an array of obstacles and checks whether or given robots + positions and paths intersect. It's sole task is to take a set of obstacles and determine scores (e.g. how badly does a path interfere with obstacles?)

include/roboteam_ai/control/positionControl/CollisionDetector.h
rolfvdhulst: Even if you insist on using pointers I would much rather see references to prevent memory leaks.

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+    computeDiagram(robotPosition, targetPosition);
+    generateGraphFromDiagram();
+    completeGraphWithOriginDestination(robotPosition, targetPosition);
+    return generatePathDijkstra(robotPosition, targetPosition);
+}
+
+rtt::Vector2 VoronoiPathPlanning::convertFromJcvPoint(jcv_point point) {
+    return rtt::Vector2(std::round(1000.0*point.x)/1000, std::round(1000.0*point.y)/1000);
+}
+
```
rolfvdhulst: No need to use extra rtt:: qualifier

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+std::list<rtt::Vector2> VoronoiPathPlanning::computePath(const rtt::Vector2 &robotPosition, const rtt::Vector2 &targetPosition) {
+    computeDiagram(robotPosition, targetPosition);
+    generateGraphFromDiagram();
+    completeGraphWithOriginDestination(robotPosition, targetPosition);
+    return generatePathDijkstra(robotPosition, targetPosition);
+}
+
+rtt::Vector2 VoronoiPathPlanning::convertFromJcvPoint(jcv_point point) {
+    return rtt::Vector2(std::round(1000.0*point.x)/1000, std::round(1000.0*point.y)/1000);
+}
```
rolfvdhulst: Also, why in the hell would I round my point precision down? This function needs a comment to why it even exist and how this  'converts' since you are losing precision putting things through here.

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+const std::unordered_map<rtt::Vector2, std::list<GraphNode>, hashPoint> &
+VoronoiPathPlanning::getGraphAdjacencyList() const {
+    return graphAdjacencyList;
+}
+
+std::list<rtt::Vector2> VoronoiPathPlanning::generatePathDijkstra(const rtt::Vector2& initialPosition, const rtt::Vector2& targetPosition) {
+    std::unordered_map<rtt::Vector2, float, hashPoint> distanceVector;
+    std::unordered_map<rtt::Vector2, rtt::Vector2, hashPoint> parentVector;
+    distanceVector.insert({initialPosition,0});
+    std::list<rtt::Vector2> nodeQueue;
```
rolfvdhulst: std::vector is faster for finding and inserting like this.

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+                distanceVector.insert({adjacentNode.nextNodePosition, distanceVector[currentNode] + adjacentNode.distance});
+                parentVector.insert({adjacentNode.nextNodePosition, currentNode});
+                continue;
+            }
+            if(distanceVector[adjacentNode.nextNodePosition] > distanceVector[currentNode] + adjacentNode.distance){
+                distanceVector[adjacentNode.nextNodePosition] = distanceVector[currentNode] + adjacentNode.distance;
+                parentVector[adjacentNode.nextNodePosition] = currentNode;
+            }
+        }
+    }
```
rolfvdhulst: This needs a comment. It's not 'super' complicated but it's certainly not easy either.

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+        if(graphAdjacencyList.empty() || graphAdjacencyList.find(secondEdgePoint) == graphAdjacencyList.end()){
+            graphAdjacencyList[secondEdgePoint] = std::list<GraphNode>(1,(GraphNode){firstEdgePoint, distance});
+        }
+        else{
+            graphAdjacencyList[secondEdgePoint].push_back((GraphNode){firstEdgePoint, distance});
+        }
+    }
+}
+
+void VoronoiPathPlanning::completeGraphWithOriginDestination(const rtt::Vector2 &robotPosition, const rtt::Vector2 &targetPosition) {
```
rolfvdhulst: This whole function does almost exactly the same thing with robotPosition as it does with targetPosition. Consider making them one input and calling the function twice.

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+VoronoiPathPlanning::VoronoiPathPlanning(CollisionDetector& collisionDetector) : collisionDetector(collisionDetector){}
+
+void VoronoiPathPlanning::computeDiagram(const rtt::Vector2 &robotPosition, const rtt::Vector2 &targetPosition) {
+    auto robots = collisionDetector.getRobotPositions();
+    std::vector<jcv_point> robotPositions(robots.size());
+    std::transform(robots.begin(), robots.end(), robotPositions.begin(),
+            [](auto robot)-> jcv_point {return {(float)robot->x, (float)robot->y};});
+    jcv_rect playArea = {
+            (jcv_point){(float)std::min(robotPosition.x,targetPosition.x),(float)std::min(robotPosition.y,targetPosition.y)},
+            (jcv_point){(float)std::max(robotPosition.x,targetPosition.x),(float)std::max(robotPosition.y,targetPosition.y)}
```
rolfvdhulst: I can tell you from experience that many optimal paths are not limited to the rectangle defined by the targetPosition and the robotPosition. This is certainly not sufficient for object avoidance.

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
```diff
@@ -0,0 +1,214 @@
...
+    auto path = tracePath(robotPosition, targetPosition);
+    double goToTimeInFuture = 0.4;
+    auto targetPathPoint = static_cast<unsigned long>(goToTimeInFuture/DT);
+    if (path.size() < targetPathPoint) {
+        return std::list<Vector2>(1, targetPosition);
+    }
+    return path;
+}
+
+std::list<Vector2> NumTreesPlanning::tracePath(const rtt::Vector2 &currentPosition, const rtt::Vector2 &targetPosition) {
```
rolfvdhulst: This function is a 100 lines of code. If you're going to refactor it please reduce it to ~30 max.

rolfvdhulst: Although feel free to disregard this as you did not change much it seems.

src/control/positionControl/PositionControl.cpp
```diff
@@ -0,0 +1,53 @@
...
+    pathPlanningAlgorithm = new NumTreesPlanning(*collisionDetector);
+    pathTrackingAlgorithm = new NumTreesTracking();
+}
+
+//TODO: add projection to outside defence area (project target position)(is this really needed?)
+RobotCommand
+PositionControl::computeAndTrackPath(world::Field *field, int robotId, const Vector2 &currentPosition,
+                                     const Vector2 &currentVelocity, const Vector2 &targetPosition) {
+    //TODO: this is a workaround caused by the fact that the field is not global
+    collisionDetector->setField(field);
```
rolfvdhulst: What? Pls don't use raw pointers but copy by reference or just copy if you can because this might cause UB at any point in time.

src/control/positionControl/PositionControl.cpp
```diff
@@ -0,0 +1,53 @@
...
+    interface::Input::drawData(interface::Visual::PATHFINDING, path, Qt::green, robotId, interface::Drawing::LINES_CONNECTED);
+    interface::Input::drawData(interface::Visual::PATHFINDING, path, Qt::blue, robotId, interface::Drawing::DOTS);
+
+    return command;
+}
+
+bool PositionControl::shouldRecalculatePath(const Vector2 &currentPosition, const Vector2 &targetPos, int robotId) {
+    return computedPaths[robotId].empty() ||
+            // distance between the new target and the former target
+            (targetPos - computedPaths[robotId].back()).length() > MAX_DEVIATION ||
```
rolfvdhulst: Although it might seem silly I would make this line (line 48) a function as it has a clear meaning.

src/control/positionControl/CollisionDetector.cpp
```diff
@@ -0,0 +1,47 @@
...
+std::vector<const Vector2 *> CollisionDetector::getRobotPositions(){
+    std::vector<const Vector2 *> robotPositions(robots.size());
+    std::transform(robots.begin(), robots.end(), robotPositions.begin(),
+                   [](const auto& robot) -> const Vector2 * { return &(robot.getPos()); });
+    return robotPositions;
+}
+
+void CollisionDetector::setField(world::Field *field) {
+    this->field = field;
+}
```
rolfvdhulst: Oof. You're leaving yourself open to a lot of UB here...

src/control/positionControl/CollisionDetector.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+bool CollisionDetector::isPointInDefenseArea(const Vector2 &point){
+    return field->pointIsInDefenceArea(point, true) || field->pointIsInDefenceArea(point, false);
+}
+
+bool CollisionDetector::isRobotCollisionBetweenPoints(const Vector2& initialPoint, const Vector2& nextPoint){
+    for (const auto& robot: robots){
+        // if the initial point is already close to a robot, then either 1. there is a collision, or 2. it is the original robot
+        if ((robot.getPos() - initialPoint).length() > Constants::ROBOT_RADIUS() && ControlUtils::distanceToLineWithEnds(robot.getPos(), initialPoint, nextPoint) < this->DEFAULT_ROBOT_COLLISION_RADIUS){
```
rolfvdhulst: Styling: pls seperate long if statements like this on multiple lines.

include/roboteam_ai/control/positionControl/pathPlanning/VoronoiPathPlanning.h
```diff
@@ -0,0 +1,81 @@
...
+    double distance;
+};
+
+//black magic vector hashing - taken from the boost hash combine
+struct hashPoint{
+    size_t operator()(const rtt::Vector2& point) const {
+        size_t xHash = std::hash<double>()(point.x) + 0x9e3779b9;
+        return xHash ^ (std::hash<double>()(point.y) + 0x9e3779b9 + (xHash<<6) + (xHash>>2));
+    }
+};
```
rolfvdhulst: How relevant is this for speed?
If not necessary I would avoid it.

include/roboteam_ai/control/positionControl/pathPlanning/Voronoi.h
```diff
@@ -0,0 +1,1607 @@
+// Copyright (c) 2015-2019 Mathias Westerdahl
```
rolfvdhulst, line 1: This file should not go with this PR but rather into roboteam_utils as it's a library. Or you actually link it as a header library. Is there a way to apt-get this?

include/roboteam_ai/control/positionControl/PositionControl.h
```diff
@@ -0,0 +1,71 @@
...
+namespace rtt::ai::control {
+
+/**
+ * The main position control class. Use this for your robot position control
+ * requirements.
+ */
+class PositionControl {
+private:
+
+    const double MAX_DEVIATION = 0.3;
```
rolfvdhulst: This si quite a lot of deviation. 0.3 meters is a lot.

include/roboteam_ai/control/positionControl/PositionControl.h
```diff
@@ -0,0 +1,71 @@
...
+
+    NumTreesPlanning *pathPlanningAlgorithm = nullptr;
+    rtt::ai::control::NumTreesTracking *pathTrackingAlgorithm = nullptr;
+    CollisionDetector *collisionDetector = nullptr;
```
rolfvdhulst: raw pointers :(

include/roboteam_ai/control/positionControl/pathPlanning/NumTreesPlanning.h
```diff
@@ -0,0 +1,68 @@
...
+#include "control/positionControl/CollisionDetector.h"
+
+namespace rtt::ai::control{
+
+/**
+ * Path planning algorithm. See method computePath for details.
+ */
+class NumTreesPlanning {
+private:
+    using PathPointer = std::shared_ptr<rtt::ai::control::PathPoint>;
```
rolfvdhulst: oh god. If there is any way you can make this a unique pointer or anything else...

rolfvdhulst (changes requested):
- There's a lot of redundant namespace qualifiers (e.g. rtt:: everywhere before Vector2)
- You need to add a ton more comments/documentation, especially to long functions or as to WHY you are doing something. I can see what your code does roughly in each function but it's really difficult to piece things together as to how it all fits together.
- std::vector is superior to std::list in 99.9% of the cases where you are working with small structs.
- Many functions are quite long. Try breaking up things into smaller subfunctions if you can, it really helps code readability a lot.
- I personally dislike the usage of reference variables as output. I know it's more standard in other places but it's not consistent with our style.
- Try to not use raw pointers. Unique pointers are almost always better and have much less chance of causing UB or problems down the line at the cost of virtually zero overhead.

include/roboteam_ai/control/positionControl/CollisionDetector.h
```diff
@@ -0,0 +1,60 @@
...
+private:
+    const double DEFAULT_ROBOT_COLLISION_RADIUS = 3.0*Constants::ROBOT_RADIUS();
+
+    const std::vector<rtt::world_new::robot::Robot> &robots;
```
Yuhanun: Should be views

include/roboteam_ai/control/positionControl/CollisionDetector.h
```diff
@@ -0,0 +1,60 @@
...
+    world::Field* field = nullptr;
+
+public:
+    /**
+     * Initialize the collision detector with the robot objects
+     * @param robots
+     */
+    explicit CollisionDetector(const std::vector<rtt::world_new::robot::Robot> &robots);
```
Yuhanun: Should be views

include/roboteam_ai/control/positionControl/CollisionDetector.h
Yuhanun: Why Vector2*'s?

include/roboteam_ai/control/positionControl/PositionControl.h
```diff
@@ -0,0 +1,71 @@
...
+#ifndef RTT_POSITIONCONTROL_H
+#define RTT_POSITIONCONTROL_H
+
+
+#include "control/RobotCommand.h"
+#include "control/positionControl/pathPlanning/NumTreesPlanning.h"
+#include "control/positionControl/pathTracking/BasicPathTracking.h"
+#include "CollisionDetector.h"
+#include <world_new/Robot.hpp>
+#include <control/positionControl/pathTracking/NumTreesTracking.h>
```
Yuhanun: inconsistent include style

include/roboteam_ai/control/positionControl/PositionControl.h
Yuhanun: constexpr static

include/roboteam_ai/control/positionControl/PositionControl.h
```diff
@@ -0,0 +1,71 @@
...
+class PositionControl {
+private:
+
+    const double MAX_DEVIATION = 0.3;
+
+    NumTreesPlanning *pathPlanningAlgorithm = nullptr;
+    rtt::ai::control::NumTreesTracking *pathTrackingAlgorithm = nullptr;
+    CollisionDetector *collisionDetector = nullptr;
+
+    const std::vector<rtt::world_new::robot::Robot> &robots;
```
Yuhanun: Should be views

include/roboteam_ai/control/positionControl/PositionControl.h
```diff
@@ -0,0 +1,71 @@
...
+
+    std::unordered_map<int, std::list<Vector2>> computedPaths;
+
+public:
+    /**
+     * Specify in the constructor the algorithms (path tracking and planning) in
+     * case another one is needed.
+     * @param robots
+     */
+    explicit PositionControl(const std::vector<rtt::world_new::robot::Robot> &robots);
```
Yuhanun: Should be views

include/roboteam_ai/control/positionControl/pathPlanning/NumTreesPlanning.h
Yuhanun: Why does this even have to be apointer

include/roboteam_ai/control/positionControl/pathPlanning/VoronoiPathPlanning.h
Yuhanun: Vector2 naturally isn't hashable so you kind of have to hash it to be able to store it in a hashmap

include/roboteam_ai/control/positionControl/pathTracking/BasicPathTracking.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <utilities/Constants.h>
+#include <roboteam_utils/Vector2.h>
+
+/**
+ * Path tracking algorithm. See method computePath for details.
+ */
+class BasicPathTracking {
+private:
+    double maxVelocity = rtt::ai::Constants::MAX_VEL();
```
Yuhanun: constexpr's

include/roboteam_ai/control/positionControl/pathTracking/NumTreesTracking.h
```diff
@@ -0,0 +1,42 @@
...
+     * NumTreesTracking looks at the target position after N steps and follows that
+     * using a PID controller (the PidTracking).
+     * @param currentPosition
+     * @param currentVelocity
+     * @param pathPoints the path as a list of points
+     * @param outputVelocity control velocity that will be outputted to the robot at the current tick
+     * @param outputAngle the resulting orientation of the robot at the current tick
+     */
+    void trackPath(const rtt::Vector2 &currentPosition, const rtt::Vector2 &currentVelocity,
+                   std::list<rtt::Vector2> &pathPoints, rtt::Vector2 &outputVelocity, double &outputAngle);
```
Yuhanun: std::list has a VERY specific use case, consider using a vector (or even a queue, not sure what you do with them)

include/roboteam_ai/control/positionControl/pathTracking/PidTracking.h
```diff
@@ -0,0 +1,52 @@
...
+#include <roboteam_utils/pid.h>
+
+namespace rtt::ai::control{
+
+/**
+ * Path tracking algorithm. See method computePath for details.
+ */
+class PidTracking {
+private:
+    double maxVelocity = rtt::ai::Constants::MAX_VEL();
```
Yuhanun: constexpr

include/roboteam_ai/world_new/World.hpp
Yuhanun: Apparently this is not a controller for a single robot, it's for all, meaning there should be only 1 instance

src/control/positionControl/PositionControl.cpp
```diff
@@ -0,0 +1,53 @@
...
+//
+
+#include "control/positionControl/PositionControl.h"
+#include <interface/api/Input.h>
+#include <control/positionControl/pathTracking/NumTreesTracking.h>
+
+namespace rtt::ai::control {
+
+PositionControl::PositionControl(const std::vector<rtt::world_new::robot::Robot> &robots): robots(robots) {
+    collisionDetector = new CollisionDetector(robots);
```
Yuhanun: oh no

src/control/positionControl/PositionControl.cpp
```diff
@@ -0,0 +1,53 @@
...
+    command.pos = computedPaths[robotId].front();
+    pathTrackingAlgorithm->trackPath(
+            currentPosition,
+            currentVelocity,
+            computedPaths[robotId],
+            command.vel,
+            angle);
+    command.angle = angle;
+
+    std::vector<Vector2> path(computedPaths[robotId].begin(),computedPaths[robotId].end());
```
Yuhanun: The fact that you have to copy this over for the interace maybe means you should use a vector in the first place

src/control/positionControl/PositionControl.cpp
```diff
@@ -0,0 +1,53 @@
...
+PositionControl::computeAndTrackPath(world::Field *field, int robotId, const Vector2 &currentPosition,
+                                     const Vector2 &currentVelocity, const Vector2 &targetPosition) {
+    //TODO: this is a workaround caused by the fact that the field is not global
+    collisionDetector->setField(field);
+    if (shouldRecalculatePath(currentPosition, targetPosition, robotId)) {
+        computedPaths[robotId] = pathPlanningAlgorithm->computePath(currentPosition, targetPosition);
+    }
+
+    RobotCommand command = RobotCommand();
+    double angle;
```
Yuhanun: initialize this to 0

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
```diff
@@ -0,0 +1,214 @@
...
+    root->currentTarget = targetPosition;
+    root->pos = currentPosition;
+    root->vel = Vector2();
+    root->acc = Vector2();
+    root->t = 0;
+    root->collisions = 0;
+    pathQueue.push(root);
+
+    auto start = std::chrono::duration_cast< std::chrono::milliseconds >(
+            std::chrono::system_clock::now().time_since_epoch()
```
Yuhanun: Dont' use system_clock, use the tick from roboteam_ai world_new.

If that's not accurate enough then maybe we should consider switching those ticks to ms or ns as im p sure it's seconds rn

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
```diff
@@ -0,0 +1,214 @@
...
+    PathPointer root = std::make_shared<rtt::ai::control::PathPoint>();
+    root->currentTarget = targetPosition;
+    root->pos = currentPosition;
+    root->vel = Vector2();
+    root->acc = Vector2();
+    root->t = 0;
+    root->collisions = 0;
+    pathQueue.push(root);
+
+    auto start = std::chrono::duration_cast< std::chrono::milliseconds >(
```
Yuhanun: Why the cast< T > spaces

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+    // backtrack to 0.75 seconds before the collision
+    double timeBeforeCollision = 0.75;
+    PathPointer startPoint = collisionPoint->backTrack(collisionPoint->t - timeBeforeCollision);
+
+    // determine the position at backtrack and vector towards the collision
+    //TODO: workaround warning: used half the distance instead of collision distance
+    Vector2 collisionPosition = (collisionPoint->parent->pos + collisionPoint->pos)/2;
+    Vector2 deltaPosition = collisionPosition - collisionPoint->pos;
+    double collisionRadius = DEFAULT_ROBOT_COLLISION_RADIUS;
```
Yuhanun: that's already a constexpr, if you don't assign it itll be most likely inlined

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+void VoronoiPathPlanning::completeGraphWithOriginDestination(const rtt::Vector2 &robotPosition, const rtt::Vector2 &targetPosition) {
+    //1 or fewer obstacles in the area - numsites == 1 or 2
+    if (voronoiDiagram.numsites <= 2){
+        double distance = (robotPosition - targetPosition).length();
+        // no obstacle or the distance to the obstacle safe
+        if (voronoiDiagram.numsites < 2 || rtt::ai::control::ControlUtils::distanceToLine(
+                convertFromJcvPoint(voronoiDiagram.internal->sites->p),robotPosition,targetPosition) > 2 * rtt::ai::Constants::ROBOT_RADIUS()) {
+            graphAdjacencyList.insert({robotPosition, std::list<GraphNode>(1, (GraphNode) {targetPosition, distance})});
```
Yuhanun: why call it a "graphAdjecencyList" if it's a map...

graphAdjecencyList[robotPosition] = ...; please

src/control/positionControl/pathTracking/BasicPathTracking.cpp
Yuhanun: Not guaranteed to be < 64 bytes, but yeah u rite

src/skills/DriveWithInterface.cpp
Yuhanun: He's initializing this numTreeGtp to nullptr, why not initialize it to a unique_ptr which actually contains the controller rather than being a null

src/skills/DriveWithInterface.cpp
```diff
@@ -5,23 +5,26 @@
...
     Vector2 targetPos = interface::Output::getInterfaceMarkerPosition();
-
-    auto robotCommand = numTreeGtp.getRobotCommand(world, field, robot, targetPos);
+    auto robotCommand = numTreeGtp->computeAndTrackPath(field, robot->id, robot->pos,
+                                                        robot->vel, targetPos);
 
     command.mutable_vel()->set_x(robotCommand.vel.x);
```
Yuhanun, line 21: This is exactly why we need a RobotPacketController for each robot

getPacketController(id)->positionPacket(vel.x, vel.y, angle)->publish();

src/world_new/World.cpp
```diff
@@ -86,4 +86,12 @@ uint64_t World::getTimeDifference() const noexcept {
 robot::RobotControllers &World::getControllersForRobot(uint8_t id) noexcept {
     return robotControllers[id];
 }
+
+ai::control::PositionControl* World::getRobotPositionController() noexcept {
+    if (positionControl == nullptr){
+        positionControl = new ai::control::PositionControl(this->currentWorld->getRobots());
```
Yuhanun: Memory leak.

src/control/positionControl/PositionControl.cpp
Yuhanun: Memory leak aswel btw

src/world_new/World.cpp
Yuhanun: unique_ptr pls

src/control/positionControl/PositionControl.cpp
Yuhanun: unique_ptr pls

Yuhanun (changes requested):

Yuhanun:
Did a quick scan, not sure about functionality but you've got some code flaws that can cause issues, nobody is perfect, dw <3

Looks good overall, though

include/roboteam_ai/world_new/World.hpp
rolfvdhulst: Still it shouldn't live in the same data space: We may want to make a more general higher level 'context' struct which contains pointers to all globals or so, but if we make the distinction between world and field it makes very little sense for controllers to exist in world.

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
rolfvdhulst: This is to measure calculation time, not world time. This was already in the original code.

rolfvdhulst: So this is perferctly fine.

include/roboteam_ai/control/positionControl/CollisionDetector.h
Yuhanun: You can't REALLY use references in std::vector, gets very hard to work with

include/roboteam_ai/control/positionControl/pathPlanning/VoronoiPathPlanning.h
rolfvdhulst: We might want to add this to roboteam_utils then for some types. 
It's nice if it's defined

src/skills/DriveWithInterface.cpp
Ratoone: This was a workaround for testing purposes until the next PR, where I want to completely replace the old position control from all the skills

include/roboteam_ai/world_new/World.hpp
Ratoone: I know, but right now there was no other place where this could be placed

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
Ratoone: Indeed, but in the end we may not even use Voronoi. If someone decides this is needed, it can be optimized then.

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
Ratoone: I didn't change anything from the initial mess. I could do some cleanup though

include/roboteam_ai/control/positionControl/pathPlanning/Voronoi.h
Ratoone: It's a header library. No apt-get, but I attached the github link

Yuhanun: This can be put in the docker image.

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
Ratoone: Graph theory. https://en.wikipedia.org/wiki/Adjacency_list

include/roboteam_ai/control/positionControl/pathPlanning/Voronoi.h
Yuhanun: Done, just `#include <jc_voronoi.h>` now

include/roboteam_ai/control/positionControl/PositionControl.h
Ratoone: 0.3m was the deviation in the old algorithm. I can decrease it to 0.2m

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
Ratoone: I wanted to round it to 3 decimals for testing purposes. It's not like anything smaller than 1mm is relevant

include/roboteam_ai/control/positionControl/CollisionDetector.h
Ratoone: Then, question: why does WorldData::getRobots() return a vector of Robots, instead of RobotViews? There isn't any other way to get all the robots, at least from what I've seen

Yuhanun: getRobotsNonOwning()

getRobots() is an O(1) operationg, getRobotsNonOwning is O(n)

include/roboteam_ai/control/positionControl/PositionControl.h
rolfvdhulst: No keep it like that then but move it to numtrees files I would say. It does not make a lot of sense for it to live here.

src/world_new/World.cpp
Ratoone: If the getRobots is returning a reference to the vector of robots (which was), then you don't need to poll for changes. I'll have to change it though, as Johan requested to change to RobotViews, so I may have to do polling for now

include/roboteam_ai/control/positionControl/PositionControl.h
Ratoone: It is used though in the condition if the path should be recalculated - which does not depend on the planning method. Maybe I can rename it to show that

include/roboteam_ai/control/positionControl/pathPlanning/NumTreesPlanning.h
Ratoone: I will open an issue with this and fix it after removing the whole NumTrees thing. Right now it's so deep entrenched in the old algorithm that this refactoring is not worth at this time

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
Ratoone: More cleanup scheduled with #777

include/roboteam_ai/control/positionControl/CollisionDetector.h
Ratoone: I'll open an issue with this and implement it after fixing that part of the world

include/roboteam_ai/control/positionControl/CollisionDetector.h
Ratoone: Opened an issue with this

include/roboteam_ai/control/positionControl/CollisionDetector.h
```diff
@@ -0,0 +1,61 @@
...
+    bool isRobotCollisionBetweenPoints(const Vector2& initialPoint, const Vector2& nextPoint);
+
+    bool isPointInsideField(const Vector2 &point);
+
+    bool isPointInDefenseArea(const Vector2 &point);
+
+    std::vector<Vector2> getRobotPositions();
+
+    void setField(world::Field &field);
+};
```
Yuhanun: docs for the first two

include/roboteam_ai/control/positionControl/PositionControl.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * The main position control class. Use this for your robot position control
+ * requirements.
+ */
+class PositionControl {
+private:
+    std::unique_ptr<NumTreesPlanning> pathPlanningAlgorithm;
+    std::unique_ptr<NumTreesTracking> pathTrackingAlgorithm;
+    std::unique_ptr<CollisionDetector> collisionDetector;
```
Yuhanun, line 25: why pointers

include/roboteam_ai/control/positionControl/PositionControlUtils.h
```diff
@@ -0,0 +1,48 @@
+//
+// Created by ratoone on 30-01-20.
+//
+
+#ifndef RTT_POSITIONCONTROLUTILS_H
+#define RTT_POSITIONCONTROLUTILS_H
+
+#include "roboteam_utils/Vector2.h"
+#include <list>
```
Yuhanun: redundant include

include/roboteam_ai/control/positionControl/pathPlanning/NumTreesPlanning.h
```diff
@@ -0,0 +1,66 @@
...
+
+    double remainingStraightLinePathLength(const Vector2 &currentPos, const Vector2 &halfwayPos, const Vector2 &finalPos);
+
+    std::vector<Vector2> backTrackPath(PathPointer point);
+
+    std::pair<std::vector<Vector2>, PathPointer> getNewTargets(const PathPointer &collisionPoint);
+
+    std::vector<Vector2> tracePath(const Vector2 &currentPosition, const Vector2 &targetPosition);
+
+    PathPointer computeNewPoint(const Vector2 &targetPosition, const PathPointer &oldPoint, const Vector2 &subTarget);
```
Yuhanun: dpcs

Yuhanun: docs*

include/roboteam_ai/control/positionControl/pathPlanning/VoronoiPathPlanning.h
```diff
@@ -0,0 +1,66 @@
...
+
+    Vector2 convertFromJcvPoint(jcv_point point);
+
+    void generateGraphFromDiagram();
+
+    std::vector<Vector2> generatePathDijkstra(const Vector2& initialPosition, const Vector2& targetPosition);
+
+    void computeDiagram(const Vector2 &robotPosition, const Vector2 &targetPosition);
+
+    void addPointToGraph(const Vector2 &pointToAdd);
```
Yuhanun: docs

include/roboteam_ai/control/positionControl/pathTracking/PidTracking.h
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Path tracking algorithm. See method computePath for details.
+ */
+class PidTracking {
+private:
+    static constexpr double MAX_VELOCITY = Constants::MAX_VEL();
+
+    // the PID controllers on the two axes
+    PID xPid = PID(0,0,0,0);
+    PID yPid = PID(0,0,0,0);
```
Yuhanun, line 26: confident this is just default construction but it's ok :)

include/roboteam_ai/control/positionControl/pathTracking/PidTracking.h
```diff
@@ -0,0 +1,51 @@
...
+
+    void updatePidValues();
```
Yuhanun: update to what? docs.

src/control/positionControl/PositionControl.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "control/positionControl/PositionControlUtils.h"
+#include "control/positionControl/pathTracking/NumTreesTracking.h"
+#include "interface/api/Input.h"
+
+namespace rtt::ai::control {
+
+PositionControl::PositionControl(const std::vector<world_new::robot::Robot> &robots) : robots(robots) {
+    collisionDetector = std::make_unique<CollisionDetector>(robots);
+    pathPlanningAlgorithm = std::make_unique<NumTreesPlanning>(*collisionDetector);
+    pathTrackingAlgorithm = std::make_unique<NumTreesTracking>();
```
Yuhanun, line 15: initializer list

src/control/positionControl/PositionControl.cpp
```diff
@@ -0,0 +1,42 @@
...
+// TODO: add projection to outside defence area (project target position)(is this really needed?)
+RobotCommand PositionControl::computeAndTrackPath(world::Field &field, int robotId, const Vector2 &currentPosition, const Vector2 &currentVelocity, const Vector2 &targetPosition) {
+    // TODO: this is a workaround caused by the fact that the field is not global
+    collisionDetector->setField(field);
+    if (shouldRecalculatePath(currentPosition, targetPosition, robotId)) {
+        computedPaths[robotId] = pathPlanningAlgorithm->computePath(currentPosition, targetPosition);
+    }
+
+    interface::Input::drawData(interface::Visual::PATHFINDING, computedPaths[robotId], Qt::green, robotId, interface::Drawing::LINES_CONNECTED);
+    interface::Input::drawData(interface::Visual::PATHFINDING, computedPaths[robotId], Qt::blue, robotId, interface::Drawing::DOTS);
```
Yuhanun, line 27: Why are we drawing here? Decouple interface time (bachelor assignments right?)

src/control/positionControl/PositionControlUtils.cpp
```diff
@@ -0,0 +1,20 @@
...
+namespace rtt::ai::control {
+
+bool PositionControlUtils::isTargetChanged(const Vector2 &targetPos, const Vector2 &oldTarget) { return (targetPos - oldTarget).length() > MAX_TARGET_DEVIATION; }
+
+bool PositionControlUtils::isTargetReached(const Vector2 &targetPos, const Vector2 &currentPosition) {
+    return (targetPos - currentPosition).length() < MIN_DISTANCE_TARGET_REACHED;
+}
+
+void PositionControlUtils::removeFirstIfReached(std::vector<Vector2> &path, const Vector2 &currentPosition) {
+    if (path.size() > 1 && isTargetReached(path.front(), currentPosition)) {
```
Yuhanun, line 16: `>= 1`, right?

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include "control/positionControl/pathPlanning/NumTreesPlanning.h"
+
+namespace rtt::ai::control {
+
+NumTreesPlanning::NumTreesPlanning(CollisionDetector &collisionDetector) : collisionDetector(collisionDetector) {}
+
+std::vector<Vector2> NumTreesPlanning::computePath(const Vector2 &robotPosition, const Vector2 &targetPosition) {
+    DT = 0.3 / GameStateManager::getCurrentGameState().getRuleSet().maxRobotVel;
+    if (DT > 0.12) DT = 0.12;
+    if (DT < 0.06) DT = 0.06;
```
Yuhanun:
```cpp
DT = std::clamp(DT, 0.06, 0.12) // maybe a 4th paramter that's a proper double comparison
```

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    auto path = tracePath(robotPosition, targetPosition);
+    double goToTimeInFuture = 0.4;
+    auto targetPathPoint = static_cast<unsigned long>(goToTimeInFuture / DT);
```
Yuhanun, line 17: confusing variable name

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
```diff
@@ -0,0 +1,200 @@
...
+    while (point) {
+        backTrackedPath.push_back(point->pos);
+        if (point->parent == nullptr) {
+            break;
+        }
+        point.swap(point->parent);
+    }
+
+    // as older points get pushed back, the vector has to be reversed
+    std::reverse(backTrackedPath.begin(), backTrackedPath.end());
```
Yuhanun, line 138: Can we not prevent this?

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,131 @@
...
+        graphAdjacencyList[pointToAdd] = std::list<GraphNode>(1, (GraphNode){closestPoint, minDist});
+        graphAdjacencyList[closestPoint].push_back({pointToAdd, minDist});
+    }
+}
+
+std::vector<Vector2> VoronoiPathPlanning::generatePathDijkstra(const Vector2& initialPosition, const Vector2& targetPosition) {
+    std::unordered_map<Vector2, float> distanceVector;
+    std::unordered_map<Vector2, Vector2> parentVector;
+    distanceVector.insert({initialPosition, 0});
+    std::list<Vector2> nodeQueue;
```
Yuhanun, line 82: nodeQueue, in an std::list?...

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
```diff
@@ -0,0 +1,131 @@
...
+    if (graphAdjacencyList.find(pointToAdd) == graphAdjacencyList.end()) {
+        graphAdjacencyList[pointToAdd] = std::list<GraphNode>(1, (GraphNode){closestPoint, minDist});
+        graphAdjacencyList[closestPoint].push_back({pointToAdd, minDist});
+    }
+}
+
+std::vector<Vector2> VoronoiPathPlanning::generatePathDijkstra(const Vector2& initialPosition, const Vector2& targetPosition) {
+    std::unordered_map<Vector2, float> distanceVector;
+    std::unordered_map<Vector2, Vector2> parentVector;
+    distanceVector.insert({initialPosition, 0});
```
Yuhanun: distanceVector[initialPosition] = 0;

src/control/positionControl/pathTracking/BasicPathTracking.cpp
```diff
@@ -0,0 +1,20 @@
...
+Position BasicPathTracking::trackPath(const Vector2 &currentPosition, const Vector2 &currentVelocity, std::vector<Vector2> &pathPoints) {
+    PositionControlUtils::removeFirstIfReached(pathPoints, currentPosition);
+
+    Vector2 velocity = pathPoints.front() - currentPosition;
+
+    if (velocity.length() > MAX_VELOCITY) {
+        velocity = velocity.stretchToLength(MAX_VELOCITY);
+    }
+
+    return Position(velocity, velocity.angle());
```
Yuhanun: Position{ velocity, velocity.angle() };

src/world_new/World.cpp
```diff
@@ -80,5 +80,12 @@ uint64_t World::getTimeDifference() const noexcept { return tickDuration; }
 
 robot::RobotControllers &World::getControllersForRobot(uint8_t id) noexcept { return robotControllers[id]; }
 
+ai::control::PositionControl* World::getRobotPositionController() noexcept {
+    if (positionControl == nullptr){
```
Yuhanun: if (!positionControl)

Yuhanun (changes requested):

src/control/positionControl/PositionControlUtils.cpp
Ratoone: No, because you don't want to pop the last element of the path

include/roboteam_ai/control/positionControl/pathTracking/PidTracking.h
Ratoone: There is no default constructor for the PID class

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
Ratoone: The alternative would be to insert at the front of the vector, which is not really efficient

src/control/positionControl/PositionControl.cpp
Ratoone: I'm open to suggestions for a better way to do this

src/control/positionControl/pathPlanning/VoronoiPathPlanning.cpp
Ratoone: The algorithm at its base uses a queue, but I had some problems implementing it as one, so I made it a list. The elements get traversed only once, so I wouldn't name it a list

PR #767, opened 2020-01-28
Create the 'Dealer' class as a potential replacement for RobotDealer

### LINKED ISSUE #623: Smarter robot dealing
Robotdealer could be smarter, now all robots are claimed in order which is never optimal. Besides that, there are sometimes multiple features which a robot should have. 
lets see the following claim list in switches to illustrate. I have written some notes for a proposed fix below. For now this is quickly written down for myself and i will improve this later on :) 

role1: working geneva and close to our goal
role2: working geneva and close to our goal
role3: working geneva and working ballsensor and close to their goal
role4: working geneva and working ball sensor and close to their goal

Problem statement:
Imagine the following set of robots in world (and ignore the distances for now)

robot1: working geneva and ballsensor
robot2: working geneva and ballsensor 
robot3: working geneva
robot4: nothing works

Currently robot 1 is claimed for role1
robot2 is claimed for role2
robot3 is claimed for role3
robot4 is claimed for role4 

this gives a lot of sadness. Robot1 and robot2 have a working ballsensor but its not used, and robot3 needs one but it is not used.

With distances taken into account this can get even worse. What if you first claim ROBOT_CLOSEST_TO_GOAL and its geneva works, and later on you cannot even claim a robot with a working geneva anymore?

----- 
#proposal (train of thought) 

This looks like a linear assignment problem so lets try to give it a go with hungarian.

distToTargetScore // m
distanceFactor // amount of failures

all preferences the same, distanceFactor = 0
amount of differences: distance = amount of failures

// the problem now is the distToTargetScore. 
// how to take that into account?
we know functionalityScore is way more important so always stick with that. but if there are robots with an equal cost ...

a robot with working gneeva and ballsensor, but is entirely on the wrong side of the field should still be claimed at the front then.... 

// idea: alwyas increase the distance of non-working robots with (field_diagonal lenght * error) so the proper robots are always having the shortest path, and between al the working robots their distances are still compared... 
// this should work (tm)
// then use hungarian only once

mrlukasbos:
> Note: this is still WIP. Because this is a score-based approach on robotdealer I would like to hear some ideas on how to implement this. (read below for more details) 

needs https://github.com/RoboTeamTwente/roboteam_ai/pull/775

### todo
- [x] implement a decent score/decent score system
- [x] create more dealer flags
- [x] distribution tests, in/out tests
- [x] priority system
- [x] inputs/outputs
- [x] apply hungarian

### Overview
Hereby I present Dealer. An idea I have for a very long time, which solves the problem that robots cannot be properly claimed by robotdealer. For example, when taking a free kick we want to use a robot that has a working geneva and a working ballsensor, and if possible we also want one that is close to the ball.

with Robotdealer that gives problems, as we can only sort on robots with working genava and ballsensor, but not on distance to ball then. Or we can sort on 'distance to ball' but not on geneva and ballsensor. Besides, if all robots with a geneva and ballsensor are already claimed a random robot is picked! 

With dealer I try to resolve these issues. For a role now multiple flags can be given instead of one. For example in switches you would get:  
```
{attacker_1, {WORKING_GENEVA, WORKING_BALLSENSOR, CLOSE_TO_BALL}
, ...}
```

For every robot all the given flags will be evaluated and put into a matrix. The matrix maps the role names to robot ids and gives each column a score based on all these flags. 

```
  ---------------------------------
          robot_1 robot_2 robot_3
  role_1     12       11     0
  role_2     3        3      1
  role_3     22       1      2
  ---------------------------------
``` 

Then this matter suddenly became the Linear Assignment problem, which we can solve with the Hungarian algorithm. This gives us the a of robots for each role. 

### Score based approach
The problem with a score based approach is that a 'score' is essentially meaningless and thus hard to quantify. In this approach scores (such as working geneva) need to be balanced out with distances. If we really need a geneva then it is worth it to let the robot drive over half the field, for example to take a penalty. However, if we might use a geneva but speed is more important, we might want to take a robot that is closer. So it looks like the score will have to be given as a distance. @rolfvdhulst and/or @RobotJesse do you guys have an opinion on this? Personally I think working with a distance metric is quite okay, especially because I will add some tweaking parameters such that you can still tweak this externally, for example:

```
{attacker_1, {
{WORKING_GENEVA, HIGH_PRIORITY},
{WORKING_BALLSENSOR, HIGH_PRIORITY, 
{CLOSE_TO_BALL, LOW_PRIORITY}
, ...}

```
The priority statements would influence the score measured in distance. for example:
HIGH_PRIORITY = 20 meter
MEDIUM_PRIORITY = 4 meter
LOW_PRIORITY = 2 meter

fixes #623


rolfvdhulst:
Without looking at the code and just at the idea: I think it's definitely a good step in the right direction. It is more dynamic and probably easier to work with. However, I do think we should look at what exactly the type of decision we want RobotDealer to make is.  

Hungarian tries to minimize the total 'cost' where cost is basically any metric we define.  This can feel 'meaningless' but certainly does not need to be. We will need to at some point make the decision somewhere whether or not having a ball sensor is worth the extra distance, so this difficult aspect will always need to be incorporated in some part of our code. In that sense I would not think of the score as a distance but rather as a 'cost'. 
I can also imagine we will want more time-based scores in the feature, especially regarding e.g. ball interception and offensively based plays. Then we might _not_ want to use hungarian, but rather e.g. an algorithm that minimizes the total time before all robots are at their place rather than the sum of the travel distances. (Look at e.g. linear bottleneck assignment problem. This stuff is quite close to some of the things I study in my master). For small instances like ours (11 robots), we should be able to solve (or approximate) many variants of assignment problems (probably even if they are NP-hard) efficiently enough to do more complicated things if we want to.

Only thing I wonder though is; what is RobotDealer doing _exactly_? What is it's input/output? Because I personally still don't really like the way it interfaces to the trees. It feels messy and there's many things that are quite hacky (e.g. the ShouldHandleBall condition). 

I personally want to go to a system where a RobotDealer receives e.g. a set of 'tasks' all with a priority (e.g. tactics?) Each task needs a certain amount of robots. Then robotdealer goes through the tasks from high to low priority and assigns robots to the task (I think having some hysteresis as @RobotJesse mentioned in his PR for trees would be helpful here: don't continuously reassign robots (e.g. lock them to a tactic/task for a while, but allow to release/swap them with other robots after a certain period of time or if they are needed for higher priority tasks. )

Still tactics' communication with robotdealer should be a bit better somehow though, as a tactic typically has some core ideas (e.g. keeper position, defense positions, important attack positions, minimal interception) that should affect what robot to choose for what task. Whether or not these tasks are defined in some 'analysis' module or decided by the tactics, I don't know. 

In a new system we could also perhaps choose these tasks through another way. The problem currently is more existing from the fact that we never have structured this global decision making because everything is made from decision trees. 

Also: how 'slow' is Hungarian? I never tested its running time but I'm wondering how long it takes to compute for e.g. 11 robots.

mrlukasbos:
My PR completely ignores the concept of 'claiming' and 'freeing'. The inputs are very simple (I have them literally defined):

**input**: pairs of (roleName -> dealerFlags) and vector of robots
**output**: pairs of (roleName -> robot).

Thus I just create a distribution of robots for the given flags/roles. You only need to change this distribution when a tree changes. Besides, this dealer does not take all robots from world. Since it takes a vector of robots as input you can 'claim' robots by not giving them to robotdealer anyway. It removes a lot of the magic and gives a lot of control when creating Plays etc. 

Because of this you don't have to call Hungarian that often. Only when tree switches or tactic switches occur.

rolfvdhulst:
> Thus I just create a distribution of robots for the given flags/roles. You only need to change this distribution when a tree changes. Besides, this dealer does not take all robots from world. Since it takes a vector of robots as input you can 'claim' robots by not giving them to robotdealer anyway. It removes a lot of the magic and gives a lot of control when creating Plays etc.

Although I do think this is a great idea and enough for this PR, I was just trying to give ideas/options of where we might go next :) It does indeed remove a lot of magic.

RobotJente:
I like that you are structuring this :) I think flags are definitely a good option, at least to start with. 

As @rolfvdhulst mentioned, the costs don't have to be arbitrary and I do think assigning some score/cost to having robot i perform task j is a good way to solve the problem. 

If I understand what you are doing correctly, then you are assigning robots eligibility based on the flags, right? So you are pruning based on how high the priority is. If we have a high priority then we look within 20 meters traveling distance, if it's low priority then we only look if there's a robot within 4 meters that has these properties.

I personally would prefer a score based approach, because the flags (if I understood them correctly) make the decision making more discrete. If we have a high priority task and there's a robot at 21 meters, we can't use it anymore. I think scoring them based on some function like cost{i,j} = Distance(i,j) + Proficiency(i,j), given some robot with id i and some task j, so you get a bit more flexibility in your costs. However, this is also more difficult to tune.

About the more complex version of the Hungarian/the linear bottleneck problem: I actually thought you could generalize the Hungarian since it finds the optimal assignments if you have a complete bipartite graph with some edge weights. So it feels like Hungarian can also solve the Linear Bottleneck problem, as long as you define your edge weights properly. For example, defining the edge between robot i and task j to be e_{i,j} = Distance(i,j) + Proficiency(i,j) where the proficiency is some measure of how good the robot is at task j. This way you are modifying the cost function but I'm pretty sure you can still solve it with the Hungarian algorithm. I could be wrong though.

I agree with @rolfvdhulst that this is a step in the right direction. I also see the use of flags for pruning certain robots out of the search space if we are sure they are not worth looking at. I think this idea is definitely a lot better than what we have currently and can be built on.

Edit: from looking at your code, I see it's a score based approach, where the flags provide some multiplier, which I think is quite nice.

CMakeLists.txt
```diff
@@ -22,12 +22,19 @@ if (APPLE)
...
+            /usr/local/lib/libgmock_main.a
+            )
+else(NOT APPLE)
     SET(Qt5Widgets_DIR "/usr/include/x86_64-linux-gnu/qt5/QtWidgets")
     SET(Qt5Charts_DIR "/usr/include/x86_64-linux-gnu/qt5/QtCharts")
-    set(GTEST_LIB PRIVATE gtest)
-endif ()
+    set(GTEST_LIB
+            PRIVATE gtest
+            PRIVATE gmock)
```
rolfvdhulst, line 62: How are these changes relevant to this PR? I don't really see the point of what you are doing here.

include/roboteam_ai/world_new/Robot.hpp
```diff
@@ -113,8 +113,13 @@ class Robot {
 
     [[nodiscard]] bool isBatteryLow() const noexcept;
 
+    [[nodiscard]] bool isFiftyWatt() const noexcept;
+
+    [[nodiscard]] bool isThirtyWatt() const noexcept;
+
     [[nodiscard]] unsigned char getDribblerState() const noexcept;
 
+
```
rolfvdhulst, line 122: unnecessary blank line

src/utilities/Dealer.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+Dealer::Dealer(v::WorldDataView world, world::Field * field)
+    : world(world), field(field) { }
+
+Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priority)
+    : title(title), priority(priority) {}
+
+// Create a distribution of robots according to their flags
+std::unordered_map<std::string, v::RobotView> Dealer::distribute(const std::vector<v::RobotView>& allRobots, const FlagMap &flagMap) {
+    std::vector<std::vector<double>> scores = getScoreMatrix(allRobots, flagMap);
```
rolfvdhulst, line 15: Not sure about runtime performance/optimization relevance, but since the amount of robotviews/robots is known this could be a static matrix perhaps. Not a huge issue for now though.

src/utilities/Dealer.cpp
```diff
@@ -0,0 +1,122 @@
...
+        }
+        case DealerFlagTitle::WITH_WORKING_BALL_SENSOR: return costForProperty(robot->isWorkingBallSensor());
+        case DealerFlagTitle::ROBOT_TYPE_50W: return costForProperty(robot->isFiftyWatt());
+        case DealerFlagTitle::ROBOT_TYPE_30W: return costForProperty(robot->isThirtyWatt());
+        case DealerFlagTitle::WITH_WORKING_DRIBBLER: return costForProperty(robot->isWorkingDribbler());
+        case DealerFlagTitle::READY_TO_INTERCEPT_GOAL_SHOT: {
+            // get distance to line between ball and goal
+            // TODO this method can be improved by choosing a better line for the interception.
+            LineSegment lineSegment = {world.getBall()->get()->getPos(), field->get_field().get(OUR_GOAL_CENTER)};
+            return control::ControlUtils::distanceToLine(robot->getPos(), lineSegment.start, lineSegment.end);
```
rolfvdhulst: return linesegment.distanceToLine(robot->getPos()) is better here. Try to avoid using controlUtils if possible.

rolfvdhulst: You can then also remove the include.

rolfvdhulst: We will probably want a call to some calculator that does line interception and trajectory planning magic in the  future though.

src/utilities/Dealer.cpp
```diff
@@ -0,0 +1,122 @@
...
+        case DealerFlagTitle::CLOSE_TO_THEIR_GOAL: 
+            return costForDistance(field->getDistanceToGoal(false, robot->getPos()), fieldWidth, fieldLength);
+        case DealerFlagTitle::CLOSE_TO_OUR_GOAL: 
+            return costForDistance(field->getDistanceToGoal(true, robot->getPos()), fieldWidth, fieldLength);
+        case DealerFlagTitle::CLOSE_TO_BALL: {
+            auto ball = world.getBall();
+            if (!ball) return 0.0;
+            return robot->getPos().dist(ball.value()->getPos());
+        }
+        case DealerFlagTitle::WITH_WORKING_BALL_SENSOR: return costForProperty(robot->isWorkingBallSensor());
```
rolfvdhulst, line 98: Make a TODO/issue that all of these properties still would need to be tuned.

test/helpers/WorldHelper.cpp
```diff
@@ -180,24 +171,28 @@ proto::World WorldHelper::getWorldMsg(int amountUs, int amountThem, bool withBal
         msg.mutable_yellow()->CopyFrom(randomYellow);
         msg.mutable_blue()->CopyFrom(randomBlue);
 
-        if (withBall) msg.set_allocated_ball(&randomBall);
-    } while (!allPositionsAreValid(msg, true));
+        if (withBall) {
+            std::cerr << "[WorldHelper] Caution: generating a world with a ball is not stable!" << std::endl;
```
rolfvdhulst, line 169: I am confused by this line. What do you mean, not stable?

test/UtilTests/DealerTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+ * The getDefaultFlagScores() function can be overridden,
+ * Since it brings in a lot of uncertainty from world and field.
+ */
+class MockDealer : public Dealer {
+ public:
+  explicit MockDealer(v::WorldDataView world) : Dealer(world, nullptr) {}
+  double getDefaultFlagScores(const RobotView &robot, const Dealer::DealerFlag &flag) override {
+      return robot->getId() + 1;
+  }
+  // MOCK_METHOD(double, getDefaultFlagScores, (const RobotView &robot, const Dealer::DealerFlag &flag), (override));
```
rolfvdhulst: ?

src/world_new/World.cpp
```diff
@@ -80,5 +80,12 @@ uint64_t World::getTimeDifference() const noexcept { return tickDuration; }
...
 robot::RobotControllers &World::getControllersForRobot(uint8_t id) noexcept { return robotControllers[id]; }
 
-size_t World::getHistorySize() const noexcept { return history.size(); }
+size_t World::getHistorySize() const noexcept {
+    return history.size();
+}
+
+Settings World::getSettings() const noexcept {
+    return Settings(*settings);
+}
```
rolfvdhulst: While I get that you need it for tests, this is one of those ugly things that will propagate through our system with people using it everywhere and then we have no seperation of concerns anymore. Get it at the source, please no more hookups to World which are unnecessary (e.g. controllers), especially not for testing purposes.

rolfvdhulst (changes requested):
The class itself looks good, nice work!
There's some small changes I don't agree with, particularly the change to World:: adding the function for getting the Settings, and some minor unclarities /style things.

CMakeLists.txt
mrlukasbos: good catch. I have disabled it in a test for now but Google Mock is something really useful for writing tests. It allows you to temporarily override functions. In that case you can test functions that depend on that functions. 

So in my case I could create a scorematrix with artificial scores, testing whether the mapping was correct, instead of generating a worldstate and depending on a lot of functions from world and robot.

I disabled it because I needed some more flexibility and did it manually (see the test). but google mock is very nice to have.

src/utilities/Dealer.cpp
mrlukasbos: ah I was looking for that function but I could not find it. might have overlooked it.

test/helpers/WorldHelper.cpp
mrlukasbos: I  mean that it sometimes gives a segfault. (I couldn't find the problem quickly and it's outside the scope of this PR). It needs an issue.

test/UtilTests/DealerTest.cpp
mrlukasbos: yeah that was the old Google mock method

src/world_new/World.cpp
mrlukasbos: OK but i need to access the settings. I need to restructure some things in the test environment then. I do agree with you however.

src/utilities/Dealer.cpp
```diff
@@ -0,0 +1,122 @@
...
+    return factor * getDefaultFlagScores(robot, flag);
+}
+
+// TODO these values need to be tuned.
+double Dealer::getFactorForPriority(const Dealer::DealerFlag &flag) {
+    switch (flag.priority) {
+        case DealerFlagPriority::LOW_PRIORITY: return 1.0;
+        case DealerFlagPriority::MEDIUM_PRIORITY: return 2.0;
+        case DealerFlagPriority::HIGH_PRIORITY: return 3.0;
+    }
```
Yuhanun, line 81: return static_cast<int>(flag.priority) + 1.0;

rolfvdhulst: Personally think this is more explicit. Although your solution is less lines this is a lot easier to read.

rolfvdhulst (approved):

mrlukasbos:
Current status:
This PR breaks the history test, because world is globally defined.

emielsteerneman (approved):

mrlukasbos:
Is there already something decided on the historytest?

emielsteerneman:
@mrlukasbos Yes, I had a small conversation about this with Jesse. We will disable the test for now. We know the implementation of the history is correct because the test works. It only fails because world is a global, and the history is already partially filled when the test starts. In any case, the test is wrong, not the code.

PR #769, opened 2020-01-28
restructured play decider and play checker, adjusted applicationmanag…

RobotJente:
In this PR the STP architecture is set up for our system. I disabled the usage of it for our system since the Plays themselves are not ready yet. 

I realize there are some technical issues with shared_ptrs. I would like to discuss this with @Yuhanun  :) because I don't think I understand the concept of unique_ptrs. Also, the new world needs to be integrated into this, but I will do that in another PR


include/roboteam_ai/ApplicationManager.h
```diff
@@ -8,16 +8,14 @@
 #include <gtest/gtest_prod.h>
 #include <roboteam_utils/Timer.h>
 #include <utilities/StrategyManager.h>
-#include "analysis/PlayChecker.h"
+#include "include/roboteam_ai/analysis/play-utilities/PlayChecker.h"
```
Yuhanun: `#include "analysis/play-utilities/PlayChecker.h"

include/roboteam_ai/analysis/play-utilities/Play.h
```diff
@@ -18,22 +18,29 @@ namespace rtt::ai::analysis {
...
     bool isValidPlay(rtt::ai::world::World* world, rtt::ai::world::Field* field);
-
+    // TODO: Move this to the derived class
+    /**
+     * Returns a score based on how fitting this play is given a world and field state
+     * @param world the current world
+     * @param field the current field
+     * @return a score between 0 and 10, 10 being the best
+     */
+    int scorePlay(world::World* world, world::Field* field) {return 1;};
```
Yuhanun: @return a score between 0 and 10, great -> uint8_t

include/roboteam_ai/analysis/play-utilities/Play.h
```diff
@@ -18,22 +18,29 @@ namespace rtt::ai::analysis {
...
     std::string getName();
 
    protected:
-    std::vector<std::function<bool(world::World*, world::Field*)>> invariants;
+    /**
```
Yuhanun, line 41: why nothave both in a single protected block

include/roboteam_ai/analysis/play-utilities/PlayDecider.h
```diff
@@ -9,10 +9,26 @@
...
+
+        /**
+            * Decides which play to choose out of the valid plays. Selects the play that best fits the world and field state. If you want to
+            * do meta logic, like making the play selection more aggressive or risky, you should do it in this function
+            * @param world the current world state
+            * @param field the current field state
+            * @param validPlays a vector of plays which have been vetted by the PlayDecider and are deemed appropriate for this situaton
+            * @return the play that best fits the world, and field.
+            */
+        std::shared_ptr<Play> decideBestPlay(world::World *world, world::Field *field, std::vector<std::shared_ptr<Play>> validPlays);
```
Yuhanun: Why does this have to be a shared_ptr, it's all owned by the PlayChecker, unique_ptr ftw

include/roboteam_ai/treeinterp/BTFactory.h
```diff
@@ -38,7 +38,9 @@ class BTFactory {
 
     static bt::BehaviorTree::Ptr getKeeperTree();
 
-    static void setCurrentTree(const std::string &currentTree);
+    static void setCurrentTree(std::shared_ptr<rtt::ai::analysis::Play> );
```
Yuhanun: name your arguments, even in headers

src/ApplicationManager.cpp
```diff
@@ -15,7 +15,7 @@
 #include <world/World.h>
 #include <bt/Node.hpp>
 #include <utilities/GameStateManager.hpp>
-#include "analysis/PlayChecker.h"
+#include "include/roboteam_ai/analysis/play-utilities/PlayChecker.h"
```
Yuhanun: include style

src/ApplicationManager.cpp
```diff
@@ -27,8 +27,8 @@ namespace rtt {
 /// Start running behaviour trees. While doing so, publish settings and log the FPS of the system
 void ApplicationManager::start() {
     // create playcheck object here
-    playcheck = rtt::ai::analysis::PlayChecker();
-
+    playChecker = ai::analysis::PlayChecker();
+    playDecider = ai::analysis::PlayDecider();
```
Yuhanun: This is default initialized, You don't have to do this manually

src/analysis/play-utilities/Play.cpp
```diff
@@ -10,20 +10,24 @@
 #include "analysis/play-utilities/invariants/BallOnOurSideInvariant.h"
 
 namespace rtt::ai::analysis {
-Play::Play() {}
+    Play::Play() {}
 
-std::string Play::getName() { return name; }
+    std::string Play::getName() { return name; }
```
Yuhanun: string_view or const&

src/analysis/play-utilities/Play.cpp
```diff
@@ -10,20 +10,24 @@
 #include "analysis/play-utilities/invariants/BallOnOurSideInvariant.h"
 
 namespace rtt::ai::analysis {
-Play::Play() {}
+    Play::Play() {}
```
Yuhanun: = default

src/analysis/play-utilities/Play.cpp
```diff
@@ -10,20 +10,24 @@
...
 
-std::string Play::getName() { return name; }
+    std::string Play::getName() { return name; }
 
-Play::Play(std::string name, std::vector<std::function<bool(world::World *, world::Field *)>> invariants) {
-    this->name = name;
-    this->invariants = invariants;
-}
+    Play::Play(std::string name, std::vector<std::function<bool(world::World *, world::Field *)>> invariants) {
+        this->name = name;
```
Yuhanun:
```
Play::Play(std::string_view name, std::vector<std::function<...>> const& invariants)
   : name{ name }, invariants{ invariants } 
{}
```

Yuhanun (changes requested):

include/roboteam_ai/treeinterp/BTFactory.h
```diff
@@ -38,7 +38,9 @@ class BTFactory {
 
     static bt::BehaviorTree::Ptr getKeeperTree();
 
-    static void setCurrentTree(const std::string &currentTree);
+    static void setCurrentTree(std::shared_ptr<rtt::ai::analysis::Play> play);
```
rolfvdhulst: Is shared_ptr necessary? Or can you just use reference to unique_ptr?

rolfvdhulst: As in what is your ownershipmodel: does the BTFactory own the trees or are they owned somewhere else as well?

src/ApplicationManager.cpp
```diff
@@ -175,4 +171,14 @@ void ApplicationManager::notifyTreeStatus(bt::Node::Status status) {
...
 }
 
+    void ApplicationManager::decidePlay(ai::world::World *world, ai::world::Field *field) {
+        bool stillValidPlay = playChecker.update(world, field);
+        if (!stillValidPlay) {
+            auto bestplay = playDecider.decideBestPlay(world, field, playChecker.getValidPlays());
+            BTFactory::setCurrentTree(bestplay);
+        }
+        else {
+        }
```
rolfvdhulst: Empty else.

src/analysis/play-utilities/Play.cpp
```diff
@@ -10,20 +10,16 @@
...
-
-bool Play::isValidPlay(rtt::ai::world::World *world, rtt::ai::world::Field *field) {
-    return BallOnOurSideInvariant::isValid(world, field) && BallBelongsToUsInvariant::isValid(world, field) && AlwaysFalseInvariant::isValid(world, field) &&
-           AlwaysTrueInvariant::isValid(world, field);
-}
-
-void Play::setInvariants(const std::vector<std::function<bool(world::World *, world::Field *)>> &invariants) { this->invariants = invariants; }
+    const std::shared_ptr<bt::BehaviorTree> &Play::getTree() const {
+        return tree;
+    }
```
rolfvdhulst: If you use referencing anyways, why not make this a unique_ptr?

src/analysis/play-utilities/PlayChecker.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace rtt::ai::analysis {
+
+    bool
+    PlayChecker::checkCurrentGameInvariants(rtt::ai::world::World *world, rtt::ai::world::Field *field) { return true; }
+
+/**
+ * Determines what plays are viable given the current world, ref states and invariants/preconditions, and stores them in the validPlays vector
+ * TODO: add lambda here, to make it faster and cleaner
+ */
```
rolfvdhulst, line 22: I prefer if there were tabs in front of this.

include/roboteam_ai/analysis/play-utilities/PlayDecider.h
```diff
@@ -9,10 +9,26 @@
...
+
+    private:
+        /**
+          * local variable to keep track of which play is the best for the tick
+          */
+        std::shared_ptr<Play> bestPlay;
```
rolfvdhulst: Explicitly initialize this as a null pointer.

src/analysis/play-utilities/PlayDecider.cpp
```diff
@@ -6,5 +6,18 @@
...
+            std::cerr << "valid plays are empty" << std::endl;
+        }
+        for (auto const& play : validPlays) {
+            int temp = play->scorePlay(world, field);
+            if (temp > max) {
+                max = temp;
+            }
+            bestPlay = play;
+        }
+        return bestPlay;
```
rolfvdhulst: If bestPlay is not explicitly initialized as nullptr you might get UB here.

include/roboteam_ai/treeinterp/BTFactory.h
```diff
@@ -52,8 +54,7 @@ class BTFactory {
     static std::string currentTree;
     static std::string keeperTree;
     static bool weMadeTrees;
-
-    static std::shared_ptr<bt::BehaviorTree> testing_tree;
+    static std::shared_ptr<rtt::ai::analysis::Play> play;
```
rolfvdhulst: Personally am not a huge fan of using static variables unless necessary, which I dont' think it is here. BtFactory is already a static class.

include/roboteam_ai/analysis/play-utilities/Play.h
```diff
@@ -15,27 +15,34 @@ namespace rtt::ai::analysis {
...
+        /**
+         * @return true if all the invariants of this strategy are true
+         */
+        bool isValidPlay(rtt::ai::world::World *world, rtt::ai::world::Field *field);
+        // TODO: Move this to the derived class
+        /**
+         * Returns a score based on how fitting this play is given a world and field state (currently hardcoded, should be moved to derived classes)
+         * @param world the current world
+         * @param field the current field
+         * @return a score between 0 and 10, 10 being the best
```
rolfvdhulst: I would suggest not using a maximum in the future, this makes it easier to extend existing code rather than having to rescale a lot of current scores every time you add a new play.

rolfvdhulst (dismissed):
Generally looks good but see my comments.
Is there an explicit reason you are using shared_ptr's?

include/roboteam_ai/analysis/play-utilities/Play.h
RobotJente: While I agree with you that this way is not optimal, I am not sure how else we can do this. I think until we get a good idea on how to fix it this is okay to keep though

include/roboteam_ai/treeinterp/BTFactory.h
RobotJente: The variable belongs to the class. Therefore it needs to be static, right?

include/roboteam_ai/treeinterp/BTFactory.h
RobotJente: The plays vector is owned by a PlayChecker instance that stays alive for the program duration. I currently made it so that this is a vector of unique pointers. Whenever I need them, I pass the internal pointer from the unique pointers in the vector. Therefore the resource is only ever owned by 1 thing. Not sure if this is convoluted

include/roboteam_ai/ApplicationManager.h
```diff
@@ -8,16 +8,14 @@
...
 #include <roboteam_utils/Timer.h>
 #include <utilities/StrategyManager.h>
-#include "analysis/PlayChecker.h"
-#include "include/roboteam_ai/utilities/IOManager.h"
+#include "analysis/play-utilities/PlayChecker.h"
+#include "analysis/play-utilities/PlayDecider.h"
 #include "treeinterp/BTFactory.h"
-
+#include "utilities/IOManager.h"
 namespace rtt {
```
Yuhanun, line 15: newline pls

include/roboteam_ai/analysis/play-utilities/Play.h
```diff
@@ -15,29 +15,36 @@ using namespace rtt::ai::world;
...
-    const std::vector<std::function<bool(world::World *, world::Field *)>> &getInvariants() const;
+    Play() = default;
 
     /**
-     *
      * @return true if all the invariants of this strategy are true
      */
-    bool isValidPlay(rtt::ai::world::World *world, const Field *field);
+    bool isValidPlay(rtt::ai::world::World *world, const Field &field);
+    // TODO: Move this to the derived class
```
Yuhanun, line 27: .

include/roboteam_ai/analysis/play-utilities/Play.h
```diff
@@ -15,29 +15,36 @@ using namespace rtt::ai::world;
...
+    const std::shared_ptr<bt::BehaviorTree> &getTree() const;
 
    protected:
-    std::vector<std::function<bool(world::World *, world::Field *)>> invariants;
+    /**
+     * Internal tree of the play, where the execution of the play is done
+     */
     std::shared_ptr<bt::BehaviorTree> tree;
-    std::string name;
+    std::string_view name;
```
Yuhanun: now this should be a copy, so std::string, you should always own your own name

Yuhanun:
```cpp
struct MyStruct {
    std::string_view name;
    void init() {
       std::string data{ "Hello World" };
       name = data;
   } // data gets popped, the data it points to gets destroyed, name is now a dangling pointer
}
```

include/roboteam_ai/analysis/play-utilities/PlayChecker.h
```diff
@@ -0,0 +1,57 @@
...
+     */
+    std::vector<std::unique_ptr<Play>> allPlays;
+
+    /**
+     * Vector of all plays that are valid for the current world and field state
+     */
+    std::vector<Play *> validPlays;
+
+   public:
+    const std::vector<Play *> getValidPlays() const;
```
Yuhanun, line 42: document these better, make their lifetimes explicit

include/roboteam_ai/treeinterp/BTFactory.h
```diff
@@ -50,8 +53,7 @@ class BTFactory {
     static std::string currentTree;
     static std::string keeperTree;
     static bool weMadeTrees;
-
-    static std::shared_ptr<bt::BehaviorTree> testing_tree;
+    static rtt::ai::analysis::Play *play;
```
Yuhanun, line 56: the monstrosity is growing :eyes: 

(What's the plan in the refactor in BT tree)

src/ApplicationManager.cpp
```diff
@@ -59,7 +57,7 @@ void ApplicationManager::start() {
 void ApplicationManager::runOneLoopCycle() {
     if (weHaveRobots && io::io.hasReceivedGeom) {
         ai::analysis::GameAnalyzer::getInstance().start();
-        playcheck.update(rtt::ai::world::world, &io::io.getField());
+        decidePlay(world, io::io.getField());
```
Yuhanun, line 60: unrelated, but what's the point in assing field everywhere when it's global anyway

src/analysis/play-utilities/Play.cpp
```diff
@@ -5,25 +5,16 @@
...
-
-bool Play::isValidPlay(rtt::ai::world::World *world, const Field *field) {
-    return BallOnOurSideInvariant::isValid(world, field) && BallBelongsToUsInvariant::isValid(world, field) && AlwaysFalseInvariant::isValid(world, field) &&
-           AlwaysTrueInvariant::isValid(world, field);
+bool Play::isValidPlay(rtt::ai::world::World *world, const Field &field) {
+    return BallOnOurSideInvariant::isValid(world, field) && BallBelongsToUsInvariant::isValid(world, field) && AlwaysFalseInvariant::isValid(world, field);
 }
 
-void Play::setInvariants(const std::vector<std::function<bool(world::World *, world::Field *)>> &invariants) { this->invariants = invariants; }
+const std::shared_ptr<bt::BehaviorTree> &Play::getTree() const { return tree; }
```
Yuhanun, line 18: don't think this has to be const returned but is whatever

src/analysis/play-utilities/invariants/BallOnOurSideInvariant.cpp
```diff
@@ -13,14 +13,14 @@
...
     Vector2 ballPos = world->getBall()->getPos();
 
     // force the ball to be in the field
     bool inField = true;
 
     if (ballPos.x < 0) {
-        return abs(ballPos.x) < field->getFieldLength() / 2 && abs(ballPos.y) < field->getFieldWidth() / 2;
+        return abs(ballPos.x) < field.getFieldLength() / 2 && abs(ballPos.y) < field.getFieldWidth() / 2;
 
         return true;
```
Yuhanun, line 25:
```suggestion
```

Double return

include/roboteam_ai/analysis/play-utilities/Play.h
HaicoDorenbos, line 27: Huh? Why haven't you done this then yet?

include/roboteam_ai/analysis/play-utilities/invariants/AlwaysFalseInvariant.h
```diff
@@ -22,7 +22,7 @@ class AlwaysFalseInvariant {
      * @param field the current field state
      * @return true if the ball belongs to us, false otherwise
      */
-    static bool isValid(rtt::ai::world::World *world, const Field *field);
+    static bool isValid(rtt::ai::world::World *world, const Field &field);
```
HaicoDorenbos, line 25: Why do you make this function static?

src/analysis/play-utilities/PlayChecker.cpp
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+}
+
+bool PlayChecker::update(rtt::ai::world::World *world, const rtt::ai::world::Field &field) {
+    if (checkCurrentGameInvariants(world, field)) {
+        std::cout << "current play is still valid" << std::endl;
+        return true;
+    }
+    // Otherwise we select new plays for the playdecider by updating the validPlays vector of this object. This can then be accessed by the playdecider.
```
HaicoDorenbos, line 37: Put this inside the else block

src/analysis/play-utilities/PlayDecider.cpp
```diff
@@ -6,5 +6,18 @@
...
+Play* PlayDecider::decideBestPlay(world::World* world, const world::Field& field, std::vector<Play*> const& validPlays) {
+    int max = -1;
+    if (validPlays.empty()) {
+        std::cerr << "valid plays are empty" << std::endl;
+    }
+    for (auto& play : validPlays) {
+        int temp = play->scorePlay(world, field);
+        if (temp > max) {
+            max = temp;
+        }
```
HaicoDorenbos: This can be in 1 line using max = std::max(max, temp). I prefer shorter code over longer code

src/analysis/play-utilities/PlayDecider.cpp
```diff
@@ -6,5 +6,18 @@
...
+        bestPlay = play;
```
HaicoDorenbos: Huh? Should this not be placed inside the if statement?!

src/ApplicationManager.cpp
```diff
@@ -174,4 +172,13 @@ void ApplicationManager::notifyTreeStatus(bt::Node::Status status) {
             break;
     }
 }
-}  // namespace rtt
\ No newline at end of file
+
+void ApplicationManager::decidePlay(ai::world::World *world, const ai::world::Field &field) {
+    bool stillValidPlay = playChecker.update(world, field);
+    if (!stillValidPlay) {
```
HaicoDorenbos, line 178: if(!playerChecker.update(world, field)) is shorter. Do not declare variables that are only used once.

include/roboteam_ai/analysis/play-utilities/PlayChecker.h
```diff
@@ -0,0 +1,57 @@
...
+
+   private:
+    /// TODO: implement this function. Not a priority right now
+    bool checkStrategyPreconditions();
+
+    /**
+     * Determines which plays are valid by cycling through the allplays vector and seeing which plays' isValid() methods return true
+     * @param world
+     * @param field
+     */
```
HaicoDorenbos, line 52: You forgot to fill in the documentation for the params

include/roboteam_ai/analysis/play-utilities/PlayDecider.h
```diff
@@ -10,8 +10,24 @@
...
-    Play decideBestPlay(world::World *world, world::Field *field, std::vector<Play> validPlays);
+
+    /**
+     * Decides which play to choose out of the valid plays. Selects the play that best fits the world and field state. If you want to
+     * do meta logic, like making the play selection more aggressive or risky, you should do it in this function
+     * @param world the current world state
+     * @param field the current field state
+     * @param validPlays a vector of plays which have been vetted by the PlayDecider and are deemed appropriate for this situaton
+     * @return the play that best fits the world, and field.
+     */
```
HaicoDorenbos, line 23: This is what I like :smiley_cat:

HaicoDorenbos (dismissed):
Inside src/analysis/play-utilities/PlayDecider.cpp you most likely made an error. Therefore I cannot approve this.

src/ApplicationManager.cpp
mrlukasbos: I think the boolean name is more readable here.

src/analysis/play-utilities/PlayDecider.cpp
RobotJente: You are right, woops!

include/roboteam_ai/analysis/play-utilities/Play.h
RobotJente: Thats not for this PR

include/roboteam_ai/analysis/play-utilities/invariants/AlwaysFalseInvariant.h
RobotJente: Because it needs to be called from outside its class

Yuhanun (changes requested):

src/analysis/play-utilities/PlayDecider.cpp
```diff
@@ -6,5 +6,18 @@
 
 namespace rtt::ai::analysis {
 // Maybe plays should have some sort of score based on the world how good they are
-Play PlayDecider::decideBestPlay(world::World *world, world::Field *field, std::vector<Play> validPlays) { return Play(); }
+Play* PlayDecider::decideBestPlay(world::World* world, const world::Field& field, std::vector<Play*> const& validPlays) {
+    int max = -1;
+    if (validPlays.empty()) {
+        std::cerr << "valid plays are empty" << std::endl;
+    }
+    for (auto& play : validPlays) {
```
Yuhanun, line 14: Just wanted to throw this suggestion since haico said he preferred less code:

```cpp
bestPlay = std::max_element(validPlays.begin(), validPlays.end(), [&](auto& elem) {
    return elem->scorePlay(world, field);
}); // already a pointer to it :)
```

Yuhanun (approved):

src/analysis/play-utilities/PlayDecider.cpp
emielsteerneman: clear code > less code

PR #773, opened 2020-01-29
Feature/pass and play play

RobotJente:
Made a new Play called PassAndPlayPlay. It is integrated with the new play decisionmaking. The pass skill it uses is obsolete, but I will build a new Pass skill in another PR.

This PR needs #769 to be merged first

Edit for reviewers: All magic values are magic by design :dancer:


include/roboteam_ai/analysis/play-utilities/Play.h
```diff
@@ -12,30 +12,49 @@
...
+        std::string_view getName();
+
+        const std::shared_ptr<bt::BehaviorTree> &getTree() const;
+
+    protected:
+        /**
+         * Internal tree of the play, where the execution of the play is done
+         */
+        std::shared_ptr<bt::BehaviorTree> tree;
+        std::string name;
```
Yuhanun, line 56: Why? Have setters and getters in the base class and then just call those from the subclasses

Base::setTree(tree);
Base::setName(name);

Yuhanun (approved):
I think this is the same as the other PR?

PR #775, opened 2020-01-29
Fix UB in worlddata.cpp for us and them vectors

mrlukasbos:
While creating tests for dealer I got UB in one of my robots in 'us'. I traced it back to this line: ```us.emplace_back(&robots.emplace_back(feedback, r, Team::us)); ```
Where the contents of `robots` is okay, but the contents of `us` is not.

The problem was that the reference returned by `emplace_back` is invalidated when the vector resizes.


Yuhanun (approved):

Yuhanun:
Looks good, merge asap

Issue #781, opened 2020-02-03
PositionControl must use RobotViews


Ratoone:
Replace the Robot objects from the position control with the nonowning RobotViews

_Originally posted by @Yuhanun in https://github.com/RoboTeamTwente/roboteam_ai/pull/765_

Ratoone:
Solved with #796

PR #782, opened 2020-02-03
Refactor/roboteam world optimize

Yuhanun:
getRobotsNonOwning is now O(1) instead of O(n)

Solves #781


mrlukasbos:
I don't see how this solves #781

emielsteerneman:
@mrlukasbos Currently, there is no way to get a RobotView vector of all robots on the field. There is this `std::vector<rtt::world_new::robot::Robot> const &getRobots() const` which does return constants, but not RobotViews. This implements that.

emielsteerneman (approved):
Looks good, but did you apply the linter?

Yuhanun:
I think I did?

emielsteerneman:
Travis fails, again because of an error in the tests. However, it builds locally and looks solid. Ready to be merged imo.

Yuhanun:
I don't think it's because of the tests, I think it's because it doesn't compile

Yuhanun:
> I don't see how this solves #781

I told paul to go ahead and use getRobotsNonOwning() already even though it would allocate and create an O(n) copy, now it's O(1), i told him to make the issue for that

emielsteerneman:
> I don't think it's because of the tests, I think it's because it doesn't compile

Yeah that's what I meant, my bad. An error in the code of the test, giving a compile error

emielsteerneman:
Well that works I guess, but I'm not reassured knowing that some tests aren't used

PR #783, opened 2020-02-03
Removed "using namespace std;"

Yuhanun:
Removed all occurences of `using namespace std;`, all of which were in headers.

Generally `using namespace std;` isn't really, uhh... good

(Even removed it from external libraries cause we include them <3 )


RobotJente (approved):
Woaw didn't know this! Nice work :)

emielsteerneman:
From which external libraries did you remove this?

emielsteerneman (approved):

PR #785, opened 2020-02-04
Make BT classes a library

rolfvdhulst:
This is basically an attempt to separate the good parts of our code from the bad, and make the good part a library so that it's harder for people to fuck up and easier to understand the structure.

I made all of the BT classes which were relatively clean (which is everything except for Leaf and DefaultTactic ) a library which can compile as a standalone, and linked this to roboteam_ai. 
This has a couple of advantages:

- Harder to hack: If you want to use other files from AI you will need to specifically link them against the bt target.
- The structure is not just enforced through directories but also through our build system.
- Slightly better compilation time (the compiler can more easily utilize multiple cores by compiling multiple libraries at the same time)
- The compiler has an easier time detecting optimizations on well-structured code.

Additionally, I did some small style fixes such as consistently using .h instead of mixing .hpp and .h.

I did not do this for Leaf and DefaultTactic because they are both very ugly classes which have a shitton of references to external files (robotDealer, analysis etc.), and because both are more extensions of behaviortrees to RobotSoccer rather than just behaviortrees themselves.


Yuhanun:
I do not agree with changing everything from hpp -> h

Rest seems fine but i thosuld be more separated

rolfvdhulst:
> 
> 
> I do not agree with changing everything from hpp -> h
> 
> Rest seems fine but i thosuld be more separated

I don´t really follow you. Why not? It was inconsistent right now (mixture of h and hpp was being used).

What do you mean with more seperated? This is all of the simple bt classes in one library for now. Clearly we can do more (e.g. all world stuff in a library) but that's for other PR's.

Yuhanun (approved):

PR #787, opened 2020-02-05
Feature/ccache

Yuhanun:
Added ccache support, to install just install ccache


Yuhanun:
I branched from wrong branch, oops

PR #788, opened 2020-02-05
Adde ccache support

Yuhanun:
Added ccache support, to install just apt install ccache


mrlukasbos (approved):
I added mac support but the brew include is a little bit different.

PR #792, opened 2020-02-11
Control is now a library

Yuhanun:
Updated control folder to now be a statically linked library known as rtt::control by cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,21 +195,6 @@ set(CONDITIONS_SOURCES
         ${PROJECT_SOURCE_DIR}/src/conditions/ResumePlayAfterPenalty.cpp
         )
 
-set(CONTROL_SOURCES
-        ${PROJECT_SOURCE_DIR}/src/control/ControlUtils.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/BasicPosControl.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/PosController.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/numtrees/NumTreePosControl.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/numtrees/PathPoint.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/numtrees/Collision.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/shot-controllers/ShotController.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/ball-handling/BallHandlePosControl.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/ball-handling/DribbleBackwards.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/ball-handling/DribbleForwards.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/ball-handling/RotateAroundBall.cpp
-        ${PROJECT_SOURCE_DIR}/src/control/ball-handling/RotateWithBall.cpp
-        )
-
 set(TEST_SOURCES
         ${PROJECT_SOURCE_DIR}/test/main.cpp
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -322,20 +307,21 @@ set(AI_LIBS
...
+        PUBLIC include/roboteam_ai
         PRIVATE ${SDL2_INCLUDE_DIRS}
         PRIVATE ${SDL2IMAGE_INCLUDE_DIRS}
         )
 
 # Main Executable
 add_executable(roboteam_ai src/roboteam_ai.cpp ${AI_SOURCES})
 target_link_libraries(roboteam_ai ${AI_LIBS})
-target_include_directories(roboteam_ai ${AI_INCLUDES})
+include_directories(roboteam_ai ${AI_INCLUDES})
```
rolfvdhulst, line 324: No. This is literally the worst cmake command man has ever created.

CMakeLists.txt
```diff
@@ -322,20 +307,21 @@ set(AI_LIBS
...
         PRIVATE roboteam_proto
         PRIVATE rtt::bt
+        PRIVATE rtt::control
         PRIVATE ${SDL2_LIBRARY}
         )
 
 set(AI_INCLUDES
         INTERFACE include
-        PRIVATE include/roboteam_ai
+        PUBLIC include/roboteam_ai
```
rolfvdhulst, line 316: wtf, no. This shouldnt be necessary.

src/control/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+        ball-handling/RotateWithBall.cpp
+        )
+
+find_package(Qt5Charts REQUIRED)
+find_package(Qt5Widgets REQUIRED)
+find_package(Qt5Gui REQUIRED)
+
+target_include_directories(control
+        PUBLIC ../../include/roboteam_ai
+        PUBLIC ../..
```
rolfvdhulst, line 22: Yeah.... If you're not going to refactor the directory inclusion and linking then you're not really solving a problem at all, just creating more by making it a library.

rolfvdhulst (changes requested):
Not gonna lie, made some raging comments first but deleted them (mostly) again.

The right idea is there but if you just hack CMake around it really does not make a difference whether or not a piece of code is a library. The point of making something a library is that it's independent and compiles on it's own given that you linked other code to it correctly. If you just include all of the roboteam_ai directories then you're not really creating modular code at all, and it will only cause more problems down the road. I don't know how easy it is to make the code more modular since that depends on it's includes. If it has includes from all over the place in roboteam_ai it might be very hard to decouple the code. The idea of a library should be that generally

1- it's headers are self-contained: E.g. you don't need to directly include external headers other than the ones from libraries you have linked to it (e.g. roboteam_utils) which are passed through linking rather than directly including their paths. This helps a lot to prevent unnecessary recompilation of files.
2- The folder structure should resemble that of a library (e.g. there's a folder for the library which contains the src and include folders of that library)
3- When you link the library against other compilation units, you give those units the public headers (e.g. the headers open for external use.) There can be headers or other files which are only used internally.

If you want we can talk more CMake/project structure later but this PR as of now really does not help anyone.

CMakeLists.txt
Yuhanun: I changed that because the includes wouldn't work in submodule, will revert, good catch

CMakeLists.txt
Yuhanun: I changed that because the includes wouldn't work in submodule, will revert, good catch

src/control/CMakeLists.txt
Yuhanun: True, will take a look at it tomorrow :)

RobotJente:
What's the status on this PR? Can it be closed?

PR #796, opened 2020-02-13
Position Control Robot -> RobotViews

Ratoone:
This PR solves #781. It also solved an UB with copying the robotView vector from WorldDataView on return


Ratoone:
For some reason it added the commits from the previous branch. There should only be the last 4 commits

emielsteerneman (approved):
Reviewed. Tested with GrSim

PR #800, opened 2020-02-14
Refactor/interface world new

emielsteerneman:
The old world implementation should be removed from the interface everywhere. There is still some stuff in Output, but I don't know what to do with that, because it's filled with statics

### DIFF
--- a/include/roboteam_ai/interface/widgets/widget.h
+++ b/include/roboteam_ai/interface/widgets/widget.h
@@ -29,8 +29,8 @@ class Visualizer : public QWidget {
     using Robot = rtt::ai::world::Robot;
     using RobotPtr = std::shared_ptr<Robot>;
     explicit Visualizer(const rtt::world_new::World& worldManager, QWidget *parent = nullptr);
-    const std::vector<Robot> &getSelectedRobots() const;
-    bool robotIsSelected(Robot robotToCheck);
+    const std::unordered_map<int, rtt::world_new::view::RobotView> &getSelectedRobots() const;
+    bool robotIsSelected(rtt::world_new::view::RobotView robot);
     bool robotIsSelected(int id);
 
    public slots:
@@ -42,7 +42,7 @@ class Visualizer : public QWidget {
     void setShowRobotInvalids(bool showPath);
     void setShowBallPlacementMarker(bool showMarker);
     void setShowDebugValueInTerminal(bool showDebug);
-    void toggleSelectedRobot(int robotId);
+    void toggleSelectedRobot(rtt::world_new::view::RobotView robot);
     void setToggleFieldDirection(bool inversed);
 
    protected:
@@ -84,7 +84,8 @@ class Visualizer : public QWidget {
                           QColor>> tacticColors;  // map colors to tactic to visualize which robots work together
     int tacticCount = 0;                          // increases when a new tactic is used
 
-    std::vector<Robot> selectedRobots;
+//    std::vector<Robot> selectedRobots;
+    std::unordered_map<int, rtt::world_new::view::RobotView> selectedRobots;
 
     // toggles
     bool showRoles = Constants::STD_SHOW_ROLES();
--- a/include/roboteam_ai/interface/widgets/widget.h
+++ b/include/roboteam_ai/interface/widgets/widget.h
@@ -19,7 +19,6 @@
 #include <include/roboteam_ai/world_new/World.hpp>
 
 namespace rtt::ai::interface {
-using namespace rtt::ai::world;
 
 class Visualizer : public QWidget {
     Q_OBJECT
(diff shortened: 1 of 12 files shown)


emielsteerneman:
There are still some parts in widget.cpp that are not yet implemented, because these things are not yet in RobotView

include/roboteam_ai/interface/widgets/widget.h
```diff
@@ -41,25 +38,26 @@ class Visualizer : public QWidget {
...
     void setToggleFieldDirection(bool inversed);
 
    protected:
     void paintEvent(QPaintEvent *event) override;
     void mousePressEvent(QMouseEvent *event) override;
 
    private:
-    float factor;
+    const rtt::world_new::World &worldManager;
+    float factor{};
```
rolfvdhulst, line 50: wtf is this bracket initialization? Have never seen this before.

src/roboteam_ai.cpp
```diff
@@ -77,7 +78,10 @@ int main(int argc, char *argv[]) {
     // initialize the interface
     QApplication a(argc, argv);
     setDarkTheme();
-    window = std::make_shared<ui::MainWindow>();
+
+    // Todo make this a not-global-static thingy
+    rtt::world_new::World* worldManager = rtt::world_new::World::instance();
+    window = std::make_shared<ui::MainWindow>(*worldManager);
     window->setWindowState(Qt::WindowMaximized);
```
rolfvdhulst, line 85: Can't you const ref from application manager? Application manager should also own an instance. Having 2 instances is quite ugly...

src/interface/widgets/widget.cpp
```diff
@@ -258,15 +260,15 @@ void Visualizer::drawBall(QPainter &painter) {
 }
 
 // draw the robots
-void Visualizer::drawRobots(QPainter &painter) {
+void Visualizer::drawRobots(QPainter &painter, rtt::world_new::view::WorldDataView world) {
     // draw us
-    for (auto &robot : rtt::ai::world::world->getUs()) {
-        drawRobot(painter, *robot, true);
+    for (rtt::world_new::view::RobotView robot : world->getUs()) {
```
rolfvdhulst, line 265: const &?

rolfvdhulst: oh wait nvm these things are pointers right? Not sure.

src/interface/widgets/widget.cpp
```diff
@@ -337,68 +337,42 @@ void Visualizer::drawRobot(QPainter &painter, Robot robot, bool ourTeam) {
         painter.drawText(robotpos.x, ypos += 20, QString::fromStdString(getRoleNameForRobot(robot)));
     }
 
+    // Todo : Get working stuff in RobotView
```
rolfvdhulst, line 340: ?

src/interface/widgets/widget.cpp
```diff
@@ -337,68 +337,42 @@ void Visualizer::drawRobot(QPainter &painter, Robot robot, bool ourTeam) {
...
-                painter.setBrush(Qt::green);
-            } else {
-                // yellow to indicate feedback is not okay
-                painter.setPen(Qt::red);
-                painter.setBrush(Qt::red);
-            }
-            painter.drawEllipse({(int)robotpos.x + 10, (int)robotpos.y - 10}, 2, 2);
-        }
-    }
+    // Todo : Get working feedback in RobotView
```
rolfvdhulst, line 356: Way to do this is by making feedback processing a roboteam_world thing. Make this an issue (not critical for now).

rolfvdhulst (commented):
I looked at all the non-linter changes and they look pretty decent. 

You might want to re-enable all of the prints you commented as it might mess up people who expect them to be printed.

src/roboteam_ai.cpp
rolfvdhulst: Maybe I am understanding it wrong here though.

include/roboteam_ai/interface/widgets/widget.h
emielsteerneman: Ah, clang did that for me. It was complaining that it was uninitialized

src/interface/widgets/widget.cpp
emielsteerneman: They are RobotViews, which only hold a single pointer to actual robot data. They wouldn't have to be const but I'll add it anyway for consistency

src/interface/widgets/widget.cpp
emielsteerneman: Woops, forgot to remove that

src/roboteam_ai.cpp
emielsteerneman: I'm getting the global instance here. There is always exactly one instance. Always.

src/interface/widgets/widget.cpp
emielsteerneman: Issue made

rolfvdhulst (approved):

PR #802, opened 2020-02-14
cleanup iomanager

### LINKED ISSUE #695: Refactor GameAnalyzer
Game analyzer is not used very much but it does use quite some cpu. It can be heavily simplified. Currently it uses the Singleton pattern, which is considered a bad pattern. It is also globally accessible which is also bad. 

- [ ] Find usages of game analyzer
- [ ] Design one or multiple new solutions to get the same kind of behavior, but cleaner
- [ ] remove/refactor the game analyzer

mrlukasbos:
Quite a large refactor adopting the new world structure.

- [x] IOmanager is mostly single purpose again. it stores recent proto messages and application manager receives messages when needed. This should remove most race conditions in our system already. 
- [x] game analyzer is not a singleton anymore. I realized that generating a full 'analysis report' is often unnecessary to calculate and the performance hit is just too big. I converted it to static function definitions: just call one when you need it. This needs restructuring in a later PR. 
- [x] some necessary field refactors in some coaches, game analyzer, etc. 

fixes #695

### DIFF
--- a/include/roboteam_ai/ApplicationManager.h
+++ b/include/roboteam_ai/ApplicationManager.h
@@ -45,10 +45,10 @@ class ApplicationManager {
     void start();
     void checkForShutdown();
     void checkForFreeRobots();
-    void updateCoaches() const;
+    void updateCoaches(const ai::world::Field & field) const;
     void updateTrees();
-    bt::Node::Status runStrategyTree();
-    void runKeeperTree();
+    bt::Node::Status runStrategyTree(const ai::world::Field & field);
+    void runKeeperTree(const ai::world::Field & field);
 };
 
 }  // namespace rtt
--- a/include/roboteam_ai/coach/defence/PossiblePass.h
+++ b/include/roboteam_ai/coach/defence/PossiblePass.h
@@ -6,6 +6,7 @@
 #define ROBOTEAM_AI_POSSIBLEPASS_H
 
 #include <roboteam_utils/Vector2.h>
+#include <include/roboteam_ai/world_new/views/RobotView.hpp>
 #include "utilities/Constants.h"
 #include "world/Field.h"
 #include "world/Robot.h"
@@ -16,23 +17,23 @@ using namespace rtt::ai::world;
 
 class PossiblePass {
    public:
-    world::Robot toBot;
+    world_new::view::RobotView toBot;
     Vector2 startPos;
     Vector2 endPos;
     const double distance();
     bool obstacleObstructsPath(const Vector2 &obstPos, double obstRadius = Constants::ROBOT_RADIUS() + Constants::BALL_RADIUS());
-    int amountOfBlockers(const world::WorldData &world);
-    PossiblePass(world::Robot _toBot, const Vector2 &ballPos);
-    double score(const Field &field, const world::WorldData &world);
+    int amountOfBlockers(std::vector<world_new::view::RobotView> us, std::vector<world_new::view::RobotView> them);
+    PossiblePass(world_new::view::RobotView _toBot, const Vector2 &ballPos);
+    double score(const Field &field, std::vector<world_new::view::RobotView> us, std::vector<world_new::view::RobotView> them);
     // scale from startPos to EndPos
     Vector2 posOnLine(double scale);
     double faceLine();
 
    private:
     Vector2 botReceivePos(const Vector2 &startPos, const Vector2 &botPos);
-    double penaltyForBlocks(const world::WorldData &world);
+    double penaltyForBlocks(std::vector<world_new::view::RobotView> us, std::vector<world_new::view::RobotView> them);
     double penaltyForDistance();
-    double scoreForGoalAngle(const Field &field, const world::WorldData &world);
+    double scoreForGoalAngle(const Field &field, std::vector<world_new::view::RobotView> us, std::vector<world_new::view::RobotView> them);
 };
 
 }  // namespace rtt::ai::coach
--- a/include/roboteam_ai/coach/defence/PossiblePass.h
+++ b/include/roboteam_ai/coach/defence/PossiblePass.h
@@ -1,16 +1,9 @@
-//
-// Created by rolf on 5-4-19.
-//
-
 #ifndef ROBOTEAM_AI_POSSIBLEPASS_H
 #define ROBOTEAM_AI_POSSIBLEPASS_H
 
 #include <roboteam_utils/Vector2.h>
 #include <include/roboteam_ai/world_new/views/RobotView.hpp>
 #include "utilities/Constants.h"
-#include "world/Field.h"
-#include "world/Robot.h"
-#include "world/WorldData.h"
 
 namespace rtt::ai::coach {
 using namespace rtt::ai::world;
@@ -20,7 +13,7 @@ class PossiblePass {
     world_new::view::RobotView toBot;
     Vector2 startPos;
     Vector2 endPos;
-    const double distance();
+    double distance();
     bool obstacleObstructsPath(const Vector2 &obstPos, double obstRadius = Constants::ROBOT_RADIUS() + Constants::BALL_RADIUS());
     int amountOfBlockers(std::vector<world_new::view::RobotView> us, std::vector<world_new::view::RobotView> them);
     PossiblePass(world_new::view::RobotView _toBot, const Vector2 &ballPos);
@@ -33,7 +26,7 @@ class PossiblePass {
     Vector2 botReceivePos(const Vector2 &startPos, const Vector2 &botPos);
     double penaltyForBlocks(std::vector<world_new::view::RobotView> us, std::vector<world_new::view::RobotView> them);
     double penaltyForDistance();
-    double scoreForGoalAngle(const Field &field, std::vector<world_new::view::RobotView> us, std::vector<world_new::view::RobotView> them);
+    double scoreForGoalAngle(const Field &field, std::vector<world_new::view::RobotView> us);
 };
 
 }  // namespace rtt::ai::coach
--- a/include/roboteam_ai/coach/heuristics/CoachHeuristics.h
+++ b/include/roboteam_ai/coach/heuristics/CoachHeuristics.h
@@ -9,6 +9,7 @@
 #include <world/Field.h>
 #include <world/World.h>
 #include <world/WorldData.h>
+#include <include/roboteam_ai/world_new/views/WorldDataView.hpp>
 
 namespace rtt::ai::coach {
 using namespace rtt::ai::world;
@@ -31,17 +32,17 @@ class CoachHeuristics {
 
    public:
     static double calculateCloseToGoalScore(const Field &field, const Vector2 &position);
-    static double calculateShotAtGoalScore(const Field &field, const Vector2 &position, const WorldData &world);
-    static double calculatePassLineScore(const Vector2 &position, const WorldData &world);
-    static double calculateBehindBallScore(const Vector2 &position, const WorldData &world);
-    static double calculatePassDistanceToBallScore(const Field &field, const Vector2 &position, const WorldData &world);
-    static double calculatePositionDistanceToBallScore(const Field &field, const Vector2 &position, const WorldData &world);
+    static double calculateShotAtGoalScore(const Field &field, const Vector2 &position, world_new::view::WorldDataView world);
+    static double calculatePassLineScore(const Vector2 &position, world_new::view::WorldDataView world);
+    static double calculateBehindBallScore(const Vector2 &position, world_new::view::WorldDataView world);
+    static double calculatePassDistanceToBallScore(const Field &field, const Vector2 &position, world_new::view::WorldDataView world);
+    static double calculatePositionDistanceToBallScore(const Field &field, const Vector2 &position, world_new::view::WorldDataView world);
     static double calculateAngleToGoalScore(const Field &field, const Vector2 &position);
 
     /// Currently not implemented, but might be again later
     static double calculateDistanceToOpponentsScore(const Vector2 &position);
     static double calculateDistanceToClosestTeamMateScore(const Vector2 &position, int thisRobotID = -1);
-    static double getClosestOpponentAngleToPassLine(const Vector2 &position, const WorldData &world, double smallestAngle = 999999);
+    static double getClosestOpponentAngleToPassLine(const Vector2 &position, world_new::view::WorldDataView world, double smallestAngle = 999999);
 };
 
 }  // namespace rtt::ai::coach
--- a/include/roboteam_ai/coach/heuristics/OffensiveScore.h
+++ b/include/roboteam_ai/coach/heuristics/OffensiveScore.h
@@ -26,7 +26,7 @@ class OffensiveScore {
 
    public:
     using WorldData = world::WorldData;
-    double calculateOffensivePositionScore(const Vector2 &zoneLocation, const Vector2 &position, const WorldData &world, const Field &field);
+    double calculateOffensivePositionScore(const Vector2 &zoneLocation, const Vector2 &position, world_new::view::WorldDataView world, const Field &field);
 };
 
 }  // namespace rtt::ai::coach
--- a/src/ApplicationManager.cpp
+++ b/src/ApplicationManager.cpp
@@ -70,11 +70,11 @@ void ApplicationManager::runOneLoopCycle() {
             this->notifyTreeStatus(status);
         } else {
             std::cout << "[ApplicationManager::runOneLoopCycle] No robots from our team in world yet" << std::endl;
-            std::this_thread::sleep_for(std::chrono::milliseconds (1000));
+            std::this_thread::sleep_for(std::chrono::milliseconds (100));
         }
     } else {
         std::cout << "[ApplicationManager::runOneLoopCycle] No field yet" << std::endl;
-        std::this_thread::sleep_for(std::chrono::milliseconds (1000));
+        std::this_thread::sleep_for(std::chrono::milliseconds (100));
     }
     /*
      * This is a hack performed at the robocup.
--- a/src/ApplicationManager.cpp
+++ b/src/ApplicationManager.cpp
@@ -1,7 +1,3 @@
-//
-// Created by mrlukasbos on 14-1-19.
-//
-
 #include <ApplicationManager.h>
 #include <bt/Node.h>
 #include <coach/GetBallCoach.h>
--- a/src/ApplicationManager.cpp
+++ b/src/ApplicationManager.cpp
@@ -35,8 +35,7 @@ void ApplicationManager::start() {
             t.limit([&]() {
                     ai::interface::Input::setFps(amountOfCycles * fpsUpdateRate);
                     amountOfCycles = 0;
-                },
-                fpsUpdateRate);
+                }, fpsUpdateRate);
 
             // publish settings, but limit this function call to only run 1 times/s at most
             t.limit([&]() { io::io.publishSettings(SETTINGS.toMessage()); }, 1);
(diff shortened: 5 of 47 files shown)


include/roboteam_ai/ApplicationManager.h
```diff
@@ -45,10 +45,10 @@ class ApplicationManager {
     void start();
     void checkForShutdown();
     void checkForFreeRobots();
-    void updateCoaches() const;
+    void updateCoaches(const ai::world::Field & field) const;
     void updateTrees();
-    bt::Node::Status runStrategyTree();
-    void runKeeperTree();
+    bt::Node::Status runStrategyTree(const ai::world::Field & field);
+    void runKeeperTree(const ai::world::Field & field);
```
rolfvdhulst, line 51: Nice.

src/analysis/GameAnalyzer.cpp
```diff
@@ -145,12 +145,12 @@ bool GameAnalyzer::isClosingInToGoal(const Field &field, RobotPtr robot, bool ou
...
 }
 
-void GameAnalyzer::start(int iterationsPerSecond) {
+void GameAnalyzer::start(world::Field const & field, int iterationsPerSecond) {
     if (!running && world::world->weHaveRobots()) {
         std::cout << "GameAnalyzer: "
                   << "Starting at " << iterationsPerSecond << " iterations per second" << std::endl;
         auto delay = (unsigned)(1000.0 / iterationsPerSecond);
-        thread = std::thread(&GameAnalyzer::loop, this, delay);
+        thread = std::thread(&GameAnalyzer::loop, this, field, delay);
```
rolfvdhulst: This should give problems since the const & field will not change after GameAnalyzer is started. I feel like a pointer/something else should work better.

rolfvdhulst: As in, if the field changes after GameAnalyzer is started we have a problem.

rolfvdhulst (commented):
Looks good in general, I like how complexity from IOManager is being removed and put to better abstraction layers. I do think there is a bug in there however with GameAnalyzer (see comments).

I am OK with Field going to World from a functional point of view, however it does kind of irk me since Field is specifically about long-lived data where world is about short-lived data. Also within the class itself you don't need data from one to the other. I feel like we should maybe have something like a 'context' struct which contains the latest WorldData, FieldMessage, referee info and Settings; all of the most relevant input data as members and that is it. Mixing abstraction layers that don't need to call eachother is usually an indication that you have too little layers.

src/world/World.cpp
```diff
@@ -8,7 +8,10 @@ namespace rtt::ai::world {
 World worldObj;
 World *world = &worldObj;
 
-void World::updateWorld(const Field &field, const proto::World &message) {
+void World::updateWorld(const proto::SSL_GeometryFieldSize &fieldMessage, const proto::World &worldMessage) {
+    ai::Field field(fieldMessage);
+    this->field = field;
```
emielsteerneman, line 13: Does this work? What happens to "this->field" if "field" goes out of scope?

mrlukasbos: nothing bad it is an implicit copy

mrlukasbos:
Field in World is a 'temporary' solution. I wanted to get it out of IOManager at least.

The const reference to the interface is probably a bug indeed. Although I am not sure if we would want a pointer.... we might want to think about a new method of communicating data to the interface. (Like rust channels or something similar). I know ER-Force uses QT slots and signals for this purpose. I do not specifically like that, but it might be the best option compared to other things i can think of.

rolfvdhulst:
> 
> 
> Field in World is a 'temporary' solution. I wanted to get it out of IOManager at least.
> 
> The const reference to the interface is probably a bug indeed. Although I am not sure if we would want a pointer.... we might want to think about a new method of communicating data to the interface. (Like rust channels or something similar). I know ER-Force uses QT slots and signals for this purpose. I do not specifically like that, but it might be the best option compared to other things i can think of.

I am a bit confused why you talk about the interface? The bug I think exists, is in GameAnalyzer. Since GameAnalyzer barely takes any performance we probably can solve this by not making it a seperate thread anymore.

For a interface decoupling I have proposed to use protobuf/something else as a sole communication channel between interface and AI. Then you can just run the interface in a seperate thread and synchronize information every AI tick by something like: 

interfaceOutput = Ai.tick(interfaceInput). Meaning you don't need to constantly monitor and forward all input from interface to AI. You just aggregate all interface inputs of the last 1/60th second. Similarly, you also aggregate all output from the AI into one message and send it to the interface. You can aggregate the interface information using QT Slots (I don't suggest doing this for our AI as QT's signal/slot communication rate is limited at around ~1 ms, and thus may give a lot of delay).

Won't be easy and will be a lot of work though. I personally heaviliy dislike ER-Force's solution to this. It's very messy and hard to keep track of. They pass around massive protobuf objects between a ton of classes and allow each of these classes to write and read from them. This is almost worse than static globals IMO.

src/analysis/GameAnalyzer.cpp
RobotJente: It is already given a const ref to field as params in the start function, so I think this should work

mrlukasbos:
I am currently restructuring the game analyzer as well. It was a singleton running in a separate thread. Although it's results where almost always unused and it gave quite a performance hit. I am removing the singleton pattern and I am making a clear division of functions.  I expect this to give a performance improvement.

RobotJente (commented):
I like the direction this PR is going in. The changes are becoming large though, I'd like to do an intermediate merge into development. Do you think we are almost at a point where that can de done?

mrlukasbos:
No. I’m stuck with the defenders now. New world doesn’t properly support them.

mrlukasbos:
This PR is ready for reviews. Would like some critical reviews. Don't forget to check things like keeper, normal play (especially defense and offense) field side switching, color switching, running with a referee, etc. 

-->Please check if the defenders are doing proper defending! 

I suspect (no guarantees) a performance improvement since I can now run normal play on mac, while previously I could not. This would make sense because game analyzer is no more, which had quite a performance hit.

src/ApplicationManager.cpp
```diff
@@ -55,18 +50,40 @@ void ApplicationManager::start() {
...
-        decidePlay(world, io::io.getField());
-        updateTrees();
-        updateCoaches();
-        runKeeperTree();
-        Status status = runStrategyTree();
-        this->notifyTreeStatus(status);
+
+    if (io::io.hasReceivedGeom) {
+        auto fieldMessage = io::io.getGeometryData().field();
+        auto worldMessage = io::io.getWorldState();
```
RobotJente, line 50: This means world messages are updated every tick and not whenever a new one is available, right?

RobotJente (approved):

src/ApplicationManager.cpp
mrlukasbos: exactly! This approach reduces the risk of race conditions or other weird bugs due to data changes while doing calculations.

mrlukasbos:
Read more about the performance of this PR here: https://github.com/RoboTeamTwente/roboteam_ai/issues/815 

Restructuring iOManager gave a 20% increase in performance on my machine.

include/roboteam_ai/coach/defence/PossiblePass.h
```diff
@@ -6,6 +6,7 @@
 #define ROBOTEAM_AI_POSSIBLEPASS_H
 
 #include <roboteam_utils/Vector2.h>
+#include <include/roboteam_ai/world_new/views/RobotView.hpp>
 #include "utilities/Constants.h"
 #include "world/Field.h"
 #include "world/Robot.h"
```
timovdk: This include should be removed since the old world/robot.h file should not be used anymore (right?)

include/roboteam_ai/coach/heuristics/CoachHeuristics.h
```diff
@@ -9,6 +9,7 @@
 #include <world/Field.h>
 #include <world/World.h>
 #include <world/WorldData.h>
```
timovdk, line 11: These includes should not be necessary anymore

include/roboteam_ai/coach/heuristics/OffensiveScore.h
```diff
@@ -26,7 +26,7 @@ class OffensiveScore {
 
    public:
     using WorldData = world::WorldData;
```
timovdk, line 28: Old world is used in this using

src/ApplicationManager.cpp
```diff
@@ -1,21 +1,16 @@
...
 #include <bt/Node.h>
 #include <coach/GetBallCoach.h>
 #include <coach/OffensiveCoach.h>
 #include <coach/PassCoach.h>
 #include <coach/defence/DefenceDealer.h>
-#include <include/roboteam_ai/world/FieldComputations.h>
+#include <include/roboteam_ai/world_new/World.hpp>
 #include <interface/api/Input.h>
 #include <roboteam_utils/Timer.h>
 #include <world/World.h>
```
timovdk, line 10: Old world

src/coach/midfield/MidFieldCoach.cpp
```diff
@@ -3,9 +3,8 @@
 //
 
 #include "coach/midfield/MidFieldCoach.h"
-#include <world/WorldData.h>
+#include <include/roboteam_ai/world_new/World.hpp>
 #include "world/Ball.h"
```
timovdk, line 7: Old world

src/conditions/HasClearShot.cpp
```diff
@@ -12,6 +12,7 @@
 #include <world/FieldComputations.h>
 #include <world/World.h>
 #include <world/WorldData.h>
```
timovdk, line 14: Old world

src/world/FieldComputations.cpp
```diff
@@ -7,6 +7,8 @@
 #include <interface/api/Input.h>
 #include "world/World.h"
 #include "world/WorldData.h"
```
timovdk: Old world

src/treeinterp/tactics/DefaultTactic.cpp
```diff
@@ -1,12 +1,13 @@
 #include "treeinterp/tactics/DefaultTactic.h"
-
+#include "world/BallPossession.h"
 #include <analysis/GameAnalyzer.h>
 #include <world/World.h>
 #include <world/WorldData.h>
```
timovdk, line 5: Old world

src/utilities/IOManager.cpp
```diff
@@ -1,22 +1,12 @@
...
+#include <include/roboteam_ai/utilities/GameStateManager.hpp>
 #include "roboteam_proto/DemoRobot.pb.h"
 #include "roboteam_proto/RobotFeedback.pb.h"
 #include "roboteam_proto/messages_robocup_ssl_geometry.pb.h"
 
 #include "interface/api/Input.h"
-#include "utilities/GameStateManager.hpp"
 #include "utilities/Pause.h"
-#include "world/FieldComputations.h"
 #include "world/Robot.h"
```
timovdk, line 10: Old world

timovdk (commented):
I am not sure if it was the intention to also completely remove the old world from the changed files, if yes, there  are still some usages in tactics, coaches and controllers. If not I will approve this PR

mrlukasbos:
No that was not the intention, the intention was to do a part of it. Changing world in one place makes it change in 10 other files so I suggest doing that in another PR

timovdk (approved):

src/world_new/Robot.cpp
```diff
@@ -24,9 +24,6 @@ Robot::Robot(std::unordered_map<uint8_t, proto::RobotFeedback> &feedback, const
     if (id < 16) {
         workingDribbler = ai::Constants::ROBOT_HAS_WORKING_DRIBBLER(id);
         workingBallSensor = ai::Constants::ROBOT_HAS_WORKING_BALL_SENSOR(id);
-    } else {
-        std::cerr << "Warning: Creating robot with id = " << id << std::endl;
-        assert(false);
```
rolfvdhulst, line 29: What are the implications of this?

src/world/FieldComputations.cpp
```diff
@@ -302,5 +265,51 @@ Polygon FieldComputations::getFieldEdge(const Field &field, double margin) {
...
+    return visibleParts;
+}
+
+
+/*
+ * Get the visible parts of a goal
+ * This is the inverse of getting the blockades of a goal
+ */
+std::vector<Line> FieldComputations::getVisiblePartsOfGoal(const Field &field, bool ourGoal, const Vector2 &point, world_new::view::WorldDataView &world) {
+    return getVisiblePartsOfGoalByObstacles(field, ourGoal, point, world.getRobotsNonOwning());
```
rolfvdhulst: Same. If this gets all the robots it's likely a problematic behavior change because we did not consider the opponent robots as obstacles.

src/world/FieldComputations.cpp
```diff
@@ -33,16 +35,16 @@ double FieldComputations::getTotalGoalAngle(const Field &field, bool ourGoal, co
 }
 
 /// id and ourteam are for a robot not to be taken into account.
-double FieldComputations::getPercentageOfGoalVisibleFromPoint(const Field &field, bool ourGoal, const Vector2 &point, const world::WorldData &data, int id, bool ourTeam) {
+double FieldComputations::getPercentageOfGoalVisibleFromPoint(const Field &field, bool ourGoal, const Vector2 &point, world_new::view::WorldDataView &world, int id, bool ourTeam) {
     double goalWidth = field.getGoalWidth();
     double blockadeLength = 0;
-    for (auto const &blockade : getBlockadesMappedToGoal(field, ourGoal, point, data, id, ourTeam)) {
+    for (auto const &blockade : getBlockadesMappedToGoal(field, ourGoal, point, world.getRobotsNonOwning(), id, ourTeam)) {
```
rolfvdhulst, line 35: Does this get all of the robots? This might be problematic as previously we would only check for our robots, not all. Function name and call don't reallly make much clear though.

src/interface/widgets/mainWindow.cpp
```diff
@@ -123,7 +123,7 @@ MainWindow::MainWindow(const rtt::world_new::World &worldManager, QWidget *paren
     // update mainwindow and field visualization
     auto *timer = new QTimer(this);
     connect(timer, SIGNAL(timeout()), this, SLOT(update()));
-    timer->start(40);  // 25fps
+    timer->start(200);  // 5fps
```
rolfvdhulst, line 117: Might want to undo this mac change for you to run it at ok speeds.

src/ApplicationManager.cpp
```diff
@@ -55,18 +45,37 @@ void ApplicationManager::start() {
...
+            Status status = runStrategyTree(field);
+            this->notifyTreeStatus(status);
+        } else {
+            std::cout << "[ApplicationManager::runOneLoopCycle] No robots from our team in world yet" << std::endl;
+            std::this_thread::sleep_for(std::chrono::milliseconds (1000));
+        }
     } else {
-        std::this_thread::sleep_for(std::chrono::microseconds(100000));
+        std::cout << "[ApplicationManager::runOneLoopCycle] No field yet" << std::endl;
+        std::this_thread::sleep_for(std::chrono::milliseconds (1000));
```
rolfvdhulst: This is a bit too long for my taste, but ok.

src/ApplicationManager.cpp
```diff
@@ -55,18 +45,37 @@ void ApplicationManager::start() {
...
+
+            decidePlay(world, field);
+            updateTrees();
+            updateCoaches(field);
+            runKeeperTree(field);
+            Status status = runStrategyTree(field);
+            this->notifyTreeStatus(status);
+        } else {
+            std::cout << "[ApplicationManager::runOneLoopCycle] No robots from our team in world yet" << std::endl;
+            std::this_thread::sleep_for(std::chrono::milliseconds (1000));
```
rolfvdhulst: I would really prefer not to sleep if we have no robots. It can happen that vision blacks out for half a second and this is really long to wait.

rolfvdhulst: Like this is fine if it's 100 ms or less but a second is too much

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -173,24 +185,32 @@ DefenderBot DefencePositionCoach::createBlockOnLine(const Field &field, const Po
...
+    bot.blockFromID = pass.toBot->getId();
     bot.targetPos = blockPoint;
     bot.orientation = pass.faceLine();
     return bot;
 }
-std::shared_ptr<Vector2> DefencePositionCoach::blockOnDefenseAreaLine(const Field &field, const PossiblePass &pass, const world::WorldData &world) {
-    auto visibleParts = FieldComputations::getVisiblePartsOfGoal(field, true, pass.endPos, world);
+
+std::shared_ptr<Vector2> DefencePositionCoach::blockOnDefenseAreaLine(const Field &field, const PossiblePass &pass) {
+    std::vector<Line> visibleParts = simulatedVisibleGoalParts(field, pass);
```
rolfvdhulst, line 195: It's quite hard to check if this behavior is still the same.

rolfvdhulst: I might be wrong but I think this is not the same behavior because previously only our bots were passed to the calculation.

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -173,24 +185,32 @@ DefenderBot DefencePositionCoach::createBlockOnLine(const Field &field, const Po
...
     std::sort(visibleParts.begin(), visibleParts.end(), [](const Line &a, const Line &b) { return abs(a.end.y - a.start.y) > abs(b.end.y - b.start.y); });
     if (!visibleParts.empty()) {
         return blockOnDefenseLine(field, visibleParts[0], pass.endPos);
     }
     return nullptr;
 }
+std::vector<Line> DefencePositionCoach::simulatedVisibleGoalParts(const Field &field, const PossiblePass &pass) const {// create a vector with all robots
+    std::vector<world_new::view::RobotView> allSimulatedRobots = simulatedRobotsUs;
+    allSimulatedRobots.insert(allSimulatedRobots.end(), simulatedRobotsThem.begin(), simulatedRobotsThem.end());
+    auto visibleParts = FieldComputations::getVisiblePartsOfGoalByObstacles(field, true, pass.endPos, allSimulatedRobots);
```
rolfvdhulst: This is probably too many robots, we tend to only consider our robots for obstacles.

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -271,33 +289,36 @@ std::shared_ptr<double> DefencePositionCoach::pickNewPosition(const Field &field
...
     return nullptr;
 }
 
-world::WorldData DefencePositionCoach::setupSimulatedWorld(const Field &field) {
-    world::WorldData sWorld = world::world->getWorld();
-    sWorld.us.clear();
-    sWorld = getTheirAttackers(field, sWorld);  // we select only the relevant robots
-    return sWorld;
+void DefencePositionCoach::setupSimulatedWorld(const Field &field) {
+    auto us = world_new::World::instance()->getWorld()->getUs();
```
rolfvdhulst: You don't need this call :)

rolfvdhulst (changes requested):
DId not have time to run the code. Quite certain there is a bug now in the defenders as you're using all of the robots for goal vision computations when previously we would only use our robots for the calculation. There's some minor things, but most of the recent changes look fine to me.

src/world_new/Robot.cpp
mrlukasbos: I think it was a nice failsafe because it was thought that 'dummy' robot object never would be made. 

maybe it would be nicer to have a dummyRobot class extend on Robot, but i'd do so later.

mrlukasbos:
I see you are correct, will fix it soon.

src/coach/defence/DefencePositionCoach.cpp
mrlukasbos: what call do you mean?

rolfvdhulst: the us call. Us is unused.

rolfvdhulst (approved):

mrlukasbos:
Did a final test with GrSim and everything looked fine.

PR #807, opened 2020-02-21
Feature/new skill class for integration with new dealer

RobotJente:
Made a new skill class that does not rely on robotdealer nor leaf. It is currently not used anywhere and contains some unimplemented features in the form of TODO's. I chose not to fix them in this PR to keep it short. Please feel free to comment on this, and suggest improvements. However, I would like to merge it asap so I can start working on the next steps, because that will probably reveal more problems

Edit: as this was unclear from my description, the purpose of this PR is not to fix all of Skill's problems (although there is definitely good reason to do so, but that would make this PR very large). The main purpose is to refactor skill and later role in such a way that it can work with the new dealer

### DIFF
--- a/include/roboteam_ai/new_skills/NewSkill.h
+++ b/include/roboteam_ai/new_skills/NewSkill.h
@@ -5,10 +5,72 @@
 #ifndef RTT_NEWSKILL_H
 #define RTT_NEWSKILL_H
 
+#include <include/roboteam_ai/treeinterp/Leaf.h>
+
+#ifndef ROBOTEAM_AI_SKILL_H
+#define ROBOTEAM_AI_SKILL_H
+
+#include <roboteam_utils/Angle.h>
+#include "include/roboteam_ai/utilities/IOManager.h"
+#include "roboteam_proto/RobotCommand.pb.h"
+#include "treeinterp/Leaf.h"
+
 namespace rtt::ai {
-    class NewSkill {
 
-    };
-
+// forward declare control Utils
+namespace control {
+class ControlUtils;
 }
-#endif //RTT_NEWSKILL_H
+
+namespace world {
+class Robot;
+class Ball;
+class WorldData;
+}  // namespace world
+
+/**
+ * \class Skill
+ * \brief Base class for all skills. Provides no additional functionality.
+ */
+class NewSkill : public bt::Leaf {
+   private:
+    /**
+     * used if we are playing on the other side of the field. By default, our side is left.
+     * @param cmd the command to be rotated
+     * @return the command, but as though we are playing on the other side of the field
+     */
+    proto::RobotCommand rotateRobotCommand(proto::RobotCommand &cmd);
+
+   protected:
+    /**
+     * Gives the robotcommand to the global instance of IO, which then publishes the command
+     */
+    void publishRobotCommand();
+    /**
+     * Creates a new, empty RobotCommand and assigns it to "command"
+     */
+    void refreshRobotCommand();
+    proto::RobotCommand command;
+
+    /**
+     * If any values in the robotcommand are beyond what the robot can handle, it caps these values and raises an error message
+     */
+    void limitRobotCommand();
+
+   public:
+    NewSkill(std::string name, bt::Blackboard::Ptr blackboard = nullptr);
+    std::string node_name() override;
+    void initialize() override;
+    Status update() override;
+    void terminate(Status s) override;
+    virtual void onInitialize() = 0;
+    virtual Status onUpdate() = 0;
+    virtual void onTerminate(Status s) = 0;
+    void refreshRobotPositionControllers();
+};
+
+}  // namespace rtt::ai
+
+#endif  // ROBOTEAM_AI_SKILL_H
+
+#endif  // RTT_NEWSKILL_H
--- a/include/roboteam_ai/new_skills/NewSkill.h
+++ b/include/roboteam_ai/new_skills/NewSkill.h
@@ -6,27 +6,12 @@
 #define RTT_NEWSKILL_H
 
 #include <include/roboteam_ai/treeinterp/Leaf.h>
-
-#ifndef ROBOTEAM_AI_SKILL_H
-#define ROBOTEAM_AI_SKILL_H
-
 #include <roboteam_utils/Angle.h>
 #include "include/roboteam_ai/utilities/IOManager.h"
 #include "roboteam_proto/RobotCommand.pb.h"
 #include "treeinterp/Leaf.h"
 
 namespace rtt::ai {
-
-// forward declare control Utils
-namespace control {
-class ControlUtils;
-}
-
-namespace world {
-class Robot;
-class Ball;
-class WorldData;
-}  // namespace world
 
 /**
  * \class Skill
@@ -71,7 +56,5 @@ class NewSkill : public bt::Leaf {
 };
 
 }  // namespace rtt::ai
-
-#endif  // ROBOTEAM_AI_SKILL_H
 
 #endif  // RTT_NEWSKILL_H
--- a/include/roboteam_ai/new_skills/NewSkill.h
+++ b/include/roboteam_ai/new_skills/NewSkill.h
@@ -1,0 +1,14 @@
+//
+// Created by jessevw on 21.02.20.
+//
+
+#ifndef RTT_NEWSKILL_H
+#define RTT_NEWSKILL_H
+
+namespace rtt::ai {
+    class NewSkill {
+
+    };
+
+}
+#endif //RTT_NEWSKILL_H
--- a/include/roboteam_ai/new_skills/NewSkill.h
+++ b/include/roboteam_ai/new_skills/NewSkill.h
@@ -43,7 +43,8 @@ class NewSkill : public bt::Leaf {
 
    protected:
     /**
-     * Gives the robotcommand to the global instance of IO, which then publishes the command
+     * Gives the robotcommand to the global instance of IO, which then publishes the command. After this function is
+     * called, the internal RobotCommand "command" is also reset by refreshRobotCommand()
      */
     void publishRobotCommand();
     /**
--- a/src/new_skills/NewSkill.cpp
+++ b/src/new_skills/NewSkill.cpp
@@ -8,6 +8,7 @@
 #include <include/roboteam_ai/utilities/Settings.h>
 
 #include <cmath>
+#include <include/roboteam_ai/world_new/views/RobotView.hpp>
 
 #include "control/ControlUtils.h"
 #include "utilities/Constants.h"
@@ -32,7 +33,7 @@ namespace rtt::ai {
         limitRobotCommand();
 
         if (command.id() == -1) {
-            std::cerr << "the command is trying to send a RobotCommand to id " << command.id() << std::endl;
+            std::cerr << "[NewSkill::publishRobotCommand] Trying to send a RobotCommand to id " << command.id() << std::endl;
             if (robot && robot->id != -1) {
                 command.set_id(robot->id);
                 io::io.publishRobotCommand(command);  // We default to our robots being on the left if parameter is not set
@@ -51,11 +52,11 @@ namespace rtt::ai {
         updateRobot();
         ball = world->getBall();  // update ball position
         if (!robot){
-            std::cerr << "The robot is not initialized" << std::endl;
+            std::cerr << "[NewSkill::update] The robot is not initialized" << std::endl;
             return Status::Failure;
         }
         if (robot->id == -1) {
-            std::cerr << "The robot id is" << robot->id << std::endl;
+            std::cerr << "[NewSkill::update] The robot id is" << robot->id << std::endl;
             return Status::Failure;
         }
         if (!ball) {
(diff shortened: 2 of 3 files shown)


include/roboteam_ai/new_skills/NewSkill.h
```diff
@@ -0,0 +1,77 @@
...
+#ifndef RTT_NEWSKILL_H
+#define RTT_NEWSKILL_H
+
+#include <include/roboteam_ai/treeinterp/Leaf.h>
+
+#ifndef ROBOTEAM_AI_SKILL_H
+#define ROBOTEAM_AI_SKILL_H
+
+#include <roboteam_utils/Angle.h>
+#include "include/roboteam_ai/utilities/IOManager.h"
```
Ratoone: Include naming. Also, why do you have two header guards?

include/roboteam_ai/new_skills/NewSkill.h
```diff
@@ -0,0 +1,77 @@
...
+   public:
+    NewSkill(std::string name, bt::Blackboard::Ptr blackboard = nullptr);
+    std::string node_name() override;
+    void initialize() override;
+    Status update() override;
+    void terminate(Status s) override;
+    virtual void onInitialize() = 0;
+    virtual Status onUpdate() = 0;
+    virtual void onTerminate(Status s) = 0;
+    void refreshRobotPositionControllers();
```
Ratoone, line 55: Why here?

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,136 @@
...
+            io::io.publishRobotCommand(command);  // We default to our robots being on the left if parameter is not set
+        }
+        // refresh the robotcommand after it has been sent
+        refreshRobotCommand();
+    }
+
+    std::string NewSkill::node_name() { return name; }
+
+    NewSkill::Status NewSkill::update() {
+        robotId = -15;// TODO: after the tactic is implemented, needs to be some way the skill figures out which id to send to. Should be known by the tactic
```
Ratoone, line 51: The robot will be assigned anyway by the robot dealer, right? Why not take directly the id of the robot?

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "utilities/RobotDealer.h"
+#include "world/Ball.h"
+#include "world/Robot.h"
+#include "world/World.h"
+
+namespace rtt::ai {
+
+    NewSkill::NewSkill(std::string name, bt::Blackboard::Ptr blackboard) : bt::Leaf(std::move(name), std::move(blackboard)) {
+        robot = std::make_shared<Robot>();
+        ball = std::make_shared<Ball>();
```
Ratoone, line 25: I guess this was taken from the old skil. Yet, if you assign the robot and the ball at initialize anyway, there is not really a need to create them here.

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,136 @@
...
+        }
+        if (!ball) {
+            return Status::Waiting;
+        }
+        return onUpdate();
+    }
+
+    void NewSkill::initialize() {
+        // TODO: be able to get robot from dealer
+        // robot = dealer.distributions.at(role).robot;
```
Ratoone, line 70: I'm really curious how this would work. I think it would be better if the skill didn't even have access to the robot - just its ID. I suppose that the bt::Leaf already has a reference to a robot

Ratoone (changes requested):
Includes and naming mostly. Some things related to design are not entirely clear, but I guess that some of them are also related to the Leaf

include/roboteam_ai/new_skills/NewSkill.h
RobotJente: Whoops I copied it from the old skill and forgot to change it. I basically removed the things I thought we didnt need as quite a lot of functionality can be kept from the old implementation

include/roboteam_ai/new_skills/NewSkill.h
RobotJente: You mean why is the refreshRobotPositionControllers() here?

src/new_skills/NewSkill.cpp
RobotJente: Yeah you are right I think

include/roboteam_ai/new_skills/NewSkill.h
```diff
@@ -0,0 +1,77 @@
...
+// forward declare control Utils
+namespace control {
+class ControlUtils;
+}
+
+namespace world {
+class Robot;
+class Ball;
+class WorldData;
+}  // namespace world
```
emielsteerneman: Why is this forward declaration here? Including the correct header should work

src/new_skills/NewSkill.cpp
RobotJente: Yeah it is definitely not meant to happen here, imo this should happen at play strategic level and then each role should be coupled to an ID, after which the roles are executed. Each role has some priority decided beforehand in the form of flags so it can be best assigned to a robot

src/new_skills/NewSkill.cpp
RobotJente: Well what should be here is that the role tells the Skill "hey you're controlling this robot, his ID is x". But I havent gotten aroudn to implementing that yet. Is that an answer to your question?

include/roboteam_ai/new_skills/NewSkill.h
```diff
@@ -0,0 +1,77 @@
...
+     */
+    void publishRobotCommand();
+    /**
+     * Creates a new, empty RobotCommand and assigns it to "command"
+     */
+    void refreshRobotCommand();
+    proto::RobotCommand command;
+
+    /**
+     * If any values in the robotcommand are beyond what the robot can handle, it caps these values and raises an error message
```
emielsteerneman, line 42: What the robot can handle or what the gamestate allows

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include "utilities/RobotDealer.h"
+#include "world/Ball.h"
+#include "world/Robot.h"
+#include "world/World.h"
+
+namespace rtt::ai {
+
+    NewSkill::NewSkill(std::string name, bt::Blackboard::Ptr blackboard) : bt::Leaf(std::move(name), std::move(blackboard)) {
+        robot = std::make_shared<Robot>();
+        ball = std::make_shared<Ball>();
```
emielsteerneman, line 25: Why are these shared pointers? Shouldn't these be RobotViews and BallViews?

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    void NewSkill::publishRobotCommand() {
+        if (!SETTINGS.isLeft()) {
+            command = rotateRobotCommand(command);
+        }
+
+        limitRobotCommand();
+
+        if (command.id() == -1) {
+            std::cerr << "the command is trying to send a RobotCommand to id " << command.id() << std::endl;
```
emielsteerneman: Please change to `std::cerr << "[NewSkill::publishRobotCommand] ..`

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,135 @@
...
+            if (robot && robot->id != -1) {
```
emielsteerneman, line 37: id = -1 should not be needed if RobotViews are used

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,135 @@
...
+            io::io.publishRobotCommand(command);  // We default to our robots being on the left if parameter is not set
+        }
+        // refresh the robotcommand after it has been sent
+        refreshRobotCommand();
+    }
+
+    std::string NewSkill::node_name() { return name; }
+
+    NewSkill::Status NewSkill::update() {
+        robotId = -15;// TODO: after the tactic is implemented, needs to be some way the skill figures out which id to send to. Should be known by the tactic
```
emielsteerneman, line 51: What??

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,135 @@
...
+        updateRobot();
+        ball = world->getBall();  // update ball position
+        if (!robot){
+            std::cerr << "The robot is not initialized" << std::endl;
```
emielsteerneman: Please add `std::cerr << "[NewSkill::update] ...`

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,135 @@
...
+            return Status::Failure;
+        }
+        if (robot->id == -1) {
+            std::cerr << "The robot id is" << robot->id << std::endl;
```
emielsteerneman: Please add `std::cerr << "[NewSkill::update] ...`

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,135 @@
...
+        bool isKeeper = command.id() == robotDealer::RobotDealer::getKeeperID();
+
+        auto limitedVel = Vector2(command.vel().x(), command.vel().y());
+        limitedVel = control::ControlUtils::velocityLimiter(limitedVel);
+        if (!(isDefendPenaltyState && isKeeper)) {
+            limitedVel = control::ControlUtils::accelerationLimiter(limitedVel, robot->getPidPreviousVel(), command.w());
+        }
+        robot->setPidPreviousVel(limitedVel);
+        if (std::isnan(limitedVel.x) || std::isnan(limitedVel.y)) {
+            std::cout << "ERROR in NewSkill::limitRobotCommand() : The robot has a velocity of NaN: " << node_name().c_str() << "!"
```
emielsteerneman: Please add `std::cerr << "[NewSkill::limitRobotCommand] ...`

src/new_skills/NewSkill.cpp
```diff
@@ -0,0 +1,135 @@
...
+        }
+        command.mutable_vel()->set_x(limitedVel.x);
+        command.mutable_vel()->set_y(limitedVel.y);
+    }
+
+    void NewSkill::refreshRobotPositionControllers() {
+        robot->resetNumTreePosControl();
+        robot->resetShotController();
+        robot->resetBallHandlePosControl();
+        robot->resetBasicPosControl();
```
emielsteerneman, line 130: I still think that the controllers should not be accessed through the robot object

emielsteerneman (changes requested):

include/roboteam_ai/new_skills/NewSkill.h
RobotJente: I copied this code from the old skill class, however, I don't see the problem with forward declarations. They are faster than includes in compile time

src/new_skills/NewSkill.cpp
RobotJente: The class still inherits through leaf, which is being refactored atm. I will change it later though, this is more of a structural PR than functional

src/new_skills/NewSkill.cpp
RobotJente: Will do

src/new_skills/NewSkill.cpp
RobotJente: I realize my comment can come over as cryptic. However, which part are you asking about?

src/new_skills/NewSkill.cpp
RobotJente: I don't have a view on this, but that is also not the purpose of this PR. I'm trying to get a stub moving for a new implementation of role/skill that is able to work with the new dealer.

timovdk (approved):
As a start for the new structure it looks fine. There is still work left to do like deciding where controllers will live, implementing the world_new in this new structure and how the tactics and plays work with skill. Since this new structure is not used yet, it seems fine to merge to be able to get this new structure working.

src/new_skills/NewSkill.cpp
timovdk: I agree with Emiel on this

mrlukasbos: I am interested why @emielsteerneman and @timovdk think this. Last year we explicitly moved the controllers from skills towards robot, because otherwise the PID controllers would reset when switching skills, resulting in spectacular collisions and backflip robots. Therefore, from a control viewpoint it makes a lot of sense to put the controllers in Robot. 

--> my main point is that each robot should have it's own set of controllers that stay alive for the whole runtime duration of AI

timovdk: Since with the RobotViews the idea was to have data only, it does not immediately make sense to have controllers in there as well. It is more a structural difference in my opinion.

rolfvdhulst: I agree with Timo and Emiel. The controllers need to be long lived but this is not the place to do so. Robots can get deleted or replaced in world whilst this would have no implications for control at all.

 I would much rather see a separate entity that manages the controllers somewhere else than here. Whether this is the right place and time to fix it, I don't know.

mrlukasbos: (just throwing in some thoughts)

@timovdk That makes sense indeed. On the other hand, the controller does contain robot-specific data. In fact, a controller is a set of functions with some robot-specific parameters. maybe the controllers can be made into static functions while the input data for these controllers is stored in robot? That would have the implication that robot is mutable however. I agree it is all not ideal.

Another option is to have a set of controllers for each robot which is stored in (for example) a map or some ControlManager. this object then needs to live in applicationmanager and be passed around. This would already be quite nice. But maybe there is a better solution...

timovdk: I would personally prefer the second option but as @mrlukasbos said, there might even be better options. That is why I am also not sure whether this is the right place and time to fix this.

Yuhanun: I think I also disagree with putting the controllers in the robot class, I don't think it makes sense to do so.

Not sure what a good solution would be though

PR #808, opened 2020-02-21
Fix/robot ball distance

emielsteerneman:
The function whichRobotHasBall didn't work, because the variable Robot::distanceToBall was never set. Added setting of distanceToBall and fixed a bug in whichRobotHasBall. Also added a test for whichRobotHasBall.

### DIFF
--- a/include/roboteam_ai/world_new/Robot.hpp
+++ b/include/roboteam_ai/world_new/Robot.hpp
@@ -13,8 +13,8 @@
 #include "control/shot-controllers/ShotController.h"
 #include "roboteam_proto/WorldRobot.pb.h"
 #include "roboteam_utils/Angle.h"
-#include "world_new/views/BallView.hpp"
 #include "world_new/Team.hpp"
+#include "world_new/views/BallView.hpp"
 
 namespace rtt::world_new::robot {
 
@@ -141,8 +141,7 @@ class Robot {
 
    public:
     explicit Robot(std::unordered_map<uint8_t, proto::RobotFeedback> &feedback, const proto::WorldRobot &copy, Team team = both,
-                   std::optional<rtt::world_new::view::BallView> ball = std::nullopt, unsigned char dribblerState = 0,
-                   unsigned long worldNumber = 0);
+                   std::optional<rtt::world_new::view::BallView> ball = std::nullopt, unsigned char dribblerState = 0, unsigned long worldNumber = 0);
 
     Robot &operator=(Robot &) = delete;
 
--- a/include/roboteam_ai/world_new/Robot.hpp
+++ b/include/roboteam_ai/world_new/Robot.hpp
@@ -13,6 +13,7 @@
 #include "control/shot-controllers/ShotController.h"
 #include "roboteam_proto/WorldRobot.pb.h"
 #include "roboteam_utils/Angle.h"
+#include "world_new/views/BallView.hpp"
 #include "world_new/Team.hpp"
 
 namespace rtt::world_new::robot {
@@ -37,7 +38,6 @@ class Robot {
     Vector2 pidPreviousVel;
 
     double distanceToBall;
-    bool iHaveBall{};
     unsigned long lastUpdatedWorldNumber = 0;
 
     double angularVelocity;
@@ -140,7 +140,8 @@ class Robot {
     [[nodiscard]] unsigned long getLastUpdatedWorldNumber() const noexcept;
 
    public:
-    explicit Robot(std::unordered_map<uint8_t, proto::RobotFeedback> &feedback, const proto::WorldRobot &copy, Team team = both, unsigned char dribblerState = 0,
+    explicit Robot(std::unordered_map<uint8_t, proto::RobotFeedback> &feedback, const proto::WorldRobot &copy, Team team = both,
+                   std::optional<rtt::world_new::view::BallView> ball = std::nullopt, unsigned char dribblerState = 0,
                    unsigned long worldNumber = 0);
 
     Robot &operator=(Robot &) = delete;
--- a/include/roboteam_ai/world_new/views/WorldDataView.hpp
+++ b/include/roboteam_ai/world_new/views/WorldDataView.hpp
@@ -162,7 +162,7 @@ class WorldDataView {
      * Checks whether a robot has the ball
      * @param id Id of robot to check
      * @param ourTeam whether or not it should be fetched from our team
-     * @param maxDist maximum distance for "having ball"
+     * @param maxDist Maximum distance the robot is allowed to be from the ball for "having ball"
      * @return true if the robot with id has the ball, false if not
      */
     [[nodiscard]] bool robotHasBall(uint8_t id, bool ourTeam, double maxDist = ai::Constants::MAX_BALL_RANGE()) const noexcept;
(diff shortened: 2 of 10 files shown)


include/roboteam_ai/world_new/Robot.hpp
```diff
@@ -140,8 +140,8 @@ class Robot {
     [[nodiscard]] unsigned long getLastUpdatedWorldNumber() const noexcept;
 
    public:
-    explicit Robot(std::unordered_map<uint8_t, proto::RobotFeedback> &feedback, const proto::WorldRobot &copy, Team team = both, unsigned char dribblerState = 0,
-                   unsigned long worldNumber = 0);
+    explicit Robot(std::unordered_map<uint8_t, proto::RobotFeedback> &feedback, const proto::WorldRobot &copy, Team team = both,
+                   std::optional<rtt::world_new::view::BallView> ball = std::nullopt, unsigned char dribblerState = 0, unsigned long worldNumber = 0);
```
RobotJente, line 144: Why is the dribblerState an unsigned char?

include/roboteam_ai/world_new/views/WorldDataView.hpp
```diff
@@ -188,7 +188,7 @@ class WorldDataView {
      * @param team Team enum of team to fetch from
      * @return A non-owning view of the robot that has the ball
      */
-    [[nodiscard]] RobotView whichRobotHasBall(Team team = both);
+    [[nodiscard]] std::optional<RobotView> whichRobotHasBall(Team team = both, double maxDist = ai::Constants::MAX_BALL_BOUNCE_RANGE());
```
RobotJente: Can you update the params doc for this function? It is not clear to me what the max_ball_bounce_range has to do with which robot has the ball

src/world/World.cpp
```diff
@@ -17,13 +17,14 @@ void World::updateWorld(const Field &field, const proto::World &message) {
 
         // create a worldData if there is none
         if (!worldDataPtr) {
-            std::cout << "Creating first world" << std::endl;
+            std::cout << "[World::updateWorld] Creating first world" << std::endl;
```
RobotJente: Nice, clear log messages :+1:

src/world_new/Ball.cpp
```diff
@@ -114,12 +114,12 @@ void Ball::updateBallAtRobotPosition() noexcept {
...
 
-    auto robotWithBall = world->whichRobotHasBall();
-    if (!robotWithBall) {
+    std::optional<rtt::world_new::view::RobotView> robotWithBall = world->whichRobotHasBall();
+    if (!robotWithBall.has_value()) {
         return;
     }
 
-    std::pair<uint8_t, Team> newRobotIdTeam = {robotWithBall->getId(), robotWithBall->getTeam()};
+    std::pair<uint8_t, Team> newRobotIdTeam = {(*robotWithBall)->getId(), (*robotWithBall)->getTeam()};
```
RobotJente, line 122: Was this change what fixed a bug? I know this caused UB before as well...

src/world_new/Robot.cpp
```diff
@@ -116,10 +117,6 @@ double Robot::getDistanceToBall() const noexcept { return distanceToBall; }
 
 void Robot::setDistanceToBall(double _distanceToBall) noexcept { Robot::distanceToBall = _distanceToBall; }
 
-bool Robot::isIHaveBall() const noexcept { return iHaveBall; }
```
RobotJente, line 119: https://civilization-v-achievements.fandom.com/wiki/I_Can_Has_Nukes%3F

src/world_new/WorldData.cpp
```diff
@@ -23,21 +30,15 @@ WorldData::WorldData(proto::World &protoMsg, rtt::Settings const &settings, std:
     them.reserve(amountThem);
 
     for (auto &each : ours) {
-        auto _ptr = &robots.emplace_back(feedback, each, Team::us);
+        auto _ptr = &robots.emplace_back(feedback, each, Team::us, getBall());
```
RobotJente, line 33: Why are robots given a ball as well?

src/world_new/views/WorldDataView.cpp
```diff
@@ -102,17 +102,16 @@ RobotView WorldDataView::whichRobotHasBall(Team team) {
...
-                bestRobot = robot;
-            }
+        if (0 <= robot->getDistanceToBall() && robot->getDistanceToBall() < bestDistance) {
+            bestRobot = robot;
+            bestDistance = robot->getDistanceToBall();
         }
     }
 
-    return bestRobot;
+    return bestRobot.get() == nullptr ? std::nullopt : std::optional<RobotView>(bestRobot);
```
RobotJente, line 114: Is this working as intended?

RobotJente (commented):
I like the changes and the new test. However I have some comments. If they are resolved then I approve this PR :)

include/roboteam_ai/world_new/Robot.hpp
emielsteerneman: I don't know, it wasn't my decision. It is already this way in the "old" world.

include/roboteam_ai/world_new/views/WorldDataView.hpp
emielsteerneman: I don't know this either. This was already used in the "old" world. I just copied it

src/world_new/Ball.cpp
emielsteerneman: No, this does not fix the bug. This just now returns a RobotView instead of a RobotPtr, as it should

src/world_new/WorldData.cpp
emielsteerneman: The robot has a variable "distanceToBall". It needs the ball to actually set this distance. This is how it's also done in the "old" world

src/world_new/views/WorldDataView.cpp
emielsteerneman: Yes, I wrote tests

src/world_new/WorldData.cpp
RobotJente: Fair enough :+1:

src/world_new/Ball.cpp
RobotJente: Wait so the extra dereference means it becomes a view?

include/roboteam_ai/world_new/views/WorldDataView.hpp
RobotJente: I will adjust it

include/roboteam_ai/world_new/Robot.hpp
RobotJente: Strange

include/roboteam_ai/world_new/views/WorldDataView.hpp
RobotJente: I think this should not be MAX_BALL_BOUNCE_RANGE, but rather MAX_BALL_RANGE, no?

emielsteerneman: I'm not sure, that might be the case. It kind of depends on why you want to call the function I guess, which is why you can give your own minimum distance.

src/world_new/Ball.cpp
emielsteerneman: world->whichRobotHasBall() returns an std::optional<RobotView>, because there might not be a robot that has the ball. (*robot) then unpacks that std::optional, which leaves you with a RobotView.

include/roboteam_ai/world_new/views/WorldDataView.hpp
RobotJente: I pushed a commit where I changed the default to MAX_BALL_RANGE so I guess its fine

src/world_new/Ball.cpp
RobotJente: Ahh okay. I didn't know that

RobotJente (approved):

PR #820, opened 2020-02-27
Refactor/skills new world

timovdk:
Refactor of skills, conditions and controllers to use the new world


mrlukasbos:
@timovdk I see your struggle. Maybe it is a good idea to at some point use the global world temporarily to keep your PR small

include/roboteam_ai/control/NewControlUtils.h
```diff
@@ -0,0 +1,105 @@
...
+#include "world_new/FieldComputations.hpp"
+
+using Vector2 = rtt::Vector2;
+using Angle = rtt::Angle;
+
+namespace rtt::ai::control {
+
+    using namespace rtt::ai::world;
+
+    class NewControlUtils {
```
HaicoDorenbos, line 22: We have decided that it is very important to document all code that you add (with Javadoc), which means that at least:
- Every class should have a short summary of what it does and includes the tag "author" and "since".
- Every public function should have a short summary of what it does, what the meaning is of all parameters and what it returns (if it returns something) and which exceptions it might throw (if the function throws an exception).

HaicoDorenbos (changes requested):
Documentation is missing :disappointed:

PR #821, opened 2020-02-27
Feature/standardized prints

### LINKED ISSUE #819: Print function
I notice that we have a lot of print statements in AI. Some error messages are hard to debug. I propose to make a print() function that helps you print messages in a nice way. then you can see where the print actually comes from, AND you can easily set compiler flags to silence certain types of logs. ( i know these exist for cerr, but having an abstraction here is just nice ;-) ) 

Something like this:

```
void print(std::string className, std::string functionName, std::string message, bool isError) { }
```

mrlukasbos:
needs #802  (merged) 
needs https://github.com/RoboTeamTwente/roboteam_utils/pull/47 (merged)
needs https://github.com/RoboTeamTwente/roboteam_utils/pull/48

Use the new print utilities from roboteam_utils in the startup phase and main loop of AI. I tried to reduce the log spamming a bit, and only keep the relevant details. 

I took the liberty of adding a cool logo at startup ;-) 

Screenshot 2020-03-01 at 12 31 55


RobotJente (approved):

RobotJente (dismissed):
The library is not known in some places, it can be fixed by including it in those files

emielsteerneman:
What is the performance hit for this logging?

mrlukasbos:
did not test. after compiling it is just a function call to Printer::print() which does some std::cout logs. So I expect approximately 0 performance hit

src/utilities/Constants.cpp
```diff
@@ -14,24 +15,21 @@ bool Constants::isInitialized = false;
...
+    rtt_info("robot_output_target is set to " + target)
     isInitialized = true;
 }
 
 bool Constants::GRSIM() {
     if (!isInitialized) {
-        std::cerr << "Ros::init() was not called yet, but you use a value that depends on a ROS parameter. "
-                  << "\n this may result in unexepected behaviour" << std::endl;
+        rtt_error("You use a value dependent on an unkown environment! This may result in unexepected behaviour")
+        assert(false);
```
emielsteerneman, line 26: Why is this added?

src/interface/widgets/SettingsWidget.cpp
```diff
@@ -27,7 +27,6 @@ SettingsWidget::SettingsWidget(QWidget *parent) {
     grsimPort->setValue(SETTINGS.getRobothubSendPort());
     grsimSettingsWidgetLayout->addWidget(grsimPort);
     grsimSettingsGroup->setLayout(grsimSettingsWidgetLayout);
-    QObject::connect(grsimPort, SIGNAL(textChanged(QString)), this, SLOT(changeGrSimPort(int)));
```
emielsteerneman, line 30: Does this mean we cannot change the grSim port anymore?

src/bt/Node.cpp
```diff
@@ -79,7 +80,10 @@ void Node::NodeTerminate(Status s) {
 
 unsigned long long Node::getAmountOfTicks() const { return amountOfTicks; }
 
-void Node::giveProperty(std::string a, std::string b) { std::cerr << "giveProperty in Node.cpp, should never be called" << std::endl; }
+void Node::giveProperty(std::string a, std::string b) {
+    rtt_error("giveProperty in Node.cpp, should never be called");
+    assert(false);
```
emielsteerneman: Would this not crash our entire code?

src/interface/widgets/SettingsWidget.cpp
mrlukasbos: That feature never worked anyway -> QT gave an error that the Signal did not exist

src/bt/Node.cpp
mrlukasbos: No it should not but, but I know a cleaner solution: making this purely virtual.

src/utilities/Constants.cpp
mrlukasbos: well if you get this error you have made a serious mistake somewhere, which can be invisible when testing with grsim but can affect real play. This makes sure you crash rightfully.

emielsteerneman (approved):

src/ApplicationManager.cpp
```diff
@@ -46,17 +49,23 @@ void ApplicationManager::start() {
 /// Run everything with regard to behaviour trees
 void ApplicationManager::runOneLoopCycle() {
     if (io::io.hasReceivedGeom) {
+        if (!fieldInitialized) rtt_success("Received a field message!")
```
Yuhanun:
```suggestion
        if (!fieldInitialized) rtt_success("Received first field message!")
```

src/ApplicationManager.cpp
```diff
@@ -46,17 +49,23 @@ void ApplicationManager::start() {
 /// Run everything with regard to behaviour trees
 void ApplicationManager::runOneLoopCycle() {
     if (io::io.hasReceivedGeom) {
```
Yuhanun, line 51: shouldn't this be an atomic boolean?

src/ApplicationManager.cpp
```diff
@@ -46,17 +49,23 @@ void ApplicationManager::start() {
...
         if (!SETTINGS.isLeft()) {
-         //   roboteam_utils::rotate(&fieldMessage);
             roboteam_utils::rotate(&worldMessage);
         }
 
         world->updateWorld(fieldMessage, worldMessage); // this one needs to be removed
 
         if (!world->getUs().empty()) {
+            if (!robotsInitialized) {
+                rtt_success("Received robots, starting behaviour trees!")
```
Yuhanun:
```suggestion
                rtt_success("Received first robots, starting behaviour trees!")
```

src/ApplicationManager.cpp
```diff
@@ -95,11 +110,13 @@ void ApplicationManager::updateTrees() {
     bool keeperStrategyChanged = oldKeeperTreeName != keeperTreeName;
 
     if (strategyChanged) {
+        rtt_info("Switching main strategy to " + strategyName);
         BTFactory::setCurrentTree(strategyName);
         oldStrategyName = strategyName;
     }
 
     if (keeperStrategyChanged) {
+        rtt_info("Switching keeper strategy to " + keeperTreeName)
```
Yuhanun: could you not make this variadic to prevent a runtime allocation?

Such as 

```
rtt_info("Switching keeper strategy to", keeperTreeName)
```

Yuhanun: Also, maybe change the macro to require a semicolon

Yuhanun (approved):

src/ApplicationManager.cpp
mrlukasbos: No. Sometimes we lose all robots out of the field and then you can get this message again when they re appear :-)

Yuhanun: Oh okay, nevermind then <3

Issue #791, opened 2020-02-10
NumTrees not working


Ratoone:
After the structure refactoring, the path is not constructed properly - the end position is not even part of the path. Consider rewriting the algorithm or implementing a different one

rolfvdhulst:
This surely must have been broken in one of the refactors, no?

Also, I believe sometimes that if the path is too long we only return part of the path towards it and recompute it later, but I am not 100% sure about that. It might be a feature.

Ratoone:
Yeah, but this time it computes only 2 points which are really close to the original robot (as in on top of it), and that's it. Probably I removed some workaround or something

Ratoone:
Solved with #822

Issue #777, opened 2020-01-30
PathPoint refactoring


Ratoone:
Right now the NumTrees implementation uses shared pointers. As the changes would affect the old structure as well, this will not be performed in #765, but will follow #776

_Originally posted by @rolfvdhulst in https://github.com/RoboTeamTwente/roboteam_ai/pull/765_

Ratoone:
Obsolete due to the new stp

PR #822, opened 2020-02-27
Rewrote Num Trees Planning

Ratoone:
Rewrote the current path planning algorithm. Added a new class as an alternative to the old PathPoint, that doesn't contain shared pointers. This PR also solves #777 and #791. Also, the position control  doesn't rely on the robots anymore (therefore of the world as well), but directly on their position. This update is done at every update of the world.


CMakeLists.txt
```diff
@@ -246,7 +247,7 @@ set(TEST_SOURCES
         ${PROJECT_SOURCE_DIR}/test/UtilTests/RefereeTest.cpp
         ${PROJECT_SOURCE_DIR}/test/UtilTests/RobotDealerTest.cpp
         ${PROJECT_SOURCE_DIR}/test/UtilTests/StrategyManagerTest.cpp
-        ${PROJECT_SOURCE_DIR}/test/helpers/WorldHelper.h
+        ${PRtOJECT_SOURCE_DIR}/test/helpers/WorldHelper.h
```
emielsteerneman: Typo here

include/roboteam_ai/control/positionControl/CollisionDetector.h
```diff
@@ -17,9 +17,9 @@ namespace rtt::ai::control {
  */
 class CollisionDetector {
    private:
-    const double DEFAULT_ROBOT_COLLISION_RADIUS = 3.0 * Constants::ROBOT_RADIUS();
+    const double DEFAULT_ROBOT_COLLISION_RADIUS = 2.0 * Constants::ROBOT_RADIUS();
```
emielsteerneman: This is technically correct, but really tight. We might get yellow cards if the autoref sees this as "too close" due to noise or something

src/control/positionControl/CollisionDetector.cpp
```diff
@@ -9,7 +9,7 @@ namespace rtt::ai::control {
 bool CollisionDetector::isCollisionBetweenPoints(const Vector2& initialPoint, const Vector2& nextPoint) {
     bool isFieldColliding = field ? !isPointInsideField(nextPoint) || isPointInDefenseArea(nextPoint) : false;
```
emielsteerneman, line 11: Returning isFieldColliding when it is true, might give a performance boost since getRobotCollisionBetweenPoints doesn't have to be called anymore

emielsteerneman: Also, is it possible to disable isPointInDefenseArea? e.g. for the keeper

src/control/positionControl/PositionControl.cpp
```diff
@@ -30,10 +27,12 @@ RobotCommand PositionControl::computeAndTrackPath(const world::Field &field, int
     return command;
 }
 
-bool PositionControl::shouldRecalculatePath(const Vector2 &currentPosition, const Vector2 &targetPos, int robotId) {
+bool PositionControl::shouldRecalculatePath(const Vector2 &currentPosition, const Vector2 &targetPos,
+                                            const Vector2 &currentVelocity, int robotId) {
     return computedPaths[robotId].empty() || PositionControlUtils::isTargetChanged(targetPos, computedPaths[robotId].back()) ||
-           collisionDetector.isRobotCollisionBetweenPoints(currentPosition, computedPaths[robotId].front());
+            (currentVelocity != Vector2() && collisionDetector.getRobotCollisionBetweenPoints(currentPosition, computedPaths[robotId].front()));
```
emielsteerneman: `currentVelocity != Vector2()` Not sure what this does. Does it check if the values are (0, 0)? If so, it might be better to check if the length of the vector is very small. The currentVelocity might suffer from noise.

src/world_new/World.cpp
```diff
@@ -59,6 +59,10 @@ std::optional<view::WorldDataView> World::getHistoryWorld(size_t ticksAgo) const
 void World::updateWorld(proto::World &protoWorld) {
     WorldData data{protoWorld, *settings, updateMap};
     setWorld(data);
+
+    std::vector<Vector2> robotPositions(getWorld()->getRobotsNonOwning().size());
+    std::transform(getWorld()->getRobotsNonOwning().begin(), getWorld()->getRobotsNonOwning().end(), robotPositions.begin(), [](const auto& robot) -> Vector2 { return (robot->getPos()); });
+    positionControl.setRobotPositions(robotPositions);
```
emielsteerneman: I don't think this should be in World. I think it would fit better in ApplicationManager, just before ticking the behaviour trees

include/roboteam_ai/control/positionControl/PositionControl.h
```diff
@@ -42,20 +42,21 @@ class PositionControl {
...
     /**
      * Updates the robot view vector
-     * @param robots the RobotView vector of robots
+     * @param robotPositions the RobotView vector of robots
      */
-    void setRobotVector(const std::vector<world_new::view::RobotView> &robots);
+    void setRobotPositions(std::vector<Vector2> &robotPositions);
 
     /**
      * The computed path should be recalculated if: <br>
```
emielsteerneman, line 49: Are these html tags?

emielsteerneman: In what case would the robot not be moving over its path?

include/roboteam_ai/control/positionControl/PositionControl.h
```diff
@@ -42,20 +42,21 @@ class PositionControl {
 
     /**
      * Updates the robot view vector
-     * @param robots the RobotView vector of robots
+     * @param robotPositions the RobotView vector of robots
```
emielsteerneman: "The position of all the robots"

include/roboteam_ai/control/positionControl/pathTracking/NumTreesTracking.h
```diff
@@ -18,7 +18,7 @@ namespace rtt::ai::control {
  */
 class NumTreesTracking {
    private:
-    static constexpr unsigned long STEPS_AHEAD = 4;
+    static constexpr unsigned long STEPS_AHEAD = 1;
```
emielsteerneman, line 22: What does this change mean?

src/control/positionControl/CollisionDetector.cpp
```diff
@@ -18,26 +18,24 @@ bool CollisionDetector::isPointInDefenseArea(const Vector2& point) {
     return FieldComputations::pointIsInDefenceArea(*field, point, true) || FieldComputations::pointIsInDefenceArea(*field, point, false);
 }
 
-bool CollisionDetector::isRobotCollisionBetweenPoints(const Vector2& initialPoint, const Vector2& nextPoint) {
-    for (const auto& robot : *robots) {
+std::optional<Vector2> CollisionDetector::getRobotCollisionBetweenPoints(const Vector2& initialPoint, const Vector2& nextPoint) {
```
emielsteerneman, line 40: Maybe there should be a check here that returns no collision if the initialPoint and nextPoint are really close to eachother (e.g. 0.1)

emielsteerneman (dismissed):
Code looks good. Tested it on GrSim, and seem to work. However, I think I've seen some bugs. It looks like the path is recalculated every time the robot is at the next node in the path. I think this is because the robot is detecting a collision with itself, triggering a recalculation (even though that seems taken care of). 

I'm able to generate paths where the robots gets stuck behind other robots.

Sometimes, the paths aren't recalculated, even when the robot is currently colliding with another robot.

Is it possible to avoid the defense area?

src/control/positionControl/PositionControl.cpp
Ratoone: That's the thing, because of machine errors the length is never zero. However, when comparing to a different vector, like (0, 0), it takes the error into account. I will modify to Vector2(0, 0) for clarity

CMakeLists.txt
Ratoone: Oops, my bad

include/roboteam_ai/control/positionControl/pathTracking/NumTreesTracking.h
Ratoone: When following a dense path, you will want to follow a point ahead, for a faster path following. However, the algorithm generates a sparse path, so it should follow the first point in the path

include/roboteam_ai/control/positionControl/CollisionDetector.h
Yuhanun:
```suggestion
    constexpr static double DEFAULT_ROBOT_COLLISION_RADIUS = 2.0 * Constants::ROBOT_RADIUS();
```

include/roboteam_ai/control/positionControl/CollisionDetector.h
```diff
@@ -17,9 +17,9 @@ namespace rtt::ai::control {
...
 
-    const std::vector<world_new::view::RobotView>* robots = nullptr;
+    std::vector<Vector2> robotPositions;
     const world::Field* field = nullptr;
```
Yuhanun, line 23: remove the const qualifier

include/roboteam_ai/control/positionControl/PathPointNode.h
```diff
@@ -0,0 +1,40 @@
...
+#include <roboteam_utils/Vector2.h>
+
+namespace rtt::ai::control{
+/**
+ * A node in the path tree generation. Has a position and a parent
+ */
+class PathPointNode {
+private:
+    Vector2 position;
+    PathPointNode *parent = nullptr;
```
Yuhanun, line 17: which is default
```suggestion
    PathPointNode *parent;
```

src/control/positionControl/PathPointNode.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace rtt::ai::control{
+PathPointNode::PathPointNode(const Vector2 &position) : position(position) {}
+
+PathPointNode::PathPointNode(const Vector2 &position, PathPointNode &parent) : position{position}, parent{&parent}{}
+
+const Vector2 &PathPointNode::getPosition() const {
+    return position;
+}
+
```
Yuhanun, line 15: redundant blank line

Yuhanun (dismissed):

include/roboteam_ai/control/positionControl/pathTracking/NumTreesTracking.h
Ratoone: Renamed the class for more clarity

include/roboteam_ai/control/positionControl/PathPointNode.h
Ratoone: UB if I don't initialize it with null

src/control/positionControl/CollisionDetector.cpp
Ratoone: As I understood, the keeper will have a different behavior. You can always just use the functions individually.

> Returning isFieldColliding when it is true, might give a performance boost since getRobotCollisionBetweenPoints doesn't have to be called anymore

Fair, I'll change the order

include/roboteam_ai/control/positionControl/PositionControl.h
Ratoone: Yes, you can add html tags to javadoc. It doesn't recognize newlines.

The path should be recalculated when, for instance, another robot moves in front of the current robot

include/roboteam_ai/control/positionControl/pathTracking/NumTreesTracking.h
Ratoone: Renamed the class for clarity

include/roboteam_ai/control/positionControl/PathPointNode.h
Yuhanun: Nope, it's not UB because you still construct the class itself, meaning it gets default initialized to 0b000000...000

mrlukasbos: Depends on what happens in the implementation. However I would consider init  with nullptr good practice

src/control/positionControl/CollisionDetector.cpp
RobotJente: What if I make a path of points that are 0.1 distance from each other, and make sure the first point is not a collision? I think optimizations like this should be added only if we find out that the algorithm is running too slow because they can introduce bugs.

src/control/positionControl/pathPlanning/NumTreesPlanning.cpp
```diff
@@ -1,194 +1,69 @@
...
-
-    Vector2 targetVel = newPoint->vel + newPoint->acc.stretchToLength(DT * std::max(newPoint->maxAcceleration(), newPoint->maxDeceleration()));
-    Angle targetAngle = (targetVel - oldPoint->vel).toAngle();
-
-    // ODE model to get new position/velocity:
-    newPoint->vel = ControlUtils::accelerationLimiter(targetVel, oldPoint->vel, targetAngle, newPoint->maxAcceleration() * DT, newPoint->maxAcceleration() * DT,
-                                                      newPoint->maxDeceleration() * DT, newPoint->maxDeceleration() * DT);
-    newPoint->pos += newPoint->vel * DT;
+        auto branches = branchPath(point, collision.value());
+        std::for_each(branches.begin(), branches.end(), [&pointQueue](const PathPointNode& newPoint){pointQueue.push(newPoint);});
```
RobotJente, line 51: I think doing this as a lambda makes it less readable than just a plain old foreach

src/world_new/World.cpp
RobotJente: Sounds like a good suggestion.

CMakeLists.txt
```diff
@@ -218,6 +220,7 @@ set(CONTROL_SOURCES
         ${PROJECT_SOURCE_DIR}/src/control/ball-handling/RotateAroundBall.cpp
         ${PROJECT_SOURCE_DIR}/src/control/ball-handling/RotateWithBall.cpp
         ${PROJECT_SOURCE_DIR}/src/control/positionControl/PositionControlUtils.cpp
+
```
rttwente:
```suggestion
```

include/roboteam_ai/control/positionControl/PathPointNode.h
rttwente, line 17: Why raw pointer?

include/roboteam_ai/control/positionControl/pathTracking/PidTracking.h
```diff
@@ -44,8 +38,9 @@ class PidTracking {
      * @param pathPoints the path as a list of points
      * @param outputVelocity control velocity that will be outputted to the robot at the current tick
      * @param outputAngle the resulting orientation of the robot at the current tick
+     * @param robotId the robotId of the controlled robot - used only for state based tracking (like PID)
      */
-    Position trackPath(const Vector2 &currentPosition, const Vector2 &currentVelocity, std::vector<Vector2> &pathPoints);
+    Position trackPath(const Vector2 &currentPosition, const Vector2 &currentVelocity, std::vector<Vector2> &pathPoints, int robotId = 0);
```
rttwente: Why default to robotId 0?

Ratoone: Changed

src/world_new/World.cpp
Ratoone: Done

include/roboteam_ai/control/positionControl/PathPointNode.h
Ratoone: It can't be the object itself, as for some reason you can't declare a member of the same class. It can't be a reference, as it should be able to be null

PR #824, opened 2020-03-01
Sensible GUI framerates

mrlukasbos:
During profiling I noticed that the visualizer was not that heavy. In fact, the treewidget updates were. I saw that all ui widgets where updated at 25hz. That is actually really a lot for something that is not that important.  I turned it down to 5hz. The visualization of the robots updated at 5hz, I turned that up to 20. 

Another thing I noticed is that the Treewidget was still using a lot of CPU. Apparantly changing text color is heavy, and for some things we changed the color every tick. I have removed that feature entirely. 

After this PR the CPU load went down drastically for me: (From 90% to 70% during halt strategy and from 160% to 130% during normal play).

### DIFF
--- a/include/roboteam_ai/interface/api/Input.h
+++ b/include/roboteam_ai/interface/api/Input.h
@@ -40,7 +40,7 @@ struct Drawing {
 
 class Input {
    public:
-    explicit Input() = default;
+    explicit Input();
     virtual ~Input();
 
     static void clearDrawings();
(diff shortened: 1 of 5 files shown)


Yuhanun:
Honestly, they're fine changes, but I think it's smarter if we don't merge this. Purely for the fact that we are planning to overhaul the way to UI works in general, so it would pressure us more if we don't improve the current system.

However it's fine to approve, but not doing it would pressure us to work on the actual problem, rather than a temporary fix.

Yuhanun (approved):

emielsteerneman (approved):

mrlukasbos:
Well, I suspect the new UI will still take quite some time before it is finished. and, having more efficient codes means more time for path planning

PR #826, opened 2020-03-02
Added copy ctor to worlddata and robot

Yuhanun:
Both are now copyable

Suggestion for the tree:

Copy the worlddata once, set it in the origin, then pass worlddataviews to  that

### DIFF
--- a/include/roboteam_ai/world_new/WorldData.hpp
+++ b/include/roboteam_ai/world_new/WorldData.hpp
@@ -80,10 +80,11 @@ class WorldData {
     WorldData() = default;
 
     /**
-     * Copy assignment operator and constructor, explicitly deleted
+     * Copy assignment operator and constructor
+     * Copy over members and set the view vectors
      */
-    WorldData &operator=(WorldData const &) = delete;
-    WorldData(WorldData const &) = delete;
+    WorldData &operator=(WorldData const &) noexcept;
+    WorldData(WorldData const &) noexcept;
 
     /**
      * Move constructor, simply moves all members
@@ -96,6 +97,11 @@ class WorldData {
      * @return *this
      */
     WorldData &operator=(WorldData &&) = default;
+
+    /**
+     * Does exactly what it says, it sets the internal vectors for the views
+     */
+    void setViewVectors() noexcept;
 
     /**
      * Gets a non-owning container of robots that are in our team
--- a/src/world_new/WorldData.cpp
+++ b/src/world_new/WorldData.cpp
@@ -83,6 +83,5 @@ namespace rtt::world_new {
             auto& matchingVector = each.getTeam() == Team::us ? us : them;
             matchingVector.emplace_back(&each);
         }
-
     }
 }  // namespace rtt::world_new
--- a/src/world_new/WorldData.cpp
+++ b/src/world_new/WorldData.cpp
@@ -64,13 +64,17 @@ namespace rtt::world_new {
         this->robots = other.robots;
         this->ball = other.ball;
 
-        setViewVectors();
+        this->us.reserve(other.getUs().size());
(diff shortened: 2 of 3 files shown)


src/world_new/WorldData.cpp
```diff
@@ -9,58 +9,79 @@
...
+        return us;
     }
-    for (auto &each : others) {
-        auto _ptr = &robots.emplace_back(feedback, each, Team::them, getBall());
-        them.emplace_back(_ptr);
-        robotsNonOwning.emplace_back(_ptr);
+
+    std::vector<view::RobotView> const &WorldData::getThem() const noexcept { return them; }
+
+    std::vector<rtt::world_new::robot::Robot> const &WorldData::getRobots() const noexcept { return robots; }
```
emielsteerneman, line 46: Is it safe to provide a function that gives access to mutable objects?

emielsteerneman (approved):

src/world_new/WorldData.cpp
Yuhanun: Resolved.

include/roboteam_ai/world_new/WorldData.hpp
```diff
@@ -97,6 +98,11 @@ class WorldData {
      */
     WorldData &operator=(WorldData &&) = default;
 
+    /**
+     * Does exactly what it says, it sets the internal vectors for the views
+     */
+    void setViewVectors() noexcept;
```
mrlukasbos, line 104: shouldn't this function be private?

src/world_new/WorldData.cpp
```diff
@@ -9,58 +9,79 @@
...
-uint64_t WorldData::getTime() const noexcept { return time; }
+    WorldData::WorldData(WorldData const & other) noexcept :
+            robots{ other.robots }, ball{ other.ball }{
+        setViewVectors();
+    }
 
-std::vector<view::RobotView> const &WorldData::getRobotsNonOwning() const noexcept { return robotsNonOwning; }
+    void WorldData::setViewVectors() noexcept {
+        /*
+         * Reserve the us and them vectors, making sure they are not being resized (which will invalidate the references)
```
mrlukasbos, line 88: where does this reserve happen?

mrlukasbos (changes requested):
It looks like you don't reserve us and them anymore

src/world_new/WorldData.cpp
Yuhanun: Oh yeah, nice catch, ill take a look at that tomorrow

Yuhanun:
Resolved the issue where iterators would be invalidated.

include/roboteam_ai/world_new/WorldData.hpp
Yuhanun: Doesn't really matter, it doesn't fuck with the actual robots so yeah, but I'll make it private.

src/world_new/WorldData.cpp
```diff
@@ -9,58 +9,83 @@
...
+        robots.reserve(amountUs + amountThem);
+
+        for (auto &each : ours) {
+            robots.emplace_back(feedback, each, Team::us, getBall());
+        }
+        for (auto &each : others) {
+            robots.emplace_back(feedback, each, Team::them, getBall());
+        }
+
+
```
mrlukasbos:
```suggestion
us.reserve(amountUs);
them.reserve(amountThem);
```

mrlukasbos (changes requested):
Im not sure but you still may have UB in the us and them vectors. When constructing a worldData from a protoMessage. The setViewVector then fills us and them with robots from the 'robots' vector.

Yuhanun:
is that what you meant?

PR #828, opened 2020-03-02
Feature/world reset

Yuhanun:
resetting the world + tests


CMakeLists.txt
```diff
@@ -256,7 +256,7 @@ set(TEST_SOURCES
         ${PROJECT_SOURCE_DIR}/test/World_newTests/HistorySizeTest.cpp
         ${PROJECT_SOURCE_DIR}/test/World_newTests/HistoryRetrievalTest.cpp
         ${PROJECT_SOURCE_DIR}/test/World_newTests/WhichRobotHasBallTest.cpp
-        )
+        test/World_newTests/WorldResetTests.cpp)
```
emielsteerneman: ${PROJECT_SOURCE_DIR}

Yuhanun: done

include/roboteam_ai/world_new/World.hpp
```diff
@@ -37,10 +38,19 @@ class World {
...
+     * @param[in] resetWorld Boolean that marks whether to reset the world before returning
+     * if set to true, world.clear() is called
+     * if set to false, worldInstance is simply reset
      * @return A pointer to a static World
      */
-    inline static World *instance() {
+    inline static World *instance(bool resetWorld = false) {
         static World worldInstance{&rtt::SETTINGS};
+        if (resetWorld) {
+#ifndef RUNNING_TEST
```
mrlukasbos: if we are using macro's anyways, why not check if there is a macro that indicates a google test environment and completely enable/disable the function with this macro? :-)

Yuhanun: true, will add

RobotJente (approved):
Nice work!

PR #835, opened 2020-03-04
Redesign/Simplify/Document FieldComputation class

HaicoDorenbos:
Summarized: improve design/documentation of the FieldComputation class and simplify the code of the FieldComputation class. Moreover the Field class is better documented

Changes included:
- All margins from now on are all outwards (no combinations of outwards and inwards margins).
- No more extraction of x and y coordinates of a vector do additions on them and recreate the vector, but directly adding a vector to another vector.
- Additional documentation to the Field class which explains some invariants on the Field class (things that should hold for the Field values).
- Added some additional Field vectors in the Field class.
- Large simplification of the code in the getBlockadesMappedToGoal() function and a small simplification of the lineIntersectionWithDefenceArea() function.
- Removed code clones in the FieldComputation class from the getDefenseArea() and getGoalArea(). The opposite area is no longer computed if it is not returned.
- A lot of documentation added to the FieldComputation class.
- Removed useless comments and added comments to non self-explanatory code in the FieldComputation class. 
- Used the NEGLIGIBLE_LENGTH constant for float comparisons (if smaller than NEGLIGIBLE_LENGTH then they are considered equal).


include/roboteam_ai/world/Field.h
```diff
@@ -33,6 +33,20 @@ struct FieldArc {
...
+ * coordinate of the form (x, 0). Another example is that the bottommost y-coordinate of the field is of the form -y where the topmost y-coordinate of the field is of the form y.
+ * - The left side of the field corresponds with the lowest x-coordinate (is a negative value), whereas the right side of the field corresponds with the highest y-coordinate (is a
+ * positive value). The bottom side of the field corresponds with the lowest y-coordinate (is a negative value), whereas the top side of the field corresponds to the highest
+ * y-coordinate (is a positive value).
+ * - Every horizontal line of the field stored as FieldLineSegment has their begin position equal to the leftmost position of that line (position with the lowest x-coordinate) and
+ * their end position equal to the rightmost position of that line (position with highest x-coordinate). Every vertical line of the field stored as FieldLineSegment has their begin
+ * position equal to the bottommost position of that line (position with the lowest y-coordinate) and ther end position equal to the topmost position of that line (position with
+ * the highest y-coordinate).
+ *
```
mrlukasbos, line 53: This is far too much. If you need this amount of documentation either your code is not readable enough, or your documentation is to detailed. (i gues the second). just make nice function names that say what they do and it is fine.

include/roboteam_ai/world/Field.h
```diff
@@ -84,7 +98,7 @@ class Field {
     };
 
    private:
-    std::vector<FieldLineSegment> allFieldLines;
+    std::vector<FieldLineSegment> allFieldLines; // Stores all field lines
```
mrlukasbos: obvious comment

include/roboteam_ai/world/Field.h
```diff
@@ -107,18 +121,32 @@ class Field {
     // The center y-coordinate of the field (the y-coordinate that corresponds with the center of the field)
     std::optional<double> centerY;
 
-    // The leftmost x-coordinate of the field (the x-coordinate closest to our goal)
+    // The leftmost x-coordinate of the field which is the lowest x-coordinate value (is a negative value) and is the x-coordinate closest to our goal.
     std::optional<double> leftmostX;
```
mrlukasbos, line 129: The comment is useless and vague as well. What is the x-coordinate closest to our goal? the x coordinate of what? what is wrong with field->getLeftFieldLine.x? which is more descriptive

include/roboteam_ai/world/Field.h
```diff
@@ -107,18 +121,32 @@ class Field {
...
 
+    // The x-coordinate of the left penalty line (the penalty line closest to our goal).
+    std::optional<double> leftPenaltyX;
+
+    // The x-coordinate of the right penalty line (the penalty line closest to the opponents goal).
+    std::optional<double> rightPenaltyX;
+
+    /* The highest y-coordinate of the penalty lines (since both penalty lines have the same y-coordinates
+     * at top and bottom, we do not need a seperate variable for the left and right penalty line) */
+    std::optional<double> penaltyTopY;
```
mrlukasbos, line 148: This variable is usually unreliable, because in ssl_vision it is often not set in a nice way. What is this different with the upperFieldLine.y anyway?

include/roboteam_ai/world/Field.h
```diff
@@ -185,6 +213,34 @@ class Field {
     // The top most point of the opponents goal (this point is on the right line).
     std::optional<Vector2> theirTopGoalSide;
 
+    // The top position, point with the highest y-coordinate, of the left penalty line (our penalty line).
+    std::optional<Vector2> leftPenaltyLineTop;
```
mrlukasbos, line 221: I don't really like that all these variables are stored separately while they are often unused. These are all copies whenever you get a new field message. Why not make your getters return references to the original proto message? This is all extra work and extra code.

mrlukasbos (changes requested):
- There is far too much documentation at some places. It makes classes super long and it makes code less readable. Shorten your comments and improve the function names instead. Developers are not that stupid that you should explain every little detail to them.

- Standardizing the margins is a good idea. But I remember we had a LOT of nasty problems with them last year. Especially problems with margins are basically impossible to see (except when there is an autoref). Because of that we made a lot of tests for the field. Including tests for field margins, pointIsInField, pointIsInDefenceArea, but also the getVisiblePartsOfGoal and Blockade functions. I would like to see those tests back.

include/roboteam_ai/world/Field.h
HaicoDorenbos: These invariants are really needed. I did not understand these things at first sight, so I added them because if I do not understand them then others are also more likely to not understand them either. And a lot of functions outside the Field class do computations based on these assumptions.

include/roboteam_ai/world/Field.h
HaicoDorenbos: The top of the penalty line is different from the top of the upper field line. The upper field line has a much higher y coordinate, whereas the top of the penalty line is the top of the defence areas.

include/roboteam_ai/world/Field.h
HaicoDorenbos: Well it is better to be consistent in your design/approach. All computations are done during construction of the Field class and all these computed values are stored. You cannot be inconsistent and store halve of the values whereas the other halve of the values is computed when the getter is executed.

include/roboteam_ai/world/Field.h
HaicoDorenbos: True...

mrlukasbos:
This is a good read on documentation:

https://medium.com/@andrewgoldis/how-to-document-source-code-responsibly-2b2f303aa525

include/roboteam_ai/world/Field.h
HaicoDorenbos: No, field->getLeftFieldLine.x is less readable than getting the leftMostX of the field. What I often see happening in the AI is that a vector is asked when only part of the vector is needed and then complicated things are done with that vector to return a part of the vector.

include/roboteam_ai/world/Field.h
mrlukasbos: It is, but it isn't. as i said this variable is unreliable: it is often not set properly in sslvision (don't be surprised if this value represents 999 meters).

include/roboteam_ai/world/Field.h
mrlukasbos: I see. It is indeed basic knowledge, which if you don't know it then you cannot work on this codebase. However, this can be more easily explained with some kind of drawing then. for example something like this, but bigger:

```
(-x,y)_________________________ (x,y)
|___                              |
|    |                            |
|    |           (0,0)            | (their side)
| __ |                            | (rightFieldLine)
|                                 |
(-x, -y)_______________________(x, -y)
              (bottomFieldLine)
```

HaicoDorenbos:
> This is a good read on documentation:
> 
> https://medium.com/@andrewgoldis/how-to-document-source-code-responsibly-2b2f303aa525

Regarding documentation I apply the following guidelines:
- If you pick a random Software Developer from the street and you do not explain anything about the code then he/she should understand the code using only the code, documentation and comments.
- Everything that can be misunderstood/misinterpreted should be documented.

I indeed agree that unnecessary documentation and comments should be avoided, but in my opinion most of these comments/documentation are needed. Moreover it is easier to remove documentation than to recover missing documentation.

include/roboteam_ai/world/Field.h
mrlukasbos: or make it shorter: 

About coordinates:
The center of the field is (0,0), therefore the left bottom corner is -x,-y and the top right corner is x,y. In roboteam_ai we our side is always the left side of the field. If we play on the right side in real-life the incoming field messages are rotated when they are received.

HaicoDorenbos: Well I like the drawing, but the invariant for the FieldLine remains, because it is not explained by this drawing.

include/roboteam_ai/world/Field.h
HaicoDorenbos: Well StopFormation::getStopPositions() is the only function that uses it. First it did some complicated stuff which always returned the top y coordinate of the defence area, but that behaviour should be left unchanged. To cover this problem I can add an additional commentswhich warns future software developers about the dangers of using this value.

mrlukasbos: The penalty line is totally different from the defense area! 
*EDIT it is not different. I thought it was, but I just checked and it appears to be correct.

HaicoDorenbos: Okay that means that I fixed all your things, so could you approve it?

mrlukasbos:
No. There are still no tests, and considering what you changed those tests are really necessary. 

Besides I still think there are far too much comments. They annoy me a lot, since I always search on function name. And I believe if you need so much comments than there is a problem with your naming or structure. But that is obviously very subjective. I'd like to hear more opinions on this. I do believe that the kind of documentation you make belongs in a readme or a wiki, but not in the code. 

Concluding: fix tests and i'll let the others decide on the comments.

include/roboteam_ai/world/Field.h
mrlukasbos: but what X? field->getleftfieldline.x clearly shows that it is the xposition of the left field line. in your case, what does it mean? you say it is the leftmost x. But you can go further than the fieldline, which is more left. This is a confusing name and confusing documentation.

include/roboteam_ai/world/Field.h
mrlukasbos: If you abstract it away under getters it is fine, you hide the (more efficient) solution anyway.

include/roboteam_ai/world/Field.h
```diff
@@ -185,6 +217,34 @@ class Field {
...
+    std::optional<Vector2> leftPenaltyLineBottom;
+
+    // The top position, point with the highest y-coordinate, of the right penalty line (their penalty line).
+    std::optional<Vector2> rightPenaltyLineTop;
+
+    // The bottom position, point with the lowest y-coordinate, of the right penalty line (their penalty line).
+    std::optional<Vector2> rightPenaltyLineBottom;
+
+    /* The bottom left corner of the field, which is the point on the field with the lowest x-coordinate (is a negative value) and lowest y-coordinate (is a negative value) and is
+     * located at our side of the field. */
```
mrlukasbos, line 233: obvious comment

include/roboteam_ai/world/Field.h
```diff
@@ -185,6 +217,34 @@ class Field {
...
+    std::optional<Vector2> leftPenaltyLineTop;
+
+    // The bottom position, point with the lowest y-coordinate, of the left penalty line (our penalty line).
+    std::optional<Vector2> leftPenaltyLineBottom;
+
+    // The top position, point with the highest y-coordinate, of the right penalty line (their penalty line).
+    std::optional<Vector2> rightPenaltyLineTop;
+
+    // The bottom position, point with the lowest y-coordinate, of the right penalty line (their penalty line).
+    std::optional<Vector2> rightPenaltyLineBottom;
```
mrlukasbos, line 230: obvious comment

include/roboteam_ai/world/Field.h
```diff
@@ -185,6 +217,34 @@ class Field {
...
     std::optional<Vector2> theirTopGoalSide;
 
+    // The top position, point with the highest y-coordinate, of the left penalty line (our penalty line).
+    std::optional<Vector2> leftPenaltyLineTop;
+
+    // The bottom position, point with the lowest y-coordinate, of the left penalty line (our penalty line).
+    std::optional<Vector2> leftPenaltyLineBottom;
+
+    // The top position, point with the highest y-coordinate, of the right penalty line (their penalty line).
+    std::optional<Vector2> rightPenaltyLineTop;
```
mrlukasbos, line 227: obvious comment

include/roboteam_ai/world/Field.h
```diff
@@ -185,6 +217,34 @@ class Field {
...
+    // The bottom position, point with the lowest y-coordinate, of the right penalty line (their penalty line).
+    std::optional<Vector2> rightPenaltyLineBottom;
+
+    /* The bottom left corner of the field, which is the point on the field with the lowest x-coordinate (is a negative value) and lowest y-coordinate (is a negative value) and is
+     * located at our side of the field. */
+    std::optional<Vector2> bottomLeftCorner;
+
+    /* The top left corner of the field, which is the point on the field with the lowest x-coordinate (is a negative value) and highest y-coordinate (is a positive value) and is
+    located at our side of the field. */
+    std::optional<Vector2> topLeftCorner;
```
mrlukasbos, line 238: obvious comment

include/roboteam_ai/world/Field.h
```diff
@@ -185,6 +217,34 @@ class Field {
...
+
+    /* The bottom right corner of the field, which is the point on the field with the highest x-coordinate (is a positive value) and lowest y-coordinate (is a negative value) and
+     * is located at the opponents side of the field. */
+    std::optional<Vector2> bottomRightCorner;
```
mrlukasbos, line 242: obvious comment

include/roboteam_ai/world/Field.h
```diff
@@ -185,6 +217,34 @@ class Field {
...
+
+    /* The top right corner of the field, which is the point on the field with the highest x-coordinate (is a positive value) and highest y-coordinate (is a positive value) and is
+     * located at the opponents side of the field. */
+    std::optional<Vector2> topRightCorner;
```
mrlukasbos, line 246: obvious comment

include/roboteam_ai/world/FieldComputations.h
```diff
@@ -23,25 +21,214 @@ namespace rtt::ai {
...
+         * that the rectangular defence area will be expanded in both y directions and towards the center of the field, a negative value means that the rectangular defence area
+         * will be shrinked in both y directions and away from the goal side of that defence area (if this parameter is not set then the defence area will not be expanded/shrinked)
+         * @param includeOutsideField If set to true then the defence area will also be expanded/shrinked at the goal side (boundary side) of the field, if set to false then it
+         * will not be expanded/shrinked at the goal side (boundary side) of the field (if this parameter is not set then the defence area will not be expanded/shrinked at the goal
+         * side (boundary side) of the field).
+         * @return True if the point is in the given defence area, false if the point is not in the given defence area.
+         */
+        static bool pointIsInDefenceArea(const Field &field, const Vector2& point, bool isOurDefenceArea = true, double margin = 0.0, bool includeOutsideField = false);
```
mrlukasbos: it is so obvious what this function does. if you rename margin to "outsidemargin" you are basically there.

mrlukasbos (changes requested):
Please do not resolve comments when you didnt change code. I give my opinion to help out and by resolving the comments  you make it looks like they are fixed while you actually just hid them away. 
I annoted some more comments I think are too much.

I also noticed something else. Github hid away the 234 lines long fieldcomputations.h, so i missed it this morning. The amount of comments in there is just factor 5 too much. for things like 'pointisinfield(field, point)' you really don't have to explain what the parameters mean right?

include/roboteam_ai/world/Field.h
HaicoDorenbos: Obvious comment indeed if you look at the documentation sketch, but less obvious if you do not have this information. In the past I created games where the top position had the lowest y-coordinate which is also a design choice.

include/roboteam_ai/world/Field.h
HaicoDorenbos: Obvious comment indeed if you look at the documentation sketch, but less obvious if you do not have this information. In the past I created games where the bottom position had the highest y-coordinate which is also a possible design choice. Since the comment is only 2 lines I guess it hurts more to remove it than to keep it.

include/roboteam_ai/world/FieldComputations.h
HaicoDorenbos: Well according to the rules all public functions need documentation where all parameters and return statements need to be documented. And I am consistent in this principle. I agree that it is indeed quite clear what it does and therefore I keep it short. The outside margin is indeed something that should be mentioned, because it might confuse developers.

include/roboteam_ai/world/Field.h
```diff
@@ -33,6 +33,24 @@ struct FieldArc {
...
+ *
+ *                          (upperFieldLine)
+ *              (-x,y)_________________________ (x,y)
+ *                 |                              |
+ *                 |                              |
+ * (your side)     |             (0,0)            | (their side)
+ * (leftFieldLine) |                              | (rightFieldLine)
+ *                 |                              |
+ *              (-x, -y)_______________________(x, -y)
+ *                          (bottomFieldLine)
```
Yuhanun, line 46: Usually -x, y is 0,0 instead.

include/roboteam_ai/world/Field.h
Yuhanun: I think what lukas is trying to say that it's okay to state: "The bottom left corner of the field", but the ```which is the point on the field with the lowest x-coordinate (is a negative value) and lowest y-coordinate (is a negative value) and is
     * located at our side of the field```

Is redundant.

include/roboteam_ai/world/Field.h
```diff
@@ -211,6 +271,10 @@ class Field {
     double getRightmostX() const;
     double getBottommostY() const;
     double getTopmostY() const;
+    double getLeftPenaltyX() const;
+    double getRightPenaltyX() const;
+    double getPenaltyTopY() const;
+    double getPenaltyBottomY() const;
```
Yuhanun, line 289: These functions arent' documented properly.

You're storing an std::optional<double> but you're returning the inner data itself, what if std::optional is std::nullopt. 

These functions should be documented, as it could result in unwanted behavior (such as value calls on empty std::optionals) when not used properly

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -202,7 +200,7 @@ std::shared_ptr<Vector2> DefencePositionCoach::blockOnDefenseAreaLine(const Fiel
     return nullptr;
 }
 std::vector<Line> DefencePositionCoach::simulatedVisibleGoalParts(const Field &field, const PossiblePass &pass) const {// create a vector with all robots
-    auto visibleParts = FieldComputations::getVisiblePartsOfGoalByObstacles(field, true, pass.endPos, simulatedRobotsUs);
+    auto visibleParts = FieldComputations::getVisiblePartsOfGoal(field, true, pass.endPos, simulatedRobotsUs);
     return visibleParts;
```
Yuhanun:
```suggestion
    return FieldComputations::getVisiblePartsOfGoal(field, true, pass.endPos, simulatedRobotsUs);
```

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -241,10 +239,10 @@ Vector2 DefencePositionCoach::findPositionForBlockBall(const Field &field, const
...
     if (position.x > maxForwardLineX) {
-        double fieldWidth = field.getFieldWidth();
-        Vector2 bottomLine(maxForwardLineX, -fieldWidth * 0.5);
-        Vector2 topLine(maxForwardLineX, fieldWidth * 0.5);
-        Vector2 intersect = control::ControlUtils::twoLineIntersection(blockLine.start, blockLine.end, bottomLine, topLine);
+        Vector2 bottomLine(maxForwardLineX, field.getBottommostY());
+        Vector2 topLine(maxForwardLineX, field.getTopmostY());
+        Vector2 intersect = control::ControlUtils::twoLineIntersection(blockLine.start, blockLine.end, bottomLine,
+                topLine);
         return intersect;
```
Yuhanun:
```suggestion
        Vector2 bottomLine(maxForwardLineX, field.getBottommostY());
        Vector2 topLine(maxForwardLineX, field.getTopmostY());
        return control::ControlUtils::twoLineIntersection(blockLine.start, blockLine.end, bottomLine, topLine);
```

HaicoDorenbos:
> No. There are still no tests, and considering what you changed those tests are really necessary.
> 
> Besides I still think there are far too much comments. They annoy me a lot, since I always search on function name. And I believe if you need so much comments than there is a problem with your naming or structure. But that is obviously very subjective. I'd like to hear more opinions on this. I do believe that the kind of documentation you make belongs in a readme or a wiki, but not in the code.
> 
> Concluding: fix tests and i'll let the others decide on the comments.

I will discuss with Jesse how important tests for the FieldComputation class are and whether it is more important that documenting/simplifying/redesigning the DefencePositionCoach. However if test cases are written then it will be done at a different branch.

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -255,7 +253,7 @@ std::vector<world_new::view::RobotView> DefencePositionCoach::getTheirAttackers(
     std::vector<world_new::view::RobotView> theirAttackers;
     for (auto &robot : them) {
```
Yuhanun: std::copy_if

src/coach/heuristics/OffensiveScore.cpp
```diff
@@ -41,8 +41,8 @@ bool OffensiveScore::positionIsValid(const Field &field, const Vector2 &defaultZ
     }
 
     // check if the point is in the field and out of the defense area
-    if (!FieldComputations::pointIsInField(field, positionToCheck, Constants::ROBOT_RADIUS() * 6) ||
-        FieldComputations::pointIsInDefenceArea(field, positionToCheck, false, Constants::ROBOT_RADIUS() * 2)) {
+    if (!FieldComputations::pointIsInField(field, positionToCheck, -Constants::ROBOT_RADIUS()*6) ||
```
Yuhanun: This entire function always returns false.

src/conditions/BallKickedToOurGoal.cpp
```diff
@@ -18,8 +18,8 @@ bt::Node::Status BallKickedToOurGoal::onUpdate() {
     }
 
     // determine the goalsides
-    Vector2 goalCentre = (*field).getOurGoalCenter();
-    double goalWidth = (*field).getGoalWidth();
+    Vector2 goalCentre = field->getOurGoalCenter();
+    double goalWidth = field->getGoalWidth();
     double margin = BALL_TO_GOAL_MARGIN;
```
Yuhanun: constexpr

src/coach/midfield/MidFieldCoach.cpp
```diff
@@ -234,7 +234,8 @@ Vector2 MidFieldCoach::calculateNewRobotPosition(const Field &field, const Robot
                                       bestPosition + thetaMinus.toVector2(2.0 * GRID_SIZE), bestPosition + thetaMinus.toVector2(4.0 * GRID_SIZE)};
 
     for (const auto &position : positions) {
-        if (!FieldComputations::pointIsInField(field, position, 0.20) || abs(position.x) > DISTANCE_FROM_MIDDLE_LINE) continue;
+        if(!FieldComputations::pointIsInField(field, position, -0.20) ||
+            abs(position.x) > DISTANCE_FROM_MIDDLE_LINE) continue;
         double score = calculateStandingFreeScore(field, position, thisRobot);
         if (score > highestScore) {
```
Yuhanun: std::max

src/control/ControlUtils.cpp
```diff
@@ -326,25 +326,25 @@ Vector2 ControlUtils::getInterceptPointOnLegalPosition(const Field &field, Vecto
...
-                double dist = point.dist(position);
-                if (dist < closestDist) {
-                    closestDist = dist;
-                    closestPoint = point;
+            intersects.insert(intersects.end(), intersectsThem.begin(), intersectsThem.end());
+            if (intersects.empty()) {
+                return projectPos;
+            }
+            double closestDist = 9e9;
+            for (const auto &point: intersects) {
```
Yuhanun, line 337: std::find_if, possibly?

src/world/Field.cpp
```diff
@@ -72,6 +74,16 @@ void Field::initFieldVectors() {
...
+
+    leftPenaltyLineBottom = leftPenaltyLine->begin;
+    leftPenaltyLineTop = leftPenaltyLine->end;
+    rightPenaltyLineBottom = rightPenaltyLine->begin;
+    rightPenaltyLineTop = rightPenaltyLine->end;
+
+    bottomLeftCorner = Vector2(leftmostX.value(), bottommostY.value());
+    topLeftCorner = Vector2(leftmostX.value(), topmostY.value());
+    bottomRightCorner = Vector2(rightmostX.value(), bottommostY.value());
+    topRightCorner = Vector2(rightmostX.value(), topmostY.value());
```
rolfvdhulst, line 89: This needs 100% tests. + There are no safety checks, ssl data can be ordered wrong! This will cause problems down the line.

rolfvdhulst (changes requested):
I agree with Lukas that if you change classes like world, field we should generally aim to fix the tests. Right now, literally your entire PR rests on some assumptions which I *know* are not satisfied at all times by the SSL data. I did not look through the entire PR but it seems it's mostly very minor style changes in a lot of files again.

Documentation wise: I also think it's too much. Generally comments should be there to explain 'why' something works or why you can do it, e.g. provide further insight. The naming should make clear WHAT the object or parameter in question is in 99% of the cases already. Even a random programmer from the street who has learned roughly what SSL does will understand what Vector2 TopRightCorner represents.

Testing is highly important. If you take this approach, I'm not sure it's wise if you touch the DefencePositionCoach, because you 100% need to make better unit tests than the ones already existing now to ensure behavior does not change. The code there is quite complicated and slightly hacky at times.

src/world/Field.cpp
HaicoDorenbos: > I agree with Lukas that if you change classes like world, field we should generally aim to fix the tests. Right now, literally your entire PR rests on some assumptions which I _know_ are not satisfied at all times by the SSL data. I did not look through the entire PR but it seems it's mostly very minor style changes in a lot of files again.
> 
> Documentation wise: I also think it's too much. Generally comments should be there to explain 'why' something works or why you can do it, e.g. provide further insight. The naming should make clear WHAT the object or parameter in question is in 99% of the cases already. Even a random programmer from the street who has learned roughly what SSL does will understand what Vector2 TopRightCorner represents.
> 
> Testing is highly important. If you take this approach, I'm not sure it's wise if you touch the DefencePositionCoach, because you 100% need to make better unit tests than the ones already existing now to ensure behavior does not change. The code there is quite complicated and slightly hacky at times.

I guess if I ask a random software developer on the streets whether the topRightCorner has positive/negative x/y-coordinates then he would not be able to tell it or guess it correctly with 25%. Regarding the tests I agree with you. I did not saw that Travis rejected it, so I will indeed fix the test cases before merging.

src/conditions/HasClearShot.cpp
```diff
@@ -20,15 +20,15 @@ namespace rtt::ai {
 HasClearShot::HasClearShot(std::string name, bt::Blackboard::Ptr blackboard) : Condition(std::move(name), std::move(blackboard)) {}
 
 HasClearShot::Status HasClearShot::onUpdate() {
-    if ((Vector2(ball->getPos()) - (*field).getTheirGoalCenter()).length() < FORCED_SHOOTING_DISTANCE) {
+    if ((Vector2(ball->getPos()) - field->getTheirGoalCenter()).length() < FORCED_SHOOTING_DISTANCE) {
```
Yuhanun: shouldn't x - y already return a Vector2?

src/conditions/HasClearShot.cpp
```diff
@@ -20,15 +20,15 @@ namespace rtt::ai {
...
         return Status::Success;
     }
 
     auto minViewAtGoal = MIN_VIEW_AT_GOAL;
     minViewAtGoal = 0.1;
 
     // return failure if the robot is too far away for a shot at goal
-    if ((Vector2(ball->getPos()) - (*field).getTheirGoalCenter()).length() > MAX_SHOOTING_DISTANCE) {
+    if ((Vector2(ball->getPos()) - field->getTheirGoalCenter()).length() > MAX_SHOOTING_DISTANCE) {
```
Yuhanun: same here

src/interface/widgets/widget.cpp
```diff
@@ -233,7 +233,7 @@ void Visualizer::drawFieldHints(const Field &field, QPainter &painter) {
     pen.setColor(Qt::gray);
     painter.setPen(pen);
 
-    auto lineStart = toScreenPosition(Vector2(field.getOurGoalCenter().x, lineY));
+    auto lineStart = toScreenPosition(Vector2(field.getLeftmostX(), lineY));
     auto lineEnd = toScreenPosition(Vector2(0, lineY));
```
Yuhanun: I'd prefer seeing this:

```suggestion
    auto lineStart = toScreenPosition({ field.getLeftmostX(), lineY });
    auto lineEnd = toScreenPosition({ 0, lineY });
```

src/control/numtrees/NumTreePosControl.cpp
```diff
@@ -354,7 +354,7 @@ Collision NumTreePosControl::getDefenseAreaCollision(const PathPointer &point) {
         bool isInOurDefenseArea = FieldComputations::pointIsInDefenceArea(*field, point->pos, true, margin, false);
         bool isInTheirDefenseArea = FieldComputations::pointIsInDefenceArea(*field, point->pos, false, margin, false);
         if (isInOurDefenseArea || isInTheirDefenseArea) {
-            double defenseAreaX = point->pos.x < 0 ? (*field).getLeftPenaltyLine().begin.x : (*field).getRightPenaltyLine().begin.x;
+            double defenseAreaX = point->pos.x < 0 ? field->getLeftPenaltyX() : field->getRightPenaltyX();
             collision.setDefenseAreaCollision(point->pos, (fabs(defenseAreaX - point->pos.x) + margin) * 1.1);
             return collision;
         }
```
Yuhanun: double logic, I suggest:

```cpp
if (in our defense) {
    defenseArea = field->getLeftPenaltyX();
} else if (in their defense) {
    // ...
}```

File: src/skills/AvoidBall.cpp [hunk 27]
```diff
@@ -51,9 +51,9 @@ bt::Node::Status AvoidBall::onUpdate() {
     force = force + cu::calculateForce(robotPos - ball->getPos(), ballWeight, minBallDistanceForForce);
 
     // forces from walls
-    double boundWidth = (*field).getBoundaryWidth();
-    double halfFieldLength = (*field).getFieldLength() / 2;
-    double halfFieldWidth = (*field).getFieldWidth() / 2;
+    double boundWidth =  field->getBoundaryWidth();
```
Yuhanun:
```suggestion
    double boundWidth = field->getBoundaryWidth();
```

src/skills/DemoAttack.cpp
```diff
@@ -52,9 +52,9 @@ bt::Node::Status DemoAttack::onUpdate() {
     }
     Vector2 velocity;
     if (FieldComputations::pointIsInDefenceArea(*field, robot->pos, ownGoal, 0.0)) {
-        velocity = ((Vector2)robot->pos - (*field).getOurGoalCenter()).stretchToLength(2.0);
+        velocity = ((Vector2)robot->pos - field->getOurGoalCenter()).stretchToLength(2.0);
     } else if (FieldComputations::pointIsInDefenceArea(*field, robot->pos, ownGoal, 0.0)) {
-        velocity = ((Vector2)robot->pos - (*field).getTheirGoalCenter()).stretchToLength(2.0);
+        velocity = ((Vector2)robot->pos - field->getTheirGoalCenter()).stretchToLength(2.0);
```
Yuhanun: robot->pos is already Vector2

src/skills/InterceptBall.cpp
```diff
@@ -234,11 +234,11 @@ bool InterceptBall::ballToGoal() {
...
 // Checks if the ball is in our Goal (e.g. the opponent scored)
 bool InterceptBall::ballInGoal() {
-    Vector2 goalCentre = (*field).getOurGoalCenter();
-    double goalWidth = (*field).getGoalWidth();
+    Vector2 goalCentre = field->getOurGoalCenter();
+    double goalWidth = field->getGoalWidth();
     Vector2 lowerPost = goalCentre + Vector2(0.0, -goalWidth);
     Vector2 upperPost = goalCentre + Vector2(0.0, goalWidth);
-    Vector2 depth = Vector2(-(*field).getGoalDepth(), 0.0);
+    Vector2 depth = Vector2(-field->getGoalDepth(), 0.0);
```
Yuhanun:
```suggestion
    Vector2 depth{ -field->getGoalDepth(), 0.0 };
```

src/skills/Keeper.cpp
```diff
@@ -109,7 +109,7 @@ Vector2 Keeper::computeBlockPoint(const Vector2 &defendPos) {
 void Keeper::setGoalPosWithAttacker(RobotPtr attacker) {
     Vector2 start;
     Vector2 end;
-    double distanceToGoal = ((Vector2)attacker->pos - (*field).getOurGoalCenter()).length();
+    double distanceToGoal = ((Vector2) attacker->pos - field->getOurGoalCenter()).length();
```
Yuhanun:
```suggestion
    double distanceToGoal = (attacker->pos - field->getOurGoalCenter()).length();
```

src/skills/PenaltyKeeper.cpp
```diff
@@ -73,10 +73,10 @@ Vector2 PenaltyKeeper::computeDefendPos() {
 
     if (attacker) {
         Vector2 beginPos = attacker->pos;
```
Yuhanun: why, rip 16 bytes

src/world/FieldComputations.cpp
```diff
@@ -38,126 +35,38 @@ double FieldComputations::getPercentageOfGoalVisibleFromPoint(const Field &field
...
-
-            // the forwardIntersection returns a double which is the scale of the vector projection
-            // this returns -1.0 if there is no intersections in the forward direction
-            double point1val = util::twoLineForwardIntersection(point, lowerSideOfRobot, lowerGoalSide, upperGoalSide);
-            double point2val = util::twoLineForwardIntersection(point, upperSideOfRobot, lowerGoalSide, upperGoalSide);
-            // Here is how we calculate the actual intersections using the above values
-            Vector2 point1 = point + (lowerSideOfRobot - point) * point1val;
-            Vector2 point2 = point + (upperSideOfRobot - point) * point2val;
-            // we can use the
+std::vector<Line> FieldComputations::getVisiblePartsOfGoal(const Field &field, bool ourGoal, const Vector2 &point, world_new::view::WorldDataView &world) {
```
Yuhanun:
```suggestion
std::vector<Line> FieldComputations::getVisiblePartsOfGoal(const Field &field, bool ourGoal, const Vector2 &point, world_new::view::WorldDataView world) {
```

src/world/FieldComputations.cpp
```diff
@@ -185,125 +94,113 @@ std::shared_ptr<Vector2> FieldComputations::lineIntersectionWithDefenceArea(cons
...
-            if (lineStart.dist(intersection) < closestIntersectionToLineStart) {
-                closestIntersection = intersection;
-                closestIntersectionToLineStart = lineStart.dist(intersection);
-            }
-        }
-        return std::make_shared<Vector2>(closestIntersection);
+    } else if (intersections.size() == 2) {
+        double distanceFirstIntersection = lineStart.dist(intersections.at(0));
+        double distanceSecondIntersection = lineStart.dist(intersections.at(1));
+        return std::make_shared<Vector2>(distanceFirstIntersection < distanceSecondIntersection ? intersections.at(0) : intersections.at(1));
```
Yuhanun, line 107: no, no no no no no no nono.

std::optional<Vector2>...

Also, operator[]

src/world/FieldComputations.cpp
```diff
@@ -185,125 +94,113 @@ std::shared_ptr<Vector2> FieldComputations::lineIntersectionWithDefenceArea(cons
 
     if (intersections.size() == 1) {
         return std::make_shared<Vector2>(intersections.at(0));
```
Yuhanun, line 103:
```suggestion
        return intersections[0];
```

src/world/FieldComputations.cpp
Yuhanun, line 107:
```suggestion
        return distanceFirstIntersection < distanceSecondIntersection ? intersections[0] : intersections.at[1];
```

src/world/FieldComputations.cpp
```diff
@@ -185,125 +94,113 @@ std::shared_ptr<Vector2> FieldComputations::lineIntersectionWithDefenceArea(cons
...
-                closestIntersectionToLineStart = lineStart.dist(intersection);
-            }
-        }
-        return std::make_shared<Vector2>(closestIntersection);
+    } else if (intersections.size() == 2) {
+        double distanceFirstIntersection = lineStart.dist(intersections.at(0));
+        double distanceSecondIntersection = lineStart.dist(intersections.at(1));
+        return std::make_shared<Vector2>(distanceFirstIntersection < distanceSecondIntersection ? intersections.at(0) : intersections.at(1));
+    } else {
+        return nullptr;
```
Yuhanun, line 109:
```suggestion
        return std::nullopt;
```

src/world/FieldComputations.cpp
```diff
@@ -185,125 +94,113 @@ std::shared_ptr<Vector2> FieldComputations::lineIntersectionWithDefenceArea(cons
...
-    return ourDefenseArea ? defenceAreaUs : defenceAreaThem;
+    double backLineChanges = includeOutSideField ? field.getBoundaryWidth() : margin;
+    Vector2 bottomGoal = ourDefenseArea ? field.getOurBottomGoalSide() + Vector2(-backLineChanges, -margin) : field.getTheirBottomGoalSide() + Vector2(backLineChanges, -margin);
+    Vector2 topGoal = ourDefenseArea ? field.getOurTopGoalSide() + Vector2(-backLineChanges, margin) : field.getTheirTopGoalSide() + Vector2(backLineChanges, margin);
+    Vector2 bottomPenalty = ourDefenseArea ? field.getLeftPenaltyLineBottom() + Vector2(margin, -margin) : field.getRightPenaltyLineBottom() + Vector2(-margin, -margin);
+    Vector2 topPenalty = ourDefenseArea ? field.getLeftPenaltyLineTop() + Vector2(margin, margin) : field.getRightPenaltyLineTop() + Vector2(-margin, margin);
+
+    std::vector<Vector2> defenseArea = {bottomPenalty, topPenalty, topGoal, bottomGoal};
+    interface::Input::drawDebugData(defenseArea);
+    return Polygon(defenseArea);
```
Yuhanun, line 123:
```suggestion
    return Polygon{ defenseArea };
```

src/world/FieldComputations.cpp
```diff
@@ -185,125 +94,113 @@ std::shared_ptr<Vector2> FieldComputations::lineIntersectionWithDefenceArea(cons
...
+    Vector2 lowerSideOfRobot = point + Vector2(length, 0).rotate((Vector2(robot->getPos()) - point).angle() - theta);
+    Vector2 upperSideOfRobot = point + Vector2(length, 0).rotate((Vector2(robot->getPos()) - point).angle() + theta);
+    Vector2 lowerMapToGoal = goalSide.project(util::twoLineIntersection(point, lowerSideOfRobot, goalSide.start, goalSide.end));
+    Vector2 upperMapToGoal = goalSide.project(util::twoLineIntersection(point, upperSideOfRobot, goalSide.start, goalSide.end));
+
+    LineSegment blockade = LineSegment(lowerMapToGoal, upperMapToGoal);
+    if (blockade.length() > NEGLIGIBLE_LENGTH) {
+        return blockade;
+    } else {
+        return {}; // Ignore blockades 'shadows' that are entirely below/above the goal side.
```
Yuhanun: Consider explicitly returning std::nullopt
```suggestion
        return std::nullopt; // Ignore blockades 'shadows' that are entirely below/above the goal side.
```

src/world/FieldComputations.cpp
```diff
@@ -185,125 +94,113 @@ std::shared_ptr<Vector2> FieldComputations::lineIntersectionWithDefenceArea(cons
...
-/*
- * Get the visible parts of a goal
- * This is the inverse of getting the blockades of a goal
- */
-std::vector<Line> FieldComputations::getVisiblePartsOfGoal(const Field &field, bool ourGoal, const Vector2 &point, world_new::view::WorldDataView &world) {
-    return getVisiblePartsOfGoalByObstacles(field, ourGoal, point, world.getUs());
+std::vector<LineSegment> FieldComputations::mergeBlockades(std::vector<LineSegment> blockades) {
+    /* If two blockades intersect (in this case, overlap), we take the beginning of the first obstacle and the end of the second obstacle, and put them back in the front of the
+     * obstacles vector. The second element gets erased. If they don't intersect, try the next two obstacles. Repeat this procedure until no overlaps are left. */
+    std::sort(blockades.begin(), blockades.end(), [](const LineSegment &a, const LineSegment &b) { return a.start.y < b.start.y; });
```
Yuhanun, line 181:
```suggestion
    std::sort(blockades.begin(), blockades.end(), [](const LineSegment &a, const LineSegment &b) {
        return a.start.y < b.start.y; 
    });
```

src/world/FieldComputations.cpp
```diff
@@ -185,125 +94,113 @@ std::shared_ptr<Vector2> FieldComputations::lineIntersectionWithDefenceArea(cons
...
+    int iterator = 0;
+    while (iterator < static_cast<int>(blockades.size()) - 1) {
```
Yuhanun, line 183:
```suggestion
    size_t index = 0;
    while (index < blockades.size() - 1) {
```

src/world/FieldComputations.cpp
```diff
@@ -185,125 +94,113 @@ std::shared_ptr<Vector2> FieldComputations::lineIntersectionWithDefenceArea(cons
...
+        LineSegment &secondBlockade = blockades.at(iterator + 1);
+        if (firstBlockade.end.y >= secondBlockade.start.y) {
+            // If the first two elements intersects, then merge these blockades into 1 single blockade.
+            auto upperbound = fmax(firstBlockade.end.y, secondBlockade.end.y);
+            auto newBlockade = LineSegment(firstBlockade.start, Vector2(firstBlockade.start.x, upperbound));
+            blockades.erase(blockades.begin() + iterator + 1);
+            blockades.at(iterator) = newBlockade;
+        } else {
+            iterator++;
+        }
```
Yuhanun, line 194:
```suggestion
        LineSegment &firstBlockade = blockades[index];
        LineSegment &secondBlockade = blockades[index + 1];
        if (firstBlockade.end.y >= secondBlockade.start.y) {
            // If the first two elements intersects, then merge these blockades into 1 single blockade.
            auto upperbound = fmax(firstBlockade.end.y, secondBlockade.end.y);
            auto newBlockade = LineSegment(firstBlockade.start, Vector2(firstBlockade.start.x, upperbound));
            blockades.erase(blockades.begin() + index + 1);
            blockades[index] = newBlockade;
        } else {
            index++;
        }
```

include/roboteam_ai/world/FieldComputations.h
Yuhanun: Which rules?

Yuhanun: Issue with anything that's not the C++ standard is that there are no rules

include/roboteam_ai/world/FieldComputations.h
```diff
@@ -23,25 +21,214 @@ namespace rtt::ai {
...
+         * LineSegment and the opponents defence area.
+         * @param lineStart The position which indicates the start of the LineSegment.
+         * @param lineEnd The position which indicates the end of the LineSegment.
+         * @param margin The outwards margin with which the defence area will be expanded/shrinked for determining the intersection between the LineSegment and defence area.
+         * A positive value means that the rectangular defence area will be expanded in both y directions and towards the center of the field, a negative value means that the
+         * rectangular defence area will be shrinked in both y directions and away from the goal of that defence area.
+         * @return The closest intersection point to the start of the LineSegment. If there is no intersection point then a null pointer will be returned.
+         */
+        static std::shared_ptr<Vector2> lineIntersectionWithDefenceArea(const Field &field, bool ourGoal, const Vector2& lineStart, const Vector2& lineEnd, double margin);
```
Yuhanun:
```suggestion
        static std::optional<Vector2> lineIntersectionWithDefenceArea(const Field &field, bool ourGoal, const Vector2& lineStart, const Vector2& lineEnd, double margin);
```

include/roboteam_ai/world/FieldComputations.h
```diff
@@ -23,25 +21,214 @@ namespace rtt::ai {
```
Yuhanun, line 19: Honestly a lot of documentation in this file feels very redundant

include/roboteam_ai/world/FieldComputations.h
```diff
@@ -13,6 +7,10 @@
 #include "mutex"
 #include "roboteam_proto/GeometryFieldSize.pb.h"
 #include "roboteam_proto/messages_robocup_ssl_geometry.pb.h"
+#include <control/ControlUtils.h>
+#include <interface/api/Input.h>
+#include "world/World.h"
+#include "world/WorldData.h"
```
Yuhanun: use "" or <>, dont mix

include/roboteam_ai/world/FieldComputations.h
```diff
@@ -13,6 +7,10 @@
 #include "mutex"
 #include "roboteam_proto/GeometryFieldSize.pb.h"
 #include "roboteam_proto/messages_robocup_ssl_geometry.pb.h"
```
Yuhanun:
```suggestion
#include <mutex>

#include <roboteam_proto/GeometryFieldSize.pb.h>
#include <roboteam_proto/messages_robocup_ssl_geometry.pb.h>
```

roboteam_proto is a library, mutex is std

include/roboteam_ai/world/FieldComputations.h
```diff
@@ -23,25 +21,214 @@ namespace rtt::ai {
 
 using namespace rtt::ai::world;
```
Yuhanun: Nope.
```suggestion
namespace rtt_world = rtt::ai::world;
```

include/roboteam_ai/world/FieldComputations.h
Yuhanun:
```suggestion
#include "control/ControlUtils.h"
#include "interface/api/Input.h"
#include "world/World.h"
#include "world/WorldData.h"
```

Yuhanun: They're all part of the same project so yeah

Yuhanun (changes requested):

include/roboteam_ai/world/Field.h
mrlukasbos: No, this is something very inherent in the rtt system. the center of the field is defined as 0,0

src/world/Field.cpp
rolfvdhulst: I am not even talking about the fact tests are breaking. I am talking that about the fact you did not write any new ones.

include/roboteam_ai/world/Field.h
rolfvdhulst: This is the SSL standard as well. We literally cannot change it (or well, we can but it would be a headache ina  lot of places)

include/roboteam_ai/world/Field.h
mrlukasbos: Guess it would be even better if these functions just return an optional. Skills relying on such a value needs to have some fail safe anyway

RobotJente:
I agree with @rolfvdhulst @mrlukasbos and @Yuhanun that there is too much documentation. A lot of it is redundant because it is self-explanatory to a large extent. I like the new picture based field documentation. I also see the point @mrlukasbos makes when he says you are making too many variables which actually ends up confusing people more. There is a fine line between too many and too few and I think now this class has too many of these variables.

include/roboteam_ai/world/Field.h
HaicoDorenbos: This confusion convinces me that all this documentation and comments are really needed

include/roboteam_ai/world/Field.h
HaicoDorenbos: I give it a more clear name and add more comments which explains it...

include/roboteam_ai/world/FieldComputations.h
HaicoDorenbos: > Which rules?

According to our unwritten rules

include/roboteam_ai/world/Field.h
HaicoDorenbos: I decided to not document these methods, since my first approach was to have 1 function that return all possible FieldLines, 1 function that returns all possible FieldValues and 1 function that returns all locations. Since nobody liked this design choice, I use multiple getters instead. However I refuse to write documentation for all of them, because this was also one of the reasons why I prefer to have just a single getter for everything instead.

src/coach/defence/DefencePositionCoach.cpp
HaicoDorenbos: This is outside the scope of this pull request. This PR should mainly change things inside the Field and FieldComputation class.

include/roboteam_ai/world/Field.h
rolfvdhulst: I would call the above a discussion, not confusion

src/coach/defence/DefencePositionCoach.cpp
HaicoDorenbos: Is also outside the scope of this PR.

src/coach/heuristics/OffensiveScore.cpp
HaicoDorenbos: So you imply that both of these functions always return False? If that is the case then an issue should be made for this. However it should not be fixed in this PR, because it is out the scope of this PR.

src/world/FieldComputations.cpp
HaicoDorenbos: Why do you want to copy a class and not pass it by reference?

src/world/FieldComputations.cpp
HaicoDorenbos: In this case I stick to the old choice, because it cost me more time to refactor it to std::optional. Also programmers often use nullptr as some kind of optional variable, so I currently don't see why it is that bad. But if you show me your arguments why nullptr is worse than an std::optional then I probably quickly agree with your reasoning.

src/world/FieldComputations.cpp
HaicoDorenbos: I am afraid to change it. I know at a lot of places in the code they did this to prevent the variable from being forgotten which causes strange random runtime errors.

include/roboteam_ai/world/FieldComputations.h
Yuhanun: I actually think we should write them down :eyes:

src/world/FieldComputations.cpp
HaicoDorenbos: What is the difference? I more like the version where the constructor is called with a vector of positions.

include/roboteam_ai/world/Field.h
Yuhanun: Also it's not really confusion, i'm quite new to the system and i'm not as experienced with roboteam code as the rest of you, i wouldn't know how we handle the field.

I'm not saying this drawing is useless, the contrary.

There's a lot of documentation that's redundant though, that's what bothers me, not this. I think this is great.

include/roboteam_ai/world/Field.h
Yuhanun: Well yeah the documentation isn't really the issue if it just returned the std::optional, im just wondering what happens to the actual optional, does it just return .value() or does it return a default value such as 0 if empty()?

Yuhanun: > Guess it would be even better if these functions just return an optional. Skills relying on such a value needs to have some fail safe anyway

this.

src/coach/heuristics/OffensiveScore.cpp
Yuhanun: I know, it's just things that i noticed

HaicoDorenbos:
I read all the feedback you gave me and made the following decisions:
- I agree that test cases regarding the Field & FieldComputation class should be added before merging this pull request.
- I agree that returning an optional instead of just a value is better. However this will be left to do for a later pull request. Otherwise this pull request gets way to big.
- Regarding a lot of @Yuhanun comments I do not solve problems outside the Field and FieldComputation class that were already present and were only partially solved.
- Regarding @Yuhanun comments about the FIeld and FieldComputation classes. I agree with almost all of your comments. But you still need to give me your reasoning why a nullptr should be replaced by an optional. If you give me your reasons then I will most likely quickly be convinced.
- I do not agree with @mrlukasbos and @RobotJesse that there are too much variables in the Field class. In the past I proposed the idea of using maps instead, but unfortunately you did not agree with me. Also I do not agree with computation of these variables outside the Field class, because this is what currently happens and what causes a lot of code cloning problems. From my point of view every interesting location, line or value about the Field should be retrievable from the Field class.
- I agree with @rolfvdhulst that the bug inside robotBlockade needs to be solved first.
- I agree with @timovdk point that documentation cloning should be avoided and I agree with @RobotJesse  that some comments are still not clear and should therefore be improved. However I do not agree that there is too much documentation, because:
1. I have never encountered a project where there was too much documentation. Shortage of documentation was almost always the problem.
2. In Dutch we always say: "Goed voorbeeld doet volgen." (English translation: "Good example will follow"). By writing a lot of documentation I encourage others and future teams to also properly document their code.
3. An old programming teacher always said that every (private/public) global variable needs at least a comment that explains it. I personally strongly agree with them especially for the Field class, because these globals are not as clear as they seem to be.
4. The team switches every year, so knowledge will be lost every year. This means that the value of documentation is even higher.
5. A lot of these documentation/comments is added, because I did not understand these things at first sight and a lot of code makes assumptions about these variables. For example that the start of every horizontal field line is the left most point of this field line.
6. Being consistent  is important. For example isPointInField is a very clear function, however it is a public function which therefore needs documentation according to our unwritten rules.

include/roboteam_ai/world/Field.h
HaicoDorenbos: It returns a default value 0 if it is undefined which is also what it did before.

include/roboteam_ai/world/Field.h
HaicoDorenbos: Well you should also be able to understand the code. So if comments were able to let you understand the code then that already shows me that it is not redundant.

include/roboteam_ai/world/FieldComputations.h
HaicoDorenbos: I indeed agree, but it does not have the highest priority level compared to other adaptive maintenance tasks. So probably it will be postponed, unless someone else wants to write them down.

include/roboteam_ai/world/Field.h
```diff
@@ -33,6 +33,24 @@ struct FieldArc {
...
+ *              (-x,y)_________________________ (x,y)
+ *                 |                              |
+ *                 |                              |
+ * (your side)     |             (0,0)            | (their side)
+ * (leftFieldLine) |                              | (rightFieldLine)
+ *                 |                              |
+ *              (-x, -y)_______________________(x, -y)
+ *                          (bottomFieldLine)
+ *
+ * Thus the left side of the field always corresponds to our side of the field and the right side of the field always corresponds to the opponents side of the field. Moreover
```
RobotJente, line 48:
```suggestion
 * Thus the left side of the field always corresponds to our side of the field. Moreover
```

HaicoDorenbos: I do not agree, that part of the explanation is important...

src/coach/GetBallCoach.cpp
```diff
@@ -18,7 +18,7 @@ bool GetBallCoach::shouldWeGetBall(const Field &field) {
     // should probably listen to ballPossession at some point
     Vector2 ballPos = world_new::World::instance()->getWorld()->getBall()->get()->getPos();
     return !FieldComputations::pointIsInDefenceArea(field, ballPos, true, 0.04) && !FieldComputations::pointIsInDefenceArea(field, ballPos, false) &&
-           FieldComputations::pointIsInField(field, ballPos, -0.05);
+           FieldComputations::pointIsInField(field, ballPos, 0.05);
```
RobotJente: Why did you change the -?

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -33,12 +33,11 @@ Vector2 DefencePositionCoach::getPosOnLine(const Line &line, double aggressionFa
...
-double DefencePositionCoach::getOrientation(const Line &line) {
-    return (line.start - line.end).angle();
-}
+double DefencePositionCoach::getOrientation(const Line &line) { return (line.start - line.end).angle(); }
 
 // computes a line segment on which the entirety of openGoalSegment is blocked as seen from point with robots with radius collissionRadius
-std::shared_ptr<Line> DefencePositionCoach::getBlockLineSegment(const Field &field, const Line &openGoalSegment, const Vector2 &point, double collisionRadius, double margin) {
+std::shared_ptr<Line> DefencePositionCoach::getBlockLineSegment(const rtt::ai::world::Field &field, const Line &openGoalSegment, const Vector2 &point, double collisionRadius,
+                                                                double margin) {
     if (margin == -1.0) {
```
RobotJesse: formatting

src/coach/defence/DefencePositionCoach.cpp
```diff
@@ -33,12 +33,11 @@ Vector2 DefencePositionCoach::getPosOnLine(const Line &line, double aggressionFa
...
 // get the direction facing towards the end of the Line
-double DefencePositionCoach::getOrientation(const Line &line) {
-    return (line.start - line.end).angle();
-}
+double DefencePositionCoach::getOrientation(const Line &line) { return (line.start - line.end).angle(); }
 
 // computes a line segment on which the entirety of openGoalSegment is blocked as seen from point with robots with radius collissionRadius
-std::shared_ptr<Line> DefencePositionCoach::getBlockLineSegment(const Field &field, const Line &openGoalSegment, const Vector2 &point, double collisionRadius, double margin) {
+std::shared_ptr<Line> DefencePositionCoach::getBlockLineSegment(const rtt::ai::world::Field &field, const Line &openGoalSegment, const Vector2 &point, double collisionRadius,
+                                                                double margin) {
```
RobotJente:
```suggestion
std::shared_ptr<Line> DefencePositionCoach::getBlockLineSegment(const rtt::ai::world::Field &field, const Line &openGoalSegment, const Vector2 &point, double collisionRadius, double margin) {
```

src/control/ball-handling/DribbleBackwards.cpp
```diff
@@ -178,7 +178,7 @@ RobotCommand DribbleBackwards::sendOvershootCommand(const Field &field, world_ne
     command.vel = (approachPosition - _robot->getPos()).stretchToLength(std::min(0.2, maxVel));
     command.angle = lockedAngle;
 
-    if (failedOnce && !FieldComputations::pointIsInField(world_new::World::instance()->getField().value(), _ball->getPos(), Constants::ROBOT_RADIUS())) {
+    if (failedOnce && !FieldComputations::pointIsInField(world_new::World::instance()->getField().value(), _ball->getPos(), -Constants::ROBOT_RADIUS())) {
```
RobotJente: Here also, why the - change?

src/interface/widgets/widget.cpp
RobotJente: I actually think the explicit Vector2 call makes it more clear

RobotJente (commented):
The changes look fine to me. However, would like to see some tests like @rolfvdhulst mentioned

src/coach/GetBallCoach.cpp
HaicoDorenbos: Margin from now on are all outwards. Some margins in the FieldComputations where inwards instead of outwards.

test/WorldTests/FieldComputationTest.cpp
```diff
@@ -0,0 +1,14 @@
...
+// Created by Haico Dorenbos on 20-03-2020
+//
+
+#include <gtest/gtest.h>
+
+namespace rtt {
+
+// Checks whether the margins of the functions are all outwards
+TEST(FieldComputationTest, outwards_margin) {
+
```
timovdk, line 179:
```suggestion
```

HaicoDorenbos:
Everything should be improved by now :+1: However this PR can only be merged if the PR Bug/polygon intersect #63 on the Roboteam_utils branch is merged first.

timovdk (approved):

include/roboteam_ai/world/Field.h
RobotJente: [A scientific paper about why you are wrong](https://me.me/i/there-are-two-types-of-people-in-this-world-1-15505643)

PR #852, opened 2020-03-06
Changed virtual/pure virtual

jordiluong:
Changed some virtual to pure virtual and vice versa to make it more logical for implementing the skills

### DIFF
--- a/include/roboteam_ai/stp/Skill.h
+++ b/include/roboteam_ai/stp/Skill.h
@@ -69,7 +69,7 @@ namespace rtt::ai::stp {
          * Initializes the skill
          * @return Status of initialization
          */
-        virtual Status onInitialize() noexcept {};
+        virtual Status onInitialize() noexcept = 0;
 
         /**
          * Resets all the robot controllers in the RobotController struct
--- a/include/roboteam_ai/stp/Skill.h
+++ b/include/roboteam_ai/stp/Skill.h
@@ -56,7 +56,7 @@ namespace rtt::ai::stp {
          * Terminates the skill
          * @return Status of termination
          */
-        virtual Status onTerminate() noexcept {};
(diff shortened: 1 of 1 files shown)


include/roboteam_ai/stp/Skill.h
```diff
@@ -56,20 +56,20 @@ namespace rtt::ai::stp {
          * Terminates the skill
          * @return Status of termination
          */
-        virtual Status onTerminate() noexcept;
+        virtual Status onTerminate() noexcept {};
```
RobotJesse: should be purely virtual

RobotJente: I think this should be purely virtual
```suggestion
        virtual Status onTerminate() noexcept = 0;
```

include/roboteam_ai/stp/Skill.h
```diff
@@ -56,20 +56,20 @@ namespace rtt::ai::stp {
...
          */
-        virtual Status onUpdate(SkillInfo const& info) noexcept;
+        virtual Status onUpdate(SkillInfo const& info) noexcept = 0;
 
         /**
          * Initializes the skill
          * @return Status of initialization
          */
-        virtual Status onInitialize() noexcept;
+        virtual Status onInitialize() noexcept {};
```
RobotJente: This too 
```suggestion
        virtual Status onInitialize() noexcept = 0;
```

include/roboteam_ai/stp/Skill.h
jordiluong: I was thinking about this. It is cleaner to make it pure, so it is the responsibility of the derived skills. But the skills mostly do not need termination/initialization, so having it not pure is just convenience. If you think making it pure would be better, I'll go with that.

RobotJente (approved):
looks good

PR #864, opened 2020-03-09
GoToPos skill

jordiluong:
Implementation of GoToPos skill

### DIFF
--- a/include/roboteam_ai/stp/StpInfo.h
+++ b/include/roboteam_ai/stp/StpInfo.h
@@ -73,7 +73,7 @@ namespace rtt::ai::stp {
         void setPosition(Vector2 position) { this->position = position; };
 
         std::optional<Field> getField() const { return field; };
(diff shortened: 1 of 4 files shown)


include/roboteam_ai/stp/StpInfo.h
```diff
@@ -69,11 +69,11 @@ namespace rtt::ai::stp {
...
 
+        Vector2 getPosition() const { return position; };
+        void setPosition(Vector2 position) { this->position = position; };
 
-        /**
-         * Position, could be position of robot, or whatever position the skill needs, maybe target position
-         */
-        Vector2 position;
+        std::optional<Field> getField() const { return field; };
+        void setField(std::optional<Field> field) { this->field = field; };
```
timovdk: I think this function should not take an optional as parameter, since you should actually have a field when you call `setField`, not maybe have a field.

timovdk (approved):

PR #868, opened 2020-03-09
Refactor/play decider checker

### LINKED ISSUE #845: Play class implementation + tests
Implementation:
Contains a central place for decision making (what robot should pass to what robot, etc).
Assigns robots to roles. See #844 for more details on role assignments with the new dealer.
The roles need to be given relevant information (for example, what robot should pass to what robot) from the play

Yuhanun:
Play decider and play checker classes, feel free to add onto them / suggest changes.

Does *no longer* update the binary tree, we'll have to play->execute() or something.


include/roboteam_ai/ApplicationManager.h
```diff
@@ -28,20 +28,25 @@ class ApplicationManager {
     bool fieldInitialized = false;
     bool robotsInitialized = false;
 
+    /**
+     * Current best play as picked by checker + decider
+     */
+    ai::stp::Play* bestPlay;
```
rttwente, line 34: Why raw pointer?

Yuhanun: Because the std::vector in PlayChecker owns the data (std::unique_ptr), and passing .get() is the easiest and cleanest way to use these values without copies. (T& would remove copy and move assignment operators)

include/roboteam_ai/stp/PlayDecider.hpp
```diff
@@ -0,0 +1,19 @@
...
+#ifndef RTT_PLAYDECIDER_HPP
+#define RTT_PLAYDECIDER_HPP
+
+#include "Play.hpp"
+
+namespace rtt::ai::stp {
+
+    class PlayDecider {
+    public:
+        Play *decideBestPlay(world_new::World *pWorld, std::vector<Play*> plays) noexcept;
```
rttwente, line 14: Why raw pointer as return type?

rttwente (approved):
Apart from the raw pointers, it looks good.

include/roboteam_ai/stp/PlayDecider.hpp
Yuhanun: Same reason.

PR #873, opened 2020-03-10
Set angle command to current robot angle

jordiluong:
Set angle command to current robot angle to keep the robot on the right angle


Yuhanun (approved):
LGTM

PR #875, opened 2020-03-11
HasBall skill

jordiluong:
HasBall is not really a skill because it does not generate robot commands. It is a modular way to implement this check in the tactics.


RobotJente (approved):
Looks good!

PR #887, opened 2020-03-12
Added failing conditions for tactics

Ratoone:
Failing conditions include:
- when does a tactic fail?
- when does a tactic reset?
- is the tactic an end tactic (can it succeed in the first place)?

Added checks in the update function for these situations


timovdk (approved):

RobotJente (approved):
Looks good!

PR #893, opened 2020-03-14
Fix race condition

rolfvdhulst:
Small fix for a race condition we have. I only found this on my home pc because it's quite fast, probably. I got a bad optional access here. This should fix it.


Yuhanun (approved):
LGTM

PR #897, opened 2020-03-16
stp receive tactic

jordiluong:
Implementation of receive tactic


RobotJente (approved):
I wants to merge this

PR #908, opened 2020-03-18
Feature/stp role class

jordiluong:
Added methods to Role class we need to use to calculate tactic info and determine when role should reset

### DIFF
--- a/include/roboteam_ai/stp/Role.hpp
+++ b/include/roboteam_ai/stp/Role.hpp
@@ -48,7 +48,6 @@ class Role {
 
     /**
      * This function should calculate any extra information that the tactics might need to be executed.
-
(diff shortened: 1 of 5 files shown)


include/roboteam_ai/stp/Role.hpp
```diff
@@ -45,6 +45,21 @@ class Role {
      * State machine that keeps track of tactic states
      */
     collections::state_machine<Tactic, Status, StpInfo> robotTactics;
+
+    /**
+     * This function should calculate any extra information that the tactics might need to be executed.
+
```
timovdk:
```suggestion
```

timovdk (approved):

RobotJente (approved):

PR #917, opened 2020-03-19
Reset state machines of role and tactics

jordiluong:
Reset all tactics state machines when resetting role. This fixes the problem of tactics being successful immediately after the first time.

### DIFF
--- a/include/roboteam_ai/stp/Tactic.h
+++ b/include/roboteam_ai/stp/Tactic.h
@@ -100,10 +100,9 @@ class Tactic {
     Tactic(Tactic &&other) = default;
 
     /**
-     * Returns the skills state machine
-     * @return Skills state machine
+     * Reset the state machine
      */
-    rtt::collections::state_machine<Skill, Status, StpInfo> & getSkills() { return skills; };
+    void reset() noexcept;
 };
 }  // namespace rtt::ai::stp
 
--- a/include/roboteam_ai/stp/Tactic.h
+++ b/include/roboteam_ai/stp/Tactic.h
@@ -98,6 +98,12 @@ class Tactic {
      * Default move-ctor, ensures proper move-construction of Tactic
      */
     Tactic(Tactic &&other) = default;
+
+    /**
+     * Returns the skills state machine
+     * @return Skills state machine
+     */
+    rtt::collections::state_machine<Skill, Status, StpInfo> & getSkills() { return skills; };
 };
 }  // namespace rtt::ai::stp
 
--- a/src/stp/Role.cpp
+++ b/src/stp/Role.cpp
@@ -23,8 +23,8 @@ Status Role::update(StpInfo const& info) noexcept {
     if (robotTactics.current_num() != 0 && shouldRoleReset(tacticInfo)) {
         RTT_INFO("State Machine reset for current role for ID = ", tacticInfo.getRobot()->get()->getId())
         // Reset all the Tactics state machines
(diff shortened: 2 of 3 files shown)


src/stp/Role.cpp
```diff
@@ -16,16 +16,20 @@ Status Role::update(StpInfo const& info) noexcept {
...
     // Update tactic info
     auto tacticInfo = calculateInfoForTactic(info);
 
+    // Reset the role
+    if (robotTactics.current_num() != 0 && shouldRoleReset(tacticInfo)) {
+        RTT_INFO("State Machine reset for current role for ID = ", tacticInfo.getRobot()->get()->getId())
+        // Reset all the Tactics state machines
+        for (auto tactics = robotTactics.begin(); tactics < robotTactics.end(); tactics++) {
+            tactics->get()->getSkills().reset();
+        }
```
RobotJente:
```suggestion
        for (auto& tactic : robotTactics) {
            tactic->getSkills().reset();
        }
```

include/roboteam_ai/stp/Tactic.h
```diff
@@ -98,6 +98,12 @@ class Tactic {
      * Default move-ctor, ensures proper move-construction of Tactic
      */
     Tactic(Tactic &&other) = default;
+
+    /**
+     * Returns the skills state machine
+     * @return Skills state machine
+     */
+    rtt::collections::state_machine<Skill, Status, StpInfo> & getSkills() { return skills; };
```
Ratoone: This is bad design, an alternative is presented in the role

src/stp/Role.cpp
```diff
@@ -16,20 +16,24 @@ Status Role::update(StpInfo const& info) noexcept {
...
-
     // Update tactic info
     auto tacticInfo = calculateInfoForTactic(info);
 
+    // Reset the role
+    if (robotTactics.current_num() != 0 && shouldRoleReset(tacticInfo)) {
+        RTT_INFO("State Machine reset for current role for ID = ", tacticInfo.getRobot()->get()->getId())
+        // Reset all the Tactics state machines
+        for (auto& tactic : robotTactics) {
+            tactic->getSkills().reset();
```
Ratoone: Insted of resetting the actual state machine of each tactic, I would rather have the base tactic contain a reset method that resets the state machine of the tactic. It's cleaner and more maintainable.

```suggestion
            tactic->.reset();
```

src/stp/Tactic.cpp
```diff
@@ -30,17 +30,16 @@ Status Tactic::update(StpInfo const &info) noexcept {
...
+
     // the tactic will not be reset if it's the first skill; it will be reset as well if the
     // state machine is finished
-    if(skills.finished() || (skills.current_num() != 0 && shouldTacticReset(info))){
-        RTT_INFO("State Machine reset for current tactic for ID = ", info.getRobot()->get()->getId())
-        // TODO: messy reset, do it in the state machine
-        skills.skip_n(- skills.current_num());
+    if(skills.finished() || (skills.current_num() != 0 && shouldTacticReset(skill_info))){
+        RTT_INFO("State Machine reset for current tactic for ID = ", skill_info.getRobot()->get()->getId())
+        skills.reset();
```
Ratoone: Considering the other requests, this could then change into *see suggestion*. There is no difference, but this makes it a bit cleaner (probably?)

```suggestion
        this.reset();
```

Ratoone (changes requested):
The reset should be performed in the Tactic class, as the Role should not access the lower state machine

Ratoone (approved):

PR #922, opened 2020-03-20
Refactor/stp remove position enum

### LINKED ISSUE #914: Refactor PositionEnum pair

jordiluong:
#914 
Changed position enum to separate variables

### DIFF
--- a/include/roboteam_ai/stp/StpInfo.h
+++ b/include/roboteam_ai/stp/StpInfo.h
@@ -104,8 +104,14 @@ struct StpInfo {
     const std::optional<world_new::view::BallView>& getBall() const { return ball; }
     void setBall(const std::optional<world_new::view::BallView>& ball) { this->ball = ball; }
 
-    const std::pair<PositionType, Vector2>& getPosition() const { return position; }
-    void setPosition(const std::pair<PositionType, Vector2>& position) { this->position = position; }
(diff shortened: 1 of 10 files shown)


timovdk (commented):
So we don't need Receive target anymore?

jordiluong:
> So we don't need Receive target anymore?

No, I dont think we need that. It is only used for the Receive tactic and the only position it ever needs to know is where to move to, so we can use positionMoveTo. Agree?

timovdk:
> > So we don't need Receive target anymore?
> 
> No, I dont think we need that. It is only used for the Receive tactic and the only position it ever needs to know is where to move to, so we can use positionMoveTo. Agree?

Yeah, sounds fine.

include/roboteam_ai/stp/StpInfo.h
```diff
@@ -144,9 +144,19 @@ struct StpInfo {
     std::optional<world_new::view::BallView> ball;
 
     /**
-     * Tuple of the PositionType and the position of this target
+     * Position to move to
+     */
+    std::optional<Vector2> positionMoveTo;
```
RobotJente: positionToMoveTo

jordiluong: Oke baas

RobotJente (approved):

PR #963, opened 2020-03-25
made print statement for play name and changed play constructor

RobotJente:
Quality of Life: Made debug messages to see which play is currently executed

### DIFF
--- a/include/roboteam_ai/stp/Play.hpp
+++ b/include/roboteam_ai/stp/Play.hpp
@@ -61,7 +61,7 @@ class Play {
     /**
      * Default ctor, ensures proper construction of Play
      */
-    Play() = default;
+    explicit Play(std::string playName);
 
     /**
      * Default move-ctor, ensures proper move-construction of Play
@@ -87,6 +87,7 @@ class Play {
     [[nodiscard]] bool arePlayRolesFinished();
 
    protected:
+    std::string playName;
     /**
      * The roles, constructed in ctor of a play
      */
--- a/include/roboteam_ai/stp/Play.hpp
+++ b/include/roboteam_ai/stp/Play.hpp
@@ -59,7 +59,7 @@ class Play {
     virtual ~Play() = default;
 
     /**
-     * Default ctor, ensures proper construction of Play
+     * Ctor that constructs a play an assigns its name
      */
     explicit Play(std::string playName);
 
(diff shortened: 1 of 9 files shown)


include/roboteam_ai/stp/Play.hpp
```diff
@@ -61,7 +61,7 @@ class Play {
     /**
      * Default ctor, ensures proper construction of Play
```
timovdk: It's not default anymore
```suggestion
     * Ctor that constructs a play an assigns its name
```

src/ApplicationManager.cpp
```diff
@@ -30,8 +30,8 @@ void ApplicationManager::start() {
     RTT_INFO("Waiting for field_data and robots...");
 
     auto plays = std::vector<std::unique_ptr<rtt::ai::stp::Play>>{};
-    plays.emplace_back(std::make_unique<rtt::ai::stp::TestPlay>());
-    plays.emplace_back(std::make_unique<rtt::ai::stp::play::Pass>());
+    plays.emplace_back(std::make_unique<rtt::ai::stp::TestPlay>("Test Play"));
+    plays.emplace_back(std::make_unique<rtt::ai::stp::play::Pass>("Pass Play"));
```
timovdk: I'd prefer just "test" and "pass"

timovdk:
For this feature I'm more in favor of what already was implemented in skill.h and .cpp, a hardcoded name so we can have the name as a constexpr

RobotJente:
I disagree with your suggestion to do it like we do it in skill, because that requires another virtual function (which is also sacrificing performance, though the performance here is not what worries me) but also because doing it in the constructor seems like a clear place to initialize something like this, and you can just call the default constructor of skill, saving a bunch of mindless lines of code when you have to reimplement the virtual function in your derived class.

timovdk:
> I disagree with your suggestion to do it like we do it in skill, because that requires another virtual function (which is also sacrificing performance, though the performance here is not what worries me) but also because doing it in the constructor seems like a clear place to initialize something like this, and you can just call the default constructor of skill, saving a bunch of mindless lines of code when you have to reimplement the virtual function in your derived class.

I am not sure what the best solution is, I am just in favor of moving potential bugs to compile time instead of runtime; not implementing the purely virtual function leads to a compilation error whereas forgetting to set name in the ctor can lead to runtime bugs.

How you do it in this PR, using the base class ctor is also quite nice though, it is clear and generic. Typing the name of the class twice just feels a bit unnecessary though and might lead to typos (which probably don't matter that much since name is used to visualise. not to do calculations or whatever)

PR #966, opened 2020-03-25
Added ballSensor input to hasBall

### LINKED ISSUE #957: Add ballSensor check to the kick skill
Sabine recommended this. Also keep in mind that there should be a way to override the ballsensor input, since it does not always work (kind of a force kick). The old shotcontrroller implements it already, maybe look at that? idk

### LINKED ISSUE #958: Add ballSensor check to the chip skill
Sabine recommended this. Also keep in mind that there should be a way to override the ballsensor input, since it does not always work (kind of a force kick). The old shotcontrroller implements it already, maybe look at that? idk

timovdk:
Added ballSensor input to the hasBall check.
closes #957, closes #958


src/world_new/Robot.cpp
```diff
@@ -88,6 +88,14 @@ bool Robot::isWorkingBallSensor() const noexcept { return workingBallSensor; }
 
 void Robot::setWorkingBallSensor(bool _workingBallSensor) noexcept { Robot::workingBallSensor = _workingBallSensor; }
 
+bool Robot::hasBallBallSensor() const noexcept { return hasBall; }
```
RobotJente: The naming is a bit weird, maybe something like this:
```suggestion
bool Robot::BallSensorSeesBall() const noexcept { return hasBall; }
```

RobotJente: Also the implementation does not make sense to me. What is hasBall? There is a function with that name as well in RobotView, that does something different.

timovdk: Done, also changed member variable name

RobotJente (dismissed):

src/world_new/views/RobotView.cpp
```diff
@@ -27,8 +31,12 @@ Vector2 RobotView::getKicker() const noexcept {
...
 RobotView::operator bool() const noexcept { return get() != nullptr; }
 
-robot::RobotControllers &RobotView::getControllers() const noexcept { 
-    return World::instance()->getControllersForRobot(get()->getId()); 
+robot::RobotControllers &RobotView::getControllers() const noexcept { return World::instance()->getControllersForRobot(get()->getId()); }
+
+bool RobotView::hasBallAccordingToVision(double maxDist, double maxAngle) const noexcept {
+    auto dist = get()->getDistanceToBall();
+    auto angle = get()->getAngleDiffToBall();
+    return dist < maxDist && angle < maxAngle*2;
```
timovdk: Missed a magic number

src/stp/new_skills/GoToPos.cpp
```diff
@@ -26,7 +26,9 @@ Status GoToPos::onUpdate(const StpInfo &info) noexcept {
     // Set velocity and angle commands
     command.mutable_vel()->set_x(targetVelocity.x);
     command.mutable_vel()->set_y(targetVelocity.y);
-    command.set_w(robotCommand.angle.getAngle());
+
+    if(info.getAngle()) command.set_w(info.getAngle().getAngle());
+    else command.set_w(robotCommand.angle.getAngle());
```
RobotJente, line 31: Do we really need to use info.getAngle().getAngle() ? It seems a bit weird to call getAngle() twice

src/world_new/Robot.cpp
```diff
@@ -32,6 +32,7 @@ Robot::Robot(std::unordered_map<uint8_t, proto::RobotFeedback> &feedback, const
 
     if (ball.has_value()) {
         setDistanceToBall(pos.dist((*ball)->getPos()));
+        setAngleDiffToBall(fabs(angle.shortestAngleDiff(((*ball)->getPos() - pos).angle())));
```
RobotJente:
```suggestion
        auto angleRobotToBall ((*ball)->getPos() - pos).angle());
        setAngleDiffToBall(fabs(angle.shortestAngleDiff(angleRobotToBall));
```

RobotJente (commented):
Code looks like it should work, but I'd advise (you may have done already) to test it in GRsim to see if the angleMargins indeed work as expected, or write a unit test for this, as these kinds of things are quite bug-prone. Other than that it looks good

include/roboteam_ai/stp/new_constants/ControlConstants.h
```diff
@@ -6,37 +6,44 @@
 #define RTT_CONTROLCONSTANTS_H
 
 namespace rtt::ai::stp::control_constants {
```
Yuhanun, line 8:
```suggestion
namespace rtt::control::constants {
```
Theyre not just for STP, right?

Yuhanun (approved):
lgtm

src/stp/new_skills/GoToPos.cpp
timovdk: Yes, we need a double. stpInfo.getAngle() returns an Angle, Angle.getAngle() returns a double. I think the last getAngle() its naming is wrong, but that's in the Angle class so out of scope for this PR

include/roboteam_ai/stp/new_constants/ControlConstants.h
timovdk: I think this is better, but Jesse made this, so what do you think @RobotJesse ?

RobotJente: For now we will leave it like this, when STP is done we will refactor the namespace.

src/world_new/Robot.cpp
```diff
@@ -32,6 +32,8 @@ Robot::Robot(std::unordered_map<uint8_t, proto::RobotFeedback> &feedback, const
 
     if (ball.has_value()) {
         setDistanceToBall(pos.dist((*ball)->getPos()));
+        auto angleRobotToBall ((*ball)->getPos() - pos).angle());
```
timovdk:
```suggestion
        auto angleRobotToBall ((*ball)->getPos() - pos).angle();
```

PR #975, opened 2020-03-27
Implemented the Harasser role

### LINKED ISSUE #974: HarasserRole

timovdk:
It uses the BlockRobot and Intercept tactics
Closes #974


CMakeLists.txt
```diff
@@ -286,7 +287,7 @@ set(TEST_SOURCES
         ${PROJECT_SOURCE_DIR}/test/StpTests/PlayCheckerTests.cpp
         ${PROJECT_SOURCE_DIR}/test/StpTests/PlayDeciderTests.cpp
         ${PROJECT_SOURCE_DIR}/test/StpTests/TacticTests.cpp
-        )
+        src/stp/new_roles/Harasser.cpp)
```
jordiluong:
```suggestion
        )
```

jordiluong (approved):

Ratoone (changes requested):
Simple and on point. Need just to fix the build. As I see it here, maybe we should also consider additional logic in the role whether it should decide by itself to skip the end tactic? Here it seems possible

timovdk:
> Simple and on point. Need just to fix the build. As I see it here, maybe we should also consider additional logic in the role whether it should decide by itself to skip the end tactic? Here it seems possible

I think we need to discuss this, since it is not that defined whether we have logic in roles or not and what logic it should be if there should be logic

RobotJente (approved):

RobotJente (approved):

PR #976, opened 2020-03-27
Feature/stp defender role

### LINKED ISSUE #968: DefenderRole

jordiluong:
Implementation of defender role and stub for defend play. Implementation of defend play was just for testing the role

Requires PR 62 of roboteam_utils
Closes #968


include/roboteam_ai/stp/Role.hpp
```diff
@@ -15,57 +15,69 @@ namespace rtt::ai::stp {
...
+    /**
+    * Forces the Role to skip to the next tactic in the state machine
+    */
+    void forceNextTactic() noexcept;
+
+    /**
+     * Forces the Role to skip to tactic n in the state machine
+     * @param n Index of the tactic in the state machine
+     */
+    void forceToTactic(int n) noexcept;
```
Ratoone: Is this really needed? Usually you either reset or force next tactic, which is more readable

Ratoone (changes requested):
Apart from forceToTactic(n) which i'm not it makes sense to exist, looks fine. Still needs to fix build though

include/roboteam_ai/stp/Role.hpp
jordiluong: The reason I did this is that the defender role has two tactics, blockrobot and intercept. For the play we want all the defenders to block a robot, except that whenever the ball is moving, the closest defender will try to intercept the ball, for example. In that case the play can tell that specific robot to skip to the intercept tactic. The problem with forcing next tactic is then that the play has to know which tactic the robot is executing, to avoid constantly forcing to the next tactic.

Ratoone: In this particular case you do know, as the Role will no longer return Waiting after forcing the next tactic

RobotJente (approved):
Looks good!

PR #990, opened 2020-04-01
Cleaned up interface naming

RobotJente:
removed some some()  here and there


src/interface/widgets/STPVisualizerWidget.cpp
```diff
@@ -107,7 +107,7 @@ namespace rtt::ai::interface {
             parent->setKeeperRole(role, state);
             return;
         }
-        updateContent << "Some(" << botView->getId() << ") => ";
```
jordiluong, line 110: Maybe make the id more readable somehow since it's only one number behind the role name which also ends with a number

jordiluong (changes requested):

PR #1024, opened 2020-04-14
Improved invariants with fuzzy logic

timovdk:
New way to decide true-ness of an invariant. Fuzzy logic is used to get a value between 0-255, where higher means more true. This value is decided by a piecewise linear function. Using a cutoff number (127 by default) the fuzzy value is compared and a boolean is returned. This way you can obtain either a boolean or a uint8_t for every invariant


include/roboteam_ai/stp/invariants/BaseInvariant.h
```diff
@@ -9,11 +9,29 @@
...
 namespace rtt::ai::stp::invariant {
 class BaseInvariant {
-public:
-    [[nodiscard]] virtual bool checkInvariant(world_new::view::WorldDataView world, const Field *field) const noexcept = 0;
+   public:
+    /**
+     * Converts the fuzzy return to a boolean return
+     * @param world the world
+     * @param field the field
+     * @param cutOff the cutoff number
```
RobotJente:
```suggestion
     * @param cutOff the cutoff number, if the truth value is higher than this number the invariant returns true
```

include/roboteam_ai/stp/invariants/WeHaveBallInvariant.h
```diff
@@ -5,13 +5,13 @@
 #ifndef RTT_WEHAVEBALLINVARIANT_H
 #define RTT_WEHAVEBALLINVARIANT_H
 
-#include "stp/invariants/BaseInvariant.h"
+#include "BaseInvariant.h"
```
RobotJente, line 8: Why did you remove the path to this file?

src/stp/invariants/BallGotShotInvariant.cpp
```diff
@@ -0,0 +1,34 @@
...
+     *          |       XX
+     *          |      XX
+     *          |     XX
+     *   (0,0)  |XXXXXX---------
+     *              (BallSpeed)
+     */
+    piecewiseLinearFunction = nativeformat::param::createParam(0, 255, 0, "ballMovesFast");
+    piecewiseLinearFunction->setYAtX(0.0, 0.0);
+    piecewiseLinearFunction->setYAtX(0.0, stp::control_constants::BALL_IS_MOVING_SLOW_LIMIT + stp::control_constants::FUZZY_MARGIN);
+    piecewiseLinearFunction->linearRampToYAtX(255, stp::control_constants::BALL_GOT_SHOT_LIMIT - stp::control_constants::FUZZY_MARGIN);
```
RobotJente, line 24: I love this function :smiley:

include/roboteam_ai/stp/invariants/BallGotShotInvariant.h
```diff
@@ -0,0 +1,34 @@
+//
+// Created by timovdk on 4/2/20.
+//
+
+#ifndef RTT_BALLGOTSHOTINVARIANT_H
+#define RTT_BALLGOTSHOTINVARIANT_H
+
+#include <NFParam/Param.h>
+
```
timovdk:
```suggestion
```

src/stp/invariants/WeHaveBallInvariant.cpp
```diff
@@ -5,8 +5,8 @@
 #include "stp/invariants/WeHaveBallInvariant.h"
 
 namespace rtt::ai::stp::invariant {
-    bool WeHaveBallInvariant::checkInvariant(world_new::view::WorldDataView world, const Field *field) const noexcept {
-        auto& us = world.getUs();
-        return std::any_of(us.begin(), us.end(), [](auto& robot){return robot.hasBall();});
-    }
+uint8_t WeHaveBallInvariant::metricCheck(world_new::view::WorldDataView world, const Field *field) const noexcept {
+    auto& us = world.getUs();
+    return std::any_of(us.begin(), us.end(), [](auto& robot) { return robot.hasBall(); }) ? stp::control_constants::FUZZY_TRUE : stp::control_constants::FUZZY_FALSE;
```
RobotJente, line 10: I think you are returning FUZZY_TRUE because now the real false and true are taken to be 0 and 255? If that's the case, this is not entirely intuitive and we might be able to find a better solution for this

RobotJente (changes requested):
In general it looks good! I made some small comments. Also, it is not entirely clear to me which parts of the library you adapted, maybe you can put an all-caps flag somewhere.

include/roboteam_ai/stp/invariants/WeHaveBallInvariant.h
timovdk: Because the file that is included is in the same folder

src/stp/invariants/WeHaveBallInvariant.cpp
timovdk: Do you have any idea how this could be improved? Basically this statement returns 255(fuzzy_true) if any of the robots has the ball and 0(fuzzy_false) if none have the ball.

CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 cmake_minimum_required(VERSION 3.10)
 project(roboteam_ai)
 
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS} -g -O0")
-set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS} -Werror -g -O0")
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS} -g -O0 -fPIE")
+set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS} -Werror -g -O0 -fPIE")
```
rolfvdhulst: You shouldn't need this. This is pretty bad, as it means your code is position dependent somehow which means you are doing some very shady c++ magic somewhere (without looking at the entire PR.)

rolfvdhulst (changes requested):

CMakeLists.txt
timovdk: Fixed it by making NFParam shared instead of static. Turns out static libs don't like Address Space Layout Randomization

rolfvdhulst (approved):

src/stp/invariants/WeHaveBallInvariant.cpp
RobotJente: Yeah it's fine functionally, but it might be a bit unclear to people. However, we can fix this later, I'll make it an issue. It is still understandable

RobotJente (approved):

PR #1026, opened 2020-04-17
Design/control utils/deadcode removal

HaicoDorenbos:
Jesse and I discussed that my PR should be much shorter. In this PR I removed the dead functions in the ControlUtils class (are not called by other functions). This PR can only be merged once the PR Redesign/Simplify/Document FieldComputation class #835 is merged.


RobotJente:
I want to point out that a lot of code in our system is "dead" because we have only scratched the surface of the strategy layer of the system. Therefore I'd like this PR to be judged based on whether the code is still relevant for our system, not whether it is currently in use

HaicoDorenbos:
> I want to point out that a lot of code in our system is "dead" because we have only scratched the surface of the strategy layer of the system. Therefore I'd like this PR to be judged based on whether the code is still relevant for our system, not whether it is currently in use

Dead code needs to be removed, because in most of the cases (95%) dead code is not used anymore by anyone. Only if you have very strong beliefs that it will be used in the future you can keep it on the master or development branch. If you are afraid that it will be used or do not want to discard valuable functions then you should store it on a separate branch (but in 95% of the case this branch will not be touched anymore).

Ratoone (commented):
In my opinion we shouldn't remove logic that is not (currently) used, at least not before properly testing our whole system. I see some helper functions that were used in the old system, but were omitted in the new implementation for some reason (e.g. project point outside def area). I propose we come back to this later, for instance in June / July

HaicoDorenbos:
I can agree with that, however 2 other PR's will then be on hold as well which includes: Design/control utils/codeclone removal #1060 and a PR that is still coming. Moreover if a PR is open for a very long time then the risk of merge conflicts gets larger.

jordiluong:
I agree with Paul that we should keep these functions since they will be useful for us. I think it must not be a problem for other PRs, at least it shouldn't be, because this PR is only removing "dead code" and other PRs should have nothing to do with this.

timovdk:
> I can agree with that, however 2 other PR's will then be on hold as well which includes: Design/control utils/codeclone removal #1060 and a PR that is still coming. Moreover if a PR is open for a very long time then the risk of merge conflicts gets larger.

I don't think it is a good reason to merge this PR because other PRs depend on this one. Besides, if you keep your PR up to date with development you will rarely get merge conflicts/trivially solvable conflicts. As Jordi said, according to you this is dead code, so if there are merge conflicts it would actually prove that the code was not dead.

Besides that, I also agree with Paul and Jordi.

HaicoDorenbos:
> > I can agree with that, however 2 other PR's will then be on hold as well which includes: Design/control utils/codeclone removal #1060 and a PR that is still coming. Moreover if a PR is open for a very long time then the risk of merge conflicts gets larger.
> 
> I don't think it is a good reason to merge this PR because other PRs depend on this one. Besides, if you keep your PR up to date with development you will rarely get merge conflicts/trivially solvable conflicts. As Jordi said, according to you this is dead code, so if there are merge conflicts it would actually prove that the code was not dead.
> 
> Besides that, I also agree with Paul and Jordi.

A merge conflict does not mean that the code was dead. Only if the code does not compile then the code was not dead. Merge conflicts for example could also happen if other developers add more functions to the ControlUtils class. Furthermore you should keep in mind that it takes a lot of time to keep this branch repeatedly up to date with the master for about a month (which is not limited to solving merge conflicts, but also to checking if the code is still compiling and running). Besides even if the code was not dead at all, you should keep in mind that there are still design issues related to these functions, e.g. some of these functions should belong to another class instead, should be better documented, are code clones or should be better tested.

Thus personally in my opinion it is better to merge it now and if someone notice that the functionality is missing then he/she just add this functionality back after checking if this functionality does not exists somewhere else and he/she should add this functionality back at the right place. But since you all want to have this PR on hold I agree to have it on hold, but in case of no objection after a month it should definitely be merged.

PR #1027, opened 2020-04-17
Dealer takes into account keeper

### LINKED ISSUE #1008: Deal keeper role to correct id

jordiluong:
Deal keeper role to robot with keeper id
closes #1008


Yuhanun (approved):
ngl this should be a getter (the keeperId), apart from that LGTM.

PR #1028, opened 2020-04-17
Fix aliasing and framerate

rolfvdhulst:
Makes the interface a lot easier on the eyes. If the framerate is too high, just lower it manually again. I now set it to 50 hz which should be reasonable.


timovdk (approved):
Thanks, this is great

Yuhanun (approved):
LGTM

PR #1037, opened 2020-04-22
Fix/fieldcomputation testcases

### LINKED ISSUE #1034: Fix and document FieldComputations test cases
Currently there are 2 cases documented out with TODO's above them
Also, a step-by-step walkthrough of why the test case is doing what it is doing is needed. This is because we want to understand what each test is testing, so when it fails we know where to look and what is going wrong.

HaicoDorenbos:
The test cases are fixed. In my point of view the documentation is sufficient, because:
- Boundary Value Analysis is put in the comments above the code which clearly indicates why these tests are used and what they are doing.
- The code is quite self-explanatory. For example it is clear to me to which location this code points:
testField.getBottomRightCorner() + Vector2(-1.0 + smallChanges, 1.0 - smallChanges);  
I guess adding documentation will not clarify this any further
- Since in most of the cases there is not even a consensus on the specification of methods, it is a waste of time to spend too much time on documenting the test cases, because if the specification changes the test cases will change as well.

If documentation is needed then a drawing would be more appropriate, but as I indicated above this is not our major priority.

Closes #1034


timovdk (approved):

test/WorldTests/FieldComputationTest.cpp
```diff
@@ -99,13 +99,11 @@ TEST(FieldComputationTest, point_in_polygon) {
     EXPECT_FALSE(FieldComputations::pointIsInField(testField, testPoint, -1.0));
 
     // Pick the corner cases (boundary value analysis) for our defence area in the pointInDefenceArea function
-    testPoint = testField.getLeftPenaltyLineTop() + Vector2(0.3, 0.3);
-
-    // TODO: FIX THIS
-    // EXPECT_TRUE(FieldComputations::pointIsInDefenseArea(testField, testPoint, true, 0.3, 0.2));
+    testPoint = testField.getLeftPenaltyLineTop() + Vector2(0.3 - smallChanges, 0.3 - smallChanges);
+    EXPECT_TRUE(FieldComputations::pointIsInDefenseArea(testField, testPoint, true, 0.3, 0.2));
```
RobotJente, line 107:
```suggestion
    // Check if a point very close but still inside the defense area is correctly evaluated as being in the defense area
    testPoint = testField.getLeftPenaltyLineTop() + Vector2(0.3 - smallChanges, 0.3 - smallChanges);
    EXPECT_TRUE(FieldComputations::pointIsInDefenseArea(testField, testPoint, true, 0.3, 0.2));
```
This way the reader doesn't have to do the mental math and decide if the smallchanges should be added or subtracted, and it's super clear what exactly you are testing, so if the tests break people can fix them more easily

RobotJente (commented):
The fix is fine and solves the first part of the issue. However, the test can be clearer on what it does and why. What does the backmargin have to do with anything? Why did you choose 0.2 and 0.3 as values? Explain what they mean.

RobotJente (changes requested):
Tests need to be documented

test/WorldTests/FieldComputationTest.cpp
HaicoDorenbos: Adding 1 line of comment per 2 lines of code is way too much. As currently explained in the code comments this is a boundary value analysis test case. This is way enough documentation. Also not everything has a reason behind it: 0.2 and 0.3 where just small enough values which is the reason why I used them. But anything different could also have been used. Moreover testing is only important once the specification is clear and I expect these specification to be changed in the near future. Thus spending too much time on test cases is a waste of time, because test cases have to be changed once the specification is changed...

jordiluong:
### **NOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO**

test/WorldTests/FieldComputationTest.cpp
RobotJente: People need to understand what you are testing, and if that takes 2 comments or 0, I don't care. I hope you can see that my suggestion makes it instantly clear, without even needing to look at the code, what is being tested. 

I doubt this specification will be changed, as we are playing robot soccer, and you are checking if the defense area boundaries work well. Unless I missed a drastic rule change, we will have a defense area for quite a while.

Furthermore, if magic numbers are magical, make sure it is known they are magical.

HaicoDorenbos:
> ### **NOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO**

You need to write test cases for test cases :+1:

HaicoDorenbos:
The documentation has been improved...

test/WorldTests/FieldComputationTest.cpp
RobotJente: I think it would really help if you look at the documentation I wrote in the code review. That documentation leaves nothing to the imagination

Ratoone (changes requested):
Looks fine, but I suggest splitting the big test into multiple smaller tests. For instance, checkLeft, checkRight, or something similar

HaicoDorenbos:
I guess that if there are no real objections against this PR then it should soon be merged, since currently a lot of PR's are on hold because it is not yet merged, including: Design/control utils/codeclone removal #1060 and Design/control utils/responsibility shift #1071

PR #1038, opened 2020-04-23
Freedom of our robots invariant

### LINKED ISSUE #1016: Freedom of robots invariant
Are robots free? (with some constant amount of free robots)

jordiluong:
For every robot determine its freedom by using the distance to the closest enemy robot as metric. Invariant is the average of each robots freedom.

closes #1016


src/stp/invariants/FreedomOfRobotsInvariant.cpp
```diff
@@ -0,0 +1,42 @@
...
+    auto& us = world.getUs();
+    std::vector<uint8_t> distanceMetrics{};
+
+    std::transform(us.begin(), us.end(), std::back_inserter(distanceMetrics), [&](auto& robot) {
+        auto robotPosition = robot.get()->getPos();
+        auto distance = (world.getRobotClosestToPoint(robotPosition, world_new::them)->getPos() - robotPosition).length();
+        return calculateMetric(distance);
+    });
+
+    return std::accumulate(distanceMetrics.begin(), distanceMetrics.end(), 0) / distanceMetrics.size();
```
timovdk, line 37: Maybe calculate the average first, and use that value as parameter for calculateMetric(). This way there is only one call to the linear function instead of a maximum of 11 times.

src/stp/invariants/FreedomOfRobotsInvariant.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+uint8_t FreedomOfRobotsInvariant::calculateMetric(const double& x) const noexcept { return piecewiseLinearFunction->yForX(x); }
```
RobotJente, line 40: Could potentially be in base class as behaviour is quite generic. However, not a priority for now

RobotJente (approved):

src/stp/invariants/FreedomOfRobotsInvariant.cpp
RobotJente: Ah smart

RobotJente: I wonder if that works though. I think that requires it to be that f(a) + f(b) = f(a+b) which is not true i think

jordiluong: I also think that wouldn't work that easily. If one robot is really really far, the average distance might be far enough but doesn't say anything about the other robots being free or not. It might work if you clamp the distances first?

timovdk: That is true actually, woops. I think if you clamp you kind of loose the advantage of calling the linear function only once. not sure tho, and seems a bit hacky to clamp that

timovdk (approved):

RobotJente (approved):

PR #1060, opened 2020-04-29
Design/control utils/codeclone removal

HaicoDorenbos:
This PR is the follow-up of Design/control utils/deadcode removal #1026. It removes almost all code clones in the ControlUtils class. All other classes which did still use this functionality will from now on use the Roboteam_utils functionality instead.


HaicoDorenbos:
This PR can only be merged once: Fix bug in ShortestAngleDiff #68 on the Roboteam_utils repository is merged

timovdk (changes requested):
I think your merge went wrong @HaicoDorenbos since I see a lot of new code that has been removed (like the updated ball placements, the updated defend play and the differences in CMakeLists)

PR #1061, opened 2020-04-30
Stp attack play

### LINKED ISSUE #965: Attack Play

jordiluong:
Added roles and invariants to attack play. Roles can still be improved at a later stage

Closes #965

### DIFF
--- a/src/stp/new_plays/Attack.cpp
+++ b/src/stp/new_plays/Attack.cpp
@@ -5,7 +5,6 @@
 #include "stp/invariants/WeHaveBallInvariant.h"
 #include "stp/invariants/GoalVisionInvariant.h"
 #include "stp/invariants/BallCloseToUsInvariant.h"
-#include "stp/invariants/BallIsFreeInvariant.h"
 #include "stp/invariants/game_states/NormalPlayGameStateInvariant.h"
 #include "stp/new_plays/Attack.h"
 #include "stp/new_roles/Attacker.h"
--- a/src/stp/new_plays/Attack.cpp
+++ b/src/stp/new_plays/Attack.cpp
@@ -61,6 +61,7 @@ Dealer::FlagMap Attack::decideRoleFlags() const noexcept {
 }
 
 void Attack::calculateInfoForRoles() noexcept {
+    // TODO: Improve roles
     // Attacker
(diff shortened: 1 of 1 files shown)


RobotJente (approved):

src/stp/new_plays/Attack.cpp
```diff
@@ -3,65 +3,116 @@
...
     startPlayInvariants.emplace_back(std::make_unique<invariant::WeHaveBallInvariant>());
+    startPlayInvariants.emplace_back(std::make_unique<invariant::GoalVisionInvariant>());
 
-    // TODO: decide keep invariants
-/*    keepPlayInvariants.clear();
-    keepPlayInvariants.emplace_back(std::make_unique<invariant::WeHaveBallInvariant>());*/
+    keepPlayInvariants.clear();
+    keepPlayInvariants.emplace_back(std::make_unique<invariant::NormalPlayGameStateInvariant>());
+    keepPlayInvariants.emplace_back(std::make_unique<invariant::BallCloseToUsInvariant>());
+    keepPlayInvariants.emplace_back(std::make_unique<invariant::BallIsFreeInvariant>());
```
timovdk: So this play will only be kept if we don't have the ball? How can we shoot the ball then?

timovdk (changes requested):
Looks good, but I don't think the keep invariants work with the attacker role

src/stp/new_plays/Attack.cpp
jordiluong: Oops hehe

src/stp/new_plays/Attack.cpp
```diff
@@ -3,65 +3,116 @@
 //
 
 #include "stp/invariants/WeHaveBallInvariant.h"
+#include "stp/invariants/GoalVisionInvariant.h"
+#include "stp/invariants/BallCloseToUsInvariant.h"
+#include "stp/invariants/BallIsFreeInvariant.h"
```
jordiluong:
```suggestion
```

src/stp/new_plays/Attack.cpp
jordiluong:
```suggestion
```

PR #67, opened 2019-01-08
Feature/fixtests

rolfvdhulst:
### Summary
Fixes a lot of tests that were broken/not compiling. 

There are still some problems with RobotDealer, the following 6 tests are breaking because they cannot access a robot at robotdealer (after we added ROLEs )
KickTest
GoToPosTest
RotateTest
IHaveBallTest
IsInDefenseAreaTest
IsRobotClosestToBallTest

This is already an issue. A fix is probably not very hard to make.

There are still two other tests which are not working simply because they require you to run GrSim at the same time, and segfault otherwise:
RobotDealerTest
DefendOnRobotTest

### Checklist for the reviewer
This acts as a guide to help you and/or the reviewer to analyze the code. 

**functionality**  
- [ ] The code is properly tested
- [ ] The tests are passing
- [ ] The code works as expected
- [ ] The code is understandable and easy to read
- [ ] Proper abstractions
- [ ] No duplicate code

**style**  
- [ ] Variables/objects/classes have proper names
- [ ] No unused variables
- [ ] No magic numbers
- [ ] The code is well documented
- [ ] No unused imports
- [ ] No leftover code / unused comments
- [ ] Proper indentation
- [ ] Proper use of namespaces

PR #1071, opened 2020-05-04
Design/control utils/responsibility shift

HaicoDorenbos:
The last PR so far related to the ControlUtils class. All responsibilities that should be handled by the Roboteam_utils repository are removed and placed in Roboteam_utils instead. Note that you should only merge this PR, once Design/control utils shift #67 on the Roboteam_utils branch is merged. Moreover Design/control utils/codeclone removal #1060 should also be merged first. And this PR should be simultaneously merged with Fix bug in ShortestAngleDiff #68 on the Roboteam_utils repository.


RobotJente:
It looks like there is a lot of functionality that is being removed from controlutils but not put back in roboteam_utils. I would like to hear what @jordiluong and @rolfvdhulst think about this change

HaicoDorenbos:
> It looks like there is a lot of functionality that is being removed from controlutils but not put back in roboteam_utils. I would like to hear what @jordiluong and @rolfvdhulst think about this change

The functionality is not removed. All dead functions are placed at the bottom of the file. Though the cloned functions are removed.

PR #1075, opened 2020-05-06
Feature/dealer positions refactor

### LINKED ISSUE #1072: Dealer Refactor
Incorporate distance from desired position using the STPInfo's.
After CalculateInfo() do AssignRoles() and give the STPInfos to the dealer to use for distance calculation, probably in getScoreMatrix()

timovdk:
Refactored dealer to also use stpInfo data to make decisions on what robot should get what role.
`distribute`, `getScoreMatrix` and `getScoreForDistance` have new code, all other changes are formatting.
closes #1072

### DIFF
--- a/src/utilities/Dealer.cpp
+++ b/src/utilities/Dealer.cpp
@@ -94,6 +94,8 @@ double Dealer::getScoreForDistance(const stp::StpInfo &stpInfo, const v::RobotVi
         distance = robot->getPos().dist(stpInfo.getPositionToMoveTo().value());
     } else if (stpInfo.getPositionToShootAt().has_value()) {
         distance = robot->getPos().dist(world.getBall()->get()->getPos());
+    } else if (stpInfo.getEnemyRobot().has_value()) {
+        distance = robot->getPos().dist(stpInfo.getEnemyRobot().value()->getPos());
     }
 
     return costForDistance(distance, field->getFieldWidth(), field->getFieldLength());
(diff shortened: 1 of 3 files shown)


RobotJente (approved):

jordiluong (approved):

Ratoone (commented):
Seems fine, only one problem. I will open a different issue for that (see #1080)

src/utilities/Dealer.cpp
```diff
@@ -1,21 +1,28 @@
-#include <roboteam_utils/Hungarian.h>
+/**
+ * The dealer will check for the flags that are set in plays, but also for the distance
+ * to a position that a robot might need to travel to. The lower the score of a robot, the better.
+ * The robot with the lowest score for a role will get assigned to that role, unless it has already been assigned.
+ * In that case it will pick the second best and so on.
```
RobotJente: I don't think this is how the Hungarian works but I could be wrong. This sounds quite greedy, I thought the hungarian tries to find a more global solution

timovdk: Makes sense, my understanding of  the hungarian was wrong

timovdk:
```suggestion
/**
 * The dealer will check for the flags that are set in plays, but also for the distance
 * to a position that a robot might need to travel to. The lower the score of a robot, the better.
```

RobotJente (approved):

PR #1086, opened 2020-05-08
Resetting all roles when re-dealing roles

### LINKED ISSUE #1084: Reset roles after being redealt
Any role that is not dealt by the dealer must be reset

timovdk:
Roles now reset the currentRobot (for interface), tactics and the skills.
closes #1084

### DIFF
--- a/src/stp/Role.cpp
+++ b/src/stp/Role.cpp
@@ -63,7 +63,7 @@ Tactic * Role::getCurrentTactic() {
 void Role::reset() noexcept {
     currentRobot.reset();
 
-    // Reset all skills
+    // Reset all tactics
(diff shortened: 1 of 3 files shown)


RobotJente (approved):

src/stp/Role.cpp
```diff
@@ -60,4 +60,15 @@ Tactic * Role::getCurrentTactic() {
     return robotTactics.get_current();
 }
 
+void Role::reset() noexcept {
+    currentRobot.reset();
+
+    // Reset all skills
```
jordiluong:
```suggestion
    // Reset all tactics
```

jordiluong (approved):

PR #1088, opened 2020-05-08
Free kick them play

### LINKED ISSUE #1046: Free kick them play
Keep distance to ball

jordiluong:
Implementation of free kick them play

Closes #1046


timovdk (approved):

include/roboteam_ai/stp/new_plays/FreeKickThem.h
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Assigns robots to roles of this play
+     */
+    Dealer::FlagMap decideRoleFlags() const noexcept override;
+
+    /**
+     * Calculates info for the roles
+     */
+    void calculateInfoForRoles() noexcept override;
```
HaicoDorenbos, line 38: Uninformative documentation. What kind of info does it calculate for the roles? When should you call this function? What is the reason for this design choice? What are the precondition when calling this function? How does the function work? All unanswered questions which some of them could be answered by adding more documentation.

include/roboteam_ai/stp/new_plays/FreeKickThem.h
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Gets the play name
+     */
+    const char* getName() override;
```
HaicoDorenbos, line 43: Uninformative documentation. What you could do for example in the documentation is explain why every play needs a name and how this name is used. Is the name just used in the UI or is it also a kind of identifier to distinguish different plays.

include/roboteam_ai/stp/new_plays/FreeKickThem.h
```diff
@@ -0,0 +1,72 @@
...
+//
+
+#ifndef RTT_FREEKICKTHEM_H
+#define RTT_FREEKICKTHEM_H
+
+#include <stp/Play.hpp>
+
+namespace rtt::ai::stp::play {
+
+class FreeKickThem : public Play {
```
HaicoDorenbos, line 12: Class documentation is missing. since and author tags are still missing. It is important to explain in the class documentation what this play does and why this play is needed, etc.

HaicoDorenbos (changes requested):
Documentation at a lot of places is still quite uninformative and insufficient. And documentation is very important for code. To illustrate this: airplanes are worth halve of their value when a logbook is missing, so a logbook (just some pieces of paper) have the same value as the airplane itself. In case of developing software, the documentation carriers even more value than the code itself.

src/stp/new_plays/FreeKickThem.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+void FreeKickThem::calculateInfoForBlockers() noexcept {
+    const double MINIMAL_DISTANCE_FROM_BALL = 0.5;
+    const int NUMBER_OF_BLOCKERS = 6;
+
+    auto enemyRobots = world->getWorld()->getThem();
+
+    // We cannot block enemy robots that are too close to the ball
+    enemyRobots.erase(std::remove_if(enemyRobots.begin(), enemyRobots.end(), [&] (const auto enemyRobot) -> bool {
+        return enemyRobot->getDistanceToBall() <= 1.5 * MINIMAL_DISTANCE_FROM_BALL;
```
HaicoDorenbos: Why is it 1.5 times the Minimal Distance From Ball? Why not 2.0 times or 3.0 times or 1.25 times? It is better to make 1.5 a constant variable and explain the meaning of this variable.

src/stp/new_plays/FreeKickThem.cpp
```diff
@@ -0,0 +1,129 @@
...
+    calculateInfoForOffenders();
+}
+
+void FreeKickThem::calculateInfoForKeeper() noexcept {
+    stpInfos["keeper"].setEnemyRobot(world->getWorld()->getRobotClosestToBall(world_new::them));
+    stpInfos["keeper"].setPositionToShootAt(Vector2());
+}
+
+void FreeKickThem::calculateInfoForBlockers() noexcept {
+    const double MINIMAL_DISTANCE_FROM_BALL = 0.5;
```
HaicoDorenbos: What is the measure used for this variable? Feet, cm, cows, ...? And is it relative or absolute?

timovdk: I think cows

include/roboteam_ai/stp/new_plays/FreeKickThem.h
RobotJente: This function is pure virtual and documented in the base class

include/roboteam_ai/stp/new_plays/FreeKickThem.h
timovdk: I don't think it is relevant to list usages for every function above that function, especially for a simple getter function. Getters and setters are kind of trivial

include/roboteam_ai/stp/new_plays/FreeKickThem.h
RobotJente: I think the naming makes it clear enough what the play is doing

src/stp/new_plays/FreeKickThem.cpp
RobotJente: It could be good to make some standard for distances. I always assume they are in meters, because that makes the most sense. If that is always the case, we can add this to the readme for AI. I know in Roboteam_World many things are in cm or mm suddenly. But this is not necessary to do before the 10th of June so I'll make a long-term issue out of it

PR #1094, opened 2020-05-11
Design/control utils/codeclone removal2

HaicoDorenbos:
This PR is the same as Design/control utils/codeclone removal #1060, however all unrelated changes are removed. All code clones in ControlUtils are removed by merging this PR.


src/control/ControlUtils.cpp
```diff
@@ -161,13 +34,13 @@ double ControlUtils::angleDifference(double A1, double A2) {
         angleDif -= 2 * M_PI;
     }
     return fabs(angleDif);
-}
+} // Code clone
```
timovdk, line 37: Should this comment be here or is this still a pending remove of a code clone?

src/control/ControlUtils.cpp
```diff
@@ -219,29 +92,46 @@ Vector2 ControlUtils::accelerationLimiter(const Vector2 &targetVel, const Vector
...
-    if (denominator != 0) {
-        double numerator = ((a1.x - b1.x) * (b1.y - b2.y) - (a1.y - b1.y) * (b1.x - b2.x));
-        double t = numerator / denominator;
-        return (a1 + (Vector2){t, t} * (a2 - a1));
-    } else {
-        return Vector2();
+/// Calculate the force of a given vector + a certain type.
+/// the basic formula is: force = weight/distance^2 * unit vector
+Vector2 ControlUtils::calculateForce(const Vector2 &vector, double weight, double minDistance) {
+    // if the object is close enough, it's forces should affect. Otherwise don't change anything.
```
timovdk, line 98: affect what?

src/control/ControlUtils.cpp
```diff
@@ -219,29 +92,46 @@ Vector2 ControlUtils::accelerationLimiter(const Vector2 &targetVel, const Vector
...
-    double denominator = ((a1.x - a2.x) * (b1.y - b2.y) - (a1.y - a2.y) * (b1.x - b2.x));
-    if (denominator != 0) {
-        double numerator = ((a1.x - b1.x) * (b1.y - b2.y) - (a1.y - b1.y) * (b1.x - b2.x));
-        double t = numerator / denominator;
-        return t;
-    } else {
-        return -1.0;
+
+bool ControlUtils::clearLine(const Vector2 &fromPos, const Vector2 &toPos, const world_new::view::WorldDataView world, double safeDistanceFactor, bool includeKeeper) {
+    double minDistance = Constants::ROBOT_RADIUS() * safeDistanceFactor;
```
timovdk, line 122: We should not use the old constants anymore but i'm not sure (@RobotJesse ?)

timovdk (commented):
Looks good but I have some questions, see comments.

src/control/ControlUtils.cpp
RobotJente: Yeah constants should not be used anymore

src/control/ControlUtils.cpp
HaicoDorenbos: That documentation was not added by me, some parts of the code have been moved to another location...

src/control/ControlUtils.cpp
HaicoDorenbos: That is the reason why dead code should be removed, because we currently are more busy discussing it than it would take to remove it instead.

src/control/ControlUtils.cpp
HaicoDorenbos: This is indeed a code clone, but since it interferes with functions in ControlUtils that are not code clones I decided to remove it in the following up PR (which is easier than to remove it here).

HaicoDorenbos:
So can we merge it?

Yuhanun (approved):
LGTM apart from constants, I'm pro-removal of code clone removal :)

timovdk (approved):

PR #1095, opened 2020-05-11
Changed goal vision invariant

jordiluong:
Changed goal vision calculation to use the position of the ball instead of the positions of the robots.

### DIFF
--- a/src/stp/invariants/GoalVisionInvariant.cpp
+++ b/src/stp/invariants/GoalVisionInvariant.cpp
@@ -26,15 +26,7 @@ GoalVisionInvariant::GoalVisionInvariant() noexcept {
 }
 
 uint8_t GoalVisionInvariant::metricCheck(world_new::view::WorldDataView world, const world::Field* field) const noexcept {
-    auto& us = world.getUs();
-    std::vector<double> visibilities{};
-    visibilities.reserve(control_constants::MAX_ROBOT_COUNT);
-
-    std::transform(us.begin(), us.end(), std::back_inserter(visibilities), [&](auto& robot) {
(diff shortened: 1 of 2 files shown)


src/stp/invariants/GoalVisionInvariant.cpp
```diff
@@ -26,15 +26,7 @@ GoalVisionInvariant::GoalVisionInvariant() noexcept {
...
 uint8_t GoalVisionInvariant::metricCheck(world_new::view::WorldDataView world, const world::Field* field) const noexcept {
-    auto& us = world.getUs();
-    std::vector<double> visibilities{};
-    visibilities.reserve(control_constants::MAX_ROBOT_COUNT);
-
-    std::transform(us.begin(), us.end(), std::back_inserter(visibilities), [&](auto& robot) {
-        return FieldComputations::getPercentageOfGoalVisibleFromPoint(*field, false, robot.get()->getPos(), world, robot.get()->getId(), true);
-    });
-
-    return calculateMetric(*std::min_element(visibilities.begin(), visibilities.end()));
```
RobotJente, line 37: This means now the invariant checks whether the ball can see the goal, rather than the robots, right?

jordiluong: Yes, because we assume the attacker already has the ball. We can also use the position of the closest robot to ball

timovdk: Maybe using the ball is a better way of judging our scoring opportunity, but the distance from our closest bot to the ball kind of decides whether it is any good...

PR #1110, opened 2020-05-14
Feature/improved harasser

### LINKED ISSUE #1100: Update Harasser role
- [x] Rename existing BlockBall for keeper role
- [x] Make new BlockBall tactic for normal bots
- [x] Update Harasser role with the new BlockBall
- [ ] Decide harass position in different scenarios

timovdk:
Harasser now blocks the ball instead of a robot
closes #1100

### DIFF
--- a/src/stp/new_tactics/BlockBall.cpp
+++ b/src/stp/new_tactics/BlockBall.cpp
@@ -67,6 +67,8 @@ Vector2 BlockBall::calculateTargetPosition(const world_new::view::BallView &ball
 
         auto ballCircle = Circle(ball->getPos(), BLOCK_DISTANCE);
 
+        // Project the enemy robot on the circle, and rotate this position with Pi.
+        // This is to stand directly opposite the enemy robot with the ball in the middle.
         targetPosition = ballCircle.project(enemyRobot->getPos()).rotateAroundPoint(M_PI, ballCircle.center);
     }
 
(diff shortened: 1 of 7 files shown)


src/stp/new_tactics/BlockBall.cpp
```diff
@@ -53,65 +53,46 @@ bool BlockBall::shouldTacticReset(const StpInfo &info) noexcept {
...
+
+    // Ball is moving
+    // Intercept ball when it is moving
+    else if (ball->getVelocity().length() > control_constants::BALL_STILL_VEL) {
+        // TODO: Tune this distance
+        const double BLOCK_DISTANCE(ball->getVelocity().length() * 0.5);
+
+        auto ballCircle = Circle(ball->getPos(), BLOCK_DISTANCE);
+
+        targetPosition = ballCircle.project(enemyRobot->getPos()).rotate(M_PI) + ballCircle.center + ballCircle.center;
```
RobotJente: The vector stuff here needs to be explained a bit better in my opinion. It is not clear to me how you are intercepting the ball. It looks like you are projecting their robot on the ballcircle, but why are you adding the ballCircle.center twice? 
I'm also not sure if this is the right place to do these calculations, maybe this should be made into a generalized helper function for calculating intercept positions (there is already something like this in controlutils I think)

timovdk: I removed the intercept code since this should be handled by a different tactic.

RobotJente (approved):

jordiluong (approved):

PR #1128, opened 2020-05-20
General cleanup

Yuhanun:
Did some general cleanup.

Removed a .h and .cpp file that were unused but I now realize they might just be for the future. If you want me to add it back i can do that real quick :)


HaicoDorenbos (commented):
Cleaning up the code is always important to do :+1:

src/world/FieldComputations.cpp
```diff
@@ -179,8 +175,8 @@ std::vector<LineSegment> FieldComputations::mergeBlockades(std::vector<LineSegme
     /* If two blockades intersect (in this case, overlap), we take the beginning of the first obstacle and the end of the second obstacle, and put them back in the front of the
      * obstacles vector. The second element gets erased. If they don't intersect, try the next two obstacles. Repeat this procedure until no overlaps are left. */
     std::sort(blockades.begin(), blockades.end(), [](const LineSegment &a, const LineSegment &b) { return a.start.y < b.start.y; });
-    int iterator = 0;
-    while (iterator < static_cast<int>(blockades.size()) - 1) {
```
HaicoDorenbos, line 183: Be careful with these type of things. It often happens in the code that variables values get forgotten and for these situations a static_cast<int> is used.

src/stp/new_roles/PenaltyKeeper.cpp
```diff
@@ -20,7 +20,7 @@ PenaltyKeeper::PenaltyKeeper(std::string name) : Role(std::move(name)) {
 Status PenaltyKeeper::update(StpInfo const& info) noexcept {
     // Failure if the required data is not present
     if (!info.getBall() || !info.getRobot() || !info.getField()) {
-        RTT_WARNING("Required information missing in the tactic info");
+        RTT_WARNING("Required information missing in the tactic info")
```
HaicoDorenbos, line 23: Huh? Every statement should be closed by a semicolon, right?

src/stp/new_roles/Keeper.cpp
```diff
@@ -19,7 +19,7 @@ Keeper::Keeper(std::string name) : Role(std::move(name)) {
 Status Keeper::update(StpInfo const& info) noexcept {
     // Failure if the required data is not present
     if (!info.getBall() || !info.getRobot() || !info.getField()) {
-        RTT_WARNING("Required information missing in the tactic info");
+        RTT_WARNING("Required information missing in the tactic info")
```
HaicoDorenbos, line 22: Why did you remove the semicolon?

HaicoDorenbos (changes requested):
Semicolons are removed at some places, where I wonder whether they should be removed. Also make sure that the code still compiles and that all checks in Travis pass.

src/stp/new_roles/Keeper.cpp
rolfvdhulst: It's a macro

timovdk (changes requested):
The implementations of the functions removed from ControlUtils.h are still in ControlUtils.cpp

src/world/FieldComputations.cpp
Yuhanun: Well static_cast<int>(size_t) can cause overflow.

Yuhanun: Well static_cast<int>(size_t) can cause overflows so yeah

src/stp/new_roles/PenaltyKeeper.cpp
Yuhanun: It's a macro defined to already contain the `;`

Yuhanun (commented):
> The implementations of the functions removed from ControlUtils.h are still in ControlUtils.cpp

will fix that soonTM, been busy again sadly

HaicoDorenbos (approved):
Good job :+1:, removing and cleaning up code is often a very boring job, but really important since the code is more understandable and therefore reduces the risk of bugs, increases the efficiency of the code, decrease maintenance time, makes the system more suitable for evolution , etc. :+1:

However make sure to compile the code first before merging it and check if all tests in Travis pass.

Yuhanun:
Issue it's having is because of roboteam_utils, either hasn't been merged yet or ye.

Yuhanun:
95% confident this is because roboteam_utils is supposed to be on another branch btw ;)

timovdk (approved):

PR #1130, opened 2020-05-27
Improvements to play progression

timovdk:
Added no ball vision invariant
various dealerflag and invariant improvements in plays
get ball now avoids defense area (we can't touch ball inside defense area)
Fixed dealer to deal required and keeperflags correctly

### DIFF
--- a/include/roboteam_ai/stp/invariants/NoGoalVisionFromBallInvariant.h
+++ b/include/roboteam_ai/stp/invariants/NoGoalVisionFromBallInvariant.h
@@ -1,0 +1,38 @@
+//
+// Created by timovdk on 5/27/20.
+//
+
+#ifndef RTT_NOGOALVISIONFROMBALLINVARIANT_H
+#define RTT_NOGOALVISIONFROMBALLINVARIANT_H
+
+#include <NFParam/Param.h>
+
+#include "BaseInvariant.h"
+
+namespace rtt::ai::stp::invariant {
+/**
+    Calculates goal vision percentage based on the position of the ball
+**/
+class NoGoalVisionFromBallInvariant : public BaseInvariant {
+   public:
+    NoGoalVisionFromBallInvariant() noexcept;
+
+    [[nodiscard]] uint8_t metricCheck(world_new::view::WorldDataView world, const world::Field* field) const noexcept override;
+
+   private:
+    /**
+     * Calculates the actual metric value using the piecewise linear function member
+     * @param x the x of the function
+     * @return metric value between 0-255
+     */
+    [[nodiscard]] uint8_t calculateMetric(const double& x) const noexcept;
+
+    /**
+     * Unique pointer to the piecewise linear function that calculates the fuzzy value
+     */
+    std::unique_ptr<nativeformat::param::Param> piecewiseLinearFunction;
+};
+
+}  // namespace rtt::ai::stp::invariant
+
+#endif  // RTT_NOGOALVISIONFROMBALLINVARIANT_H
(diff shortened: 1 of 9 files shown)


HaicoDorenbos:
> _No description provided._

Very clear what you did...

timovdk:
yes indeed, the PR title is quite descriptive if I might say so myself.
EDIT: I added a description

include/roboteam_ai/stp/invariants/NoGoalVisionFromBallInvariant.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <NFParam/Param.h>
+
+#include "BaseInvariant.h"
+
+namespace rtt::ai::stp::invariant {
+/**
+    Calculates goal vision percentage based on the position of the ball
+**/
+class NoGoalVisionFromBallInvariant : public BaseInvariant {
```
HaicoDorenbos, line 16: This is why documentation is useful, because I have the feeling that the documentation does not match the implementation. The class is called NoGoalVision... whereas in the documentation you describe that you compute the percentage of goal vision (and not the percentage of goal that is not visible)...

HaicoDorenbos: Also make sure to add the date tag when the class is created and the author tag who created this class. Moreover documentation always ends with */ instead of **/.

include/roboteam_ai/stp/invariants/NoGoalVisionFromBallInvariant.h
```diff
@@ -0,0 +1,38 @@
...
+   public:
+    NoGoalVisionFromBallInvariant() noexcept;
+
+    [[nodiscard]] uint8_t metricCheck(world_new::view::WorldDataView world, const world::Field* field) const noexcept override;
+
+   private:
+    /**
+     * Calculates the actual metric value using the piecewise linear function member
+     * @param x the x of the function
+     * @return metric value between 0-255
```
HaicoDorenbos, line 26: What does this metric value represent? That is currently unclear from your description.

src/stp/new_plays/AttackingPass.cpp
```diff
@@ -53,12 +60,14 @@ Dealer::FlagMap AttackingPass::decideRoleFlags() const noexcept {
...
     flagMap.insert({"test_role_8", {not_important}});
     flagMap.insert({"test_role_9", {not_important}});
-    flagMap.insert({"test_role_10", {not_important}});
 
     return flagMap;
 }
 
 void AttackingPass::calculateInfoForRoles() noexcept {
+    stpInfos["keeper"].setEnemyRobot(world->getWorld()->getRobotClosestToBall(world_new::them));
+    stpInfos["keeper"].setPositionToShootAt(Vector2());
```
HaicoDorenbos, line 69: Huh? You shoot to an undefined position?

HaicoDorenbos (dismissed):
Documentation is inconsistent with the naming of your class. Also the documentation could be improved at some places. Moreover you shoot to an undefined position...

src/stp/new_plays/AttackingPass.cpp
timovdk: Nope, this is a default constructed Vector2, which is (0,0), so absolutely not undefined. This is temporary, but changing this to a position that makes more sense is not for this PR but a future one.

include/roboteam_ai/stp/invariants/NoGoalVisionFromBallInvariant.h
timovdk: If you look at the documentation in the base class, it might become clearer

include/roboteam_ai/stp/invariants/NoGoalVisionFromBallInvariant.h
timovdk: Date and author tag is already there, but thanks. BTW, it's not documentation that ends with **/ but a comment block :)

rttwente (approved):

PR #1132, opened 2020-05-28
Fix/handle lack off data in tactics

timovdk:
If some data is not present anymore, AI segfaults. This PR fixes that issue


Ratoone (approved):
Boring PR

HaicoDorenbos (dismissed):
When changing the code you should also update the documentation and explain that this happens due to segfaults...

rttwente (approved):

PR #1134, opened 2020-06-01
Feature/stp reflect kick

timovdk:
Reflect kick


src/stp/new_plays/ReflectKick.cpp
```diff
@@ -0,0 +1,132 @@
...
+    // TODO: Change this
+    auto passPosition = Vector2();
+
+    auto ball = world->getWorld()->getBall().value();
+    std::optional<Vector2> intersection;
+
+    if ((ball->getPos() - passPosition).length() <= 2.0 && ball->getVelocity().length() > 0.1) {
+        intersection = Line(ball->getPos(), ball->getPos() + ball->getVelocity()).intersects(
+                Line(passPosition, field.getTheirGoalCenter()));
+    }
```
rolfvdhulst: This will be problematic if the ball ever rolls towards our goal, there will still be an intersection.

src/stp/new_plays/ReflectKick.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    auto reflectPosition = intersection.has_value() ? intersection.value() : passPosition;
+
+    reflectPosition = field.getTheirGoalCenter() + (reflectPosition - field.getTheirGoalCenter()).stretchToLength((reflectPosition - field.getTheirGoalCenter()).length() + control_constants::CENTER_TO_FRONT);
```
rolfvdhulst: This line is way too long. Just cut it up into multiple lines.

rolfvdhulst: Especially vector calculus should be easy to read.

rolfvdhulst: You can even consider making this a function

src/stp/new_plays/ReflectKick.cpp
rolfvdhulst: Lines are infinite

src/stp/new_skills/Kick.cpp
```diff
@@ -15,13 +15,15 @@ Status Kick::onUpdate(const StpInfo &info) noexcept {
...
+    command.set_chip_kick_forced(true);
 
     // Set angle command
     command.set_w(info.getRobot().value()->getAngle());
 
     publishRobotCommand();
 
-    if (info.getBall()->get()->getVelocity().length() > stp::control_constants::HAS_KICKED_ERROR_MARGIN) {
+    if (info.getBall()->get()->getVelocity().length() > stp::control_constants::HAS_KICKED_ERROR_MARGIN &&
+            info.getRobot().value()->getAngleDiffToBall() <= control_constants::GO_TO_POS_ANGLE_ERROR_MARGIN * M_PI) {
```
rolfvdhulst, line 26: This seems like a pretty unrelated change. How does Go TO POS affect the kick skill?

src/stp/new_tactics/KickAtPos.cpp
```diff
@@ -84,7 +84,7 @@ double KickAtPos::determineKickForce(double distance, KickChipType desiredBallSp
 
 bool KickAtPos::isEndTactic() noexcept {
     // This is not an end tactic
-    return false;
+    return true;
```
rolfvdhulst: Are you really sure this does not break other tactics?

rolfvdhulst: Atleast, change the documentation

rolfvdhulst (changes requested):
See comments. IMO, most things look neat but don't put a lot of vector calculus on one line because it makes things very hard to read/understand for others

rttwente (approved):

include/roboteam_ai/stp/new_plays/ReflectKick.h
```diff
@@ -0,0 +1,51 @@
...
+//
+
+#ifndef RTT_REFLECTKICK_H
+#define RTT_REFLECTKICK_H
+
+#include "stp/Play.hpp"
+
+namespace rtt::ai::stp::play {
+
+class ReflectKick : public Play {
```
HaicoDorenbos, line 12: In most cases you do not document overriden methods if they a properly documented in the superclass. Except of course if the behaviour really changes. But make sure that class documentation is not missing.

include/roboteam_ai/stp/new_roles/BallReflecter.h
```diff
@@ -0,0 +1,23 @@
...
+//
+
+#ifndef RTT_BALLREFLECTER_H
+#define RTT_BALLREFLECTER_H
+
+#include "stp/Role.hpp"
+
+namespace rtt::ai::stp::role {
+
+class BallReflecter : public Role {
```
HaicoDorenbos, line 12: Class documentation is missing

include/roboteam_ai/stp/new_tactics/PositionAndAim.h
```diff
@@ -0,0 +1,63 @@
...
+//
+
+#ifndef RTT_POSITIONANDAIM_H
+#define RTT_POSITIONANDAIM_H
+
+#include "stp/Tactic.h"
+
+namespace rtt::ai::stp::tactic {
+
+class PositionAndAim : public Tactic {
```
HaicoDorenbos, line 12: Where is the class documentation?

include/roboteam_ai/stp/new_tactics/PositionAndAim.h
```diff
@@ -0,0 +1,63 @@
...
+public:
+    PositionAndAim();
```
HaicoDorenbos, line 14: This method should properly be documented.

src/stp/new_plays/ReflectKick.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+void ReflectKick::calculateInfoForRoles() noexcept {
+    // TODO: Change this
+    auto passPosition = Vector2();
+
+    auto ball = world->getWorld()->getBall().value();
+    std::optional<Vector2> intersection;
+
+    if ((ball->getPos() - passPosition).length() <= 2.0 && ball->getVelocity().length() > 0.1) {
+        intersection = Line(ball->getPos(), ball->getPos() + ball->getVelocity()).intersects(Line(passPosition, field.getTheirGoalCenter()));
```
HaicoDorenbos, line 74: Make sure that you select the right line type. Consider using LineSegment or HalfLine instead:
- Line: Has infinite length in both directions.
- HalfLine: Has a starting point and infinite length in one direction.
- LineSegment: Has a start and end point, which means it has finite length

src/stp/new_plays/ReflectKick.cpp
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    // Passer
+    stpInfos["passer"].setPositionToShootAt(passPosition);
+    stpInfos["passer"].setKickChipType(TARGET);
+
+    // Offenders
+    stpInfos["offender_1"].setPositionToMoveTo(Vector2(field.getFieldLength() / 4, field.getFieldWidth() / 4));
+    stpInfos["offender_2"].setPositionToMoveTo(Vector2(field.getFieldLength() / 4, -field.getFieldWidth() / 4));
```
HaicoDorenbos, line 103: This is not the way how to encode positions in the field. Using this technique you will hard code coordinates. Consider making a constant for this location stored in the Field class or consider to create a method in the FieldComputation class that returns relative positions.

src/stp/new_plays/ReflectKick.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    // Midfielders
+    stpInfos["midfielder_1"].setPositionToMoveTo(Vector2(0.0, field.getFieldWidth() / 4));
+    stpInfos["midfielder_2"].setPositionToMoveTo(Vector2(0.0, -field.getFieldWidth() / 4));
+    stpInfos["midfielder_3"].setPositionToMoveTo(Vector2(field.getFieldLength() / 8, 0.0));
```
HaicoDorenbos, line 108: Do not hard code coordinates this way. It makes things a lot less clear and enforces you to use a certain coordinate system.

HaicoDorenbos (changes requested):
I cannot accept this PR in this current state, because: 
- A lot of documentation is missing. This is mostly related to class documentation, which is really important, since other software developers can then better understand your code and test is properly.
- At some places there is unnecessary documentation. You should not document overridden methods if that method in the super class is properly documented except if the behaviour of that method does not comply anymore with the specification in the super class (usually a bad practice to have these kind of cases).
- Some coordinates get hard coded. Special locations should be defined in the Field class. Also if you want to get a relative position on the Field then you should create a method in the FieldComputation class for this.
- Be consistent where you use Line, LineSegment or HalfLine. And know the differences between these classes. 

During the team meeting on Wednessday I noticed that the preventive maintenance tasks of earlier PR's will be skipped, because some tasks are delayed and will therefore be worked on after the deadline which means that there is no time left for the preventive maintenance tasks of earlier PR's. Therefore PR's should only be merged once the preventive maintenance is also completed for that PR, otherwise you will create technical debt for future teams.

PR #1136, opened 2020-06-02
Fix pass

timovdk:
fixing receivers

### DIFF
--- a/src/stp/new_plays/AttackingPass.cpp
+++ b/src/stp/new_plays/AttackingPass.cpp
@@ -27,12 +27,10 @@ AttackingPass::AttackingPass() : Play() {
 
     keepPlayInvariants.clear();
     keepPlayInvariants.emplace_back(std::make_unique<invariant::NormalPlayGameStateInvariant>());
-    keepPlayInvariants.emplace_back(std::make_unique<invariant::BallMovesSlowInvariant>());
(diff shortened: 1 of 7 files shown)


src/stp/new_plays/AttackingPass.cpp
```diff
@@ -77,8 +77,20 @@ void AttackingPass::calculateInfoForRoles() noexcept {
...
     const Vector2 passingPosition = calculatePassLocation();
 
+    auto ball = world->getWorld()->getBall().value();
+    std::optional<Vector2> intersection;
+
+    if ((ball->getPos() - passingPosition).length() <= 2.0 && ball->getVelocity().length() > 0.1) {
+        intersection = Line(ball->getPos(), ball->getPos() + ball->getVelocity()).intersects(
+            Line(passingPosition, field.getTheirGoalCenter()));
+    }
+
```
RobotJente: If I understood correctly, this code assumes the ball will travel over the line between the passposition and their goal. This is not necessarily true, we might want a failsafe for if the shot was much softer than what we expected

RobotJente (commented):
Maybe we want some failsafe for if the pass was much softer than expected

src/stp/new_plays/AttackingPass.cpp
timovdk: true

rttwente (approved):

PR #1139, opened 2020-06-10
hotfix pls merge

timovdk:


Ratoone (approved):

Ratoone:
I rate the commit messages 5/7

src/world/FieldComputations.cpp
```diff
@@ -175,8 +175,8 @@ std::vector<LineSegment> FieldComputations::mergeBlockades(std::vector<LineSegme
     /* If two blockades intersect (in this case, overlap), we take the beginning of the first obstacle and the end of the second obstacle, and put them back in the front of the
      * obstacles vector. The second element gets erased. If they don't intersect, try the next two obstacles. Repeat this procedure until no overlaps are left. */
     std::sort(blockades.begin(), blockades.end(), [](const LineSegment &a, const LineSegment &b) { return a.start.y < b.start.y; });
-    size_t iterator = 0;
-    while (iterator < blockades.size() - 1) {
+    int iterator = 0;
+    while (iterator < static_cast<int>(blockades.size()) - 1) {
```
rolfvdhulst, line 179: what?

rolfvdhulst: Why not use an normal, iterator, you know?

timovdk: because we use .at() and we add/subtract to compare different entries of the vector. The actual issue we had was that size_t is unsigned, so if blockades.size() returns 0 and you subtract 1, you get like 5 million or smth

PR #1140, opened 2020-06-10
Fixed a bug when the robot would not follow a moving ball

Ratoone:
Fixed a bug where the robot was not following a moving ball, due to mistakenly considering it an obstacle


timovdk (approved):
LGTM

PR #1143, opened 2020-06-16
Fixed segfault in getBallRisky

### LINKED ISSUE #1142: Segfault in GetBallRisky.cpp
**Describe the bug**
Roboteam_AI segfaults in GetBallRisky.cpp, line 49 on branch development.

**To Reproduce**
Run with referee and one robot. I clicked force start and it repeatedly happened.

**Expected behavior**
Code that does not segfault?

On ubuntu 20.04

timovdk:
There was a segfault if there were no enemy robots, this PR will fix that for multiple plays. 
closes #1142


rolfvdhulst (commented):
LGTM. I'll test later if this prevents the segfaults I mentioned before.

RobotJente (approved):
I think we need an enemyrobots invariant

PR #1147, opened 2020-06-23
World/no trust

Yuhanun:
World now no longer trusts anyone that instance()'s it, it now REQUIRES a mutex to be locked which gets returned in an Acquire structure. 

No more need to mutex on a low level becuase mutexing is enforced on a high level.


timovdk:
It seems like the lock locks forever in paintEvent. when ai gets a world, and it tries to run the first update, it gets stuck on this line. (That stacktrace is found by using the pause execution fucntionality from the debugger)
Screenshot from 2020-06-23 14-37-53
Screenshot from 2020-06-23 14-37-35
Other thread:
Screenshot from 2020-06-23 14-54-52
Another thread:
Screenshot from 2020-06-23 14-56-53

timovdk (approved):
LGTM

PR #1151, opened 2020-06-24
Fixed a deadlock when clicking the pause button in the interface

timovdk:
When clicking the pause button, there was a deadlock because of world mutexes


Yuhanun (approved):
LGTM

PR #1155, opened 2020-06-27
Working on knowledge transfer

Yuhanun:
This is now completed.

(better than a class diagram ;) )


wiki/World.md
```diff
@@ -0,0 +1,142 @@
...
+
+`updateTickTime` -> Updates `lastTick` and `tickDuration`
+
+`setWorld` -> Sets `currentWorld` to the parameter.
+
+`toHistory` -> Pushes a `world` to the history vector.
+
+## Advice for next year ##
+The world is really nice right now, it properly abstracts the hard parts away.
+It tries to guarantee shared immutability which is a very complicate topic and will be covered
```
timovdk:
```suggestion
It tries to guarantee shared immutability which is a very complicated topic and will be covered
```

HaicoDorenbos:
Looking good, documentation of the Wiki is indeed very important :+1:

RobotJente:
I've got some general advice as well. 

I think certain members should be pre-initialized on construction of the world. For example, getRobotClosestToBall, whichRobotHasBall, and things like that are used every tick, and it's computationally faster to set this as a member variable rather than run the calculation each time. Maybe something like this:
```
World::World(const& World) {
     setRobotClosestToBall(calculateRobotClosestToBall());
     ...
}
```

wiki/WorldData.md
```diff
@@ -0,0 +1,121 @@
...
+ * RobotViewArray, or you could of course just have a second std::array<RobotView, N> 
+ * to which you return slices.
+ */
+std::span<RobotView> getOurs() {
+    return std::span(&*robots_views.begin(), &*robots_views.begin() + ROBOT_COUNT);
+}
+
+std::span<RobotView> getTheirs() {
+    return std::span(&*robots_views.begin() + ROBOT_COUNT, &*robots_views.end());
+}
```
RobotJente:
```suggestion
std::span<RobotView> getUs() {
    return std::span(&*robots_views.begin(), &*robots_views.begin() + ROBOT_COUNT);
}

std::span<RobotView> getThem() {
    return std::span(&*robots_views.begin() + ROBOT_COUNT, &*robots_views.end());
}
```

RobotJente: I think it's important to keep consisting naming

Yuhanun:
> I've got some general advice as well.
> 
> I think certain members should be pre-initialized on construction of the world. For example, getRobotClosestToBall, whichRobotHasBall, and things like that are used every tick, and it's computationally faster to set this as a member variable rather than run the calculation each time. Maybe something like this:
> 
> ```
> World::World(const& World) {
>      setRobotClosestToBall(calculateRobotClosestToBall());
>      ...
> }
> ```

depends on how often it's called, but yes. ill add it soon.

Yuhanun:
> I've got some general advice as well.
> 
> I think certain members should be pre-initialized on construction of the world. For example, getRobotClosestToBall, whichRobotHasBall, and things like that are used every tick, and it's computationally faster to set this as a member variable rather than run the calculation each time. Maybe something like this:
> 
> ```
> World::World(const& World) {
>      setRobotClosestToBall(calculateRobotClosestToBall());
>      ...
> }
> ```

done.

wiki/World.md
```diff
@@ -0,0 +1,140 @@
+# World #
+
+## General ##
+[World](https://github.com/RoboTeamTwente/roboteam_ai/blob/development/include/roboteam_ai/world_new/World.hpp#L36) by itself is a very interesting conversation, because we don't just have a single world.
+We have two worlds, one called [roboteam_world](http://github.com/roboteamtwente/roboteam_world), and one that we refer 
+to as [ai world](https://github.com/RoboTeamTwente/roboteam_ai/blob/development/include/roboteam_ai/world_new/World.hpp)
```
rolfvdhulst, line 6: Maybe unrelated, but I couldn't help when reading this to feel that one of these two things needs a rename to avoid confusion. Even during the discussion today Jesse was confused. So I would make this part of our advice.

Yuhanun: very true, we should. Roboteam Data maybe?

Yuhanun: Roboteam (pre)process?

Yuhanun:
This has been completed. no longer draft.

wiki/Robot.md
```diff
@@ -0,0 +1,52 @@
+# Robot #
+
+## General ##
+The structure that represents a robot, it represents the type of robot, the team it's on, and some constants. POD.
```
timovdk: It does not represent type anymore
```suggestion
The structure that represents a robot, it represents the team it's on, and some constants. POD.
```

wiki/WorldData.md
```diff
@@ -0,0 +1,131 @@
...
+namespace std {
+  template <typename T>
+  struct span {
+      T* begin;
+      T* end;
+  };
+}
+
+// Wow look at that proficiency at C++ how 
+// am I not hired at GCC yet.
```
timovdk, line 70: nice

wiki/WorldData.md
```diff
@@ -0,0 +1,131 @@
...
+## Advice ##
+Oh boy do I have a lot of advice for this one.
+
+* Rework the `robots` vector to be an `std::array` instead
+* Rework the `robotsNonOwning` vector to be an `std::array` instead
+* Rework `robotsNonOwning` to be a sorted array of _all_ the robots, not just ours or theirs. Sorted by team.
+* Rework `getUs` and `getThem` to return an `std::span` of the robots in the `robotsNonOwning` array.
+
+---
+### Optional advice ###
```
timovdk, line 52: Interesting advice, thanks!

Yuhanun:
This is ready to be merged.

timovdk (approved):
LGTM

RobotJente (approved):

PR #1157, opened 2020-07-03
Wiki: Preventive Maintenance Advice next year

HaicoDorenbos:
WIP (Work In Progress not Werkgroep Infectie Preventie). But you can already review the things I have written so far.

This PR is about the Preventive Maintenance Advise Wiki for coming years.


wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,60 @@
...
+
+## How?
+Keep in mind that documentation is different than comments. Comments are placed between code lines in the source file, whereas documentation is placed in the header file before function and class definitions. Good code should be self-explanatory which means that it often does not need any comments, when properly split into smaller functions that have been well-documented. And when writing comments, it should mostly be used to answer the why questions rather than the what questions, i.e. comments should focus on why does this part of code works, why don't I have to check certain situations, etc. rather than what the code actually does. Documentation on the other hand can be used to explain what the code does, but should also be used to explain why this code works. Documentation should give an answer to the most important questions regarding the code which can include answers on questions like:
+* What does the code compute? What is the meaning of the output? What are the postconditions on the output? What is the meaning of the input arguments? Which preconditions hold on the input arguments? Which requirements should be satisfied before using the code? Etc.
```
Yuhanun:
```suggestion
* What does the code compute? What is the meaning of the output? What are the postconditions on the output? What is the meaning of the input arguments? Which preconditions should hold on the input arguments? Which requirements should be satisfied before using the code? Etc.
```

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,60 @@
...
+
+## How?
+Keep in mind that documentation is different than comments. Comments are placed between code lines in the source file, whereas documentation is placed in the header file before function and class definitions. Good code should be self-explanatory which means that it often does not need any comments, when properly split into smaller functions that have been well-documented. And when writing comments, it should mostly be used to answer the why questions rather than the what questions, i.e. comments should focus on why does this part of code works, why don't I have to check certain situations, etc. rather than what the code actually does. Documentation on the other hand can be used to explain what the code does, but should also be used to explain why this code works. Documentation should give an answer to the most important questions regarding the code which can include answers on questions like:
```
Yuhanun:
```suggestion
Keep in mind that documentation is different than comments. Comments are placed between code lines in the source file, whereas documentation is placed in the header file before function and class definitions. Good code should be self-explanatory which means that it often does not need any comments, when properly split into smaller functions that have been well-documented. And when writing comments, it should mostly be used to answer the why questions rather than the what questions, i.e. comments should focus on why does this part of code work, why don't I have to check certain situations, etc. rather than what the code actually does. Documentation on the other hand can be used to explain what the code does, but should also be used to explain why this code works. Documentation should give an answer to the most important questions regarding the code which can include answers on questions like:
```

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,60 @@
...
+* In case of class documentation: Who has written/maintained this code? At which date has this code been written?
+* Less important, but could be included as well: Why does this code work? Why is this code efficient? Why did you choose this type of design? What other solutions have you tried out for this code? What major decisions have been made regarding this code? 
+
+## Good Example
+Below is an example how you should document your code:
+```
+/**
+ * Get the intersection between two (finite) LineSegment. Note that this function knows how to deal with LineSegments that are points and how to deal with
+ * parallel LineSegments. Moreover this function guarantees to return an intersection point if the LineSegments intersects. In case:
+ * - There is no intersection then std:nullopt is returned.
```
Yuhanun: This line is double, you mention it in @return.

```suggestion
```

Yuhanun (dismissed):

wiki/PreventiveMaintenance.md
HaicoDorenbos: I don't see the change...

wiki/PreventiveMaintenance.md
HaicoDorenbos: Well in the code base it is written this way. So if I would change this then I need to change it as well in the code base.

RobotJente: But do you agree with the suggested change?

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,64 @@
...
+* What is the meaning of abbreviations used inside the code? Is there a website which explains (partially) the meaning of the code? From which source has this code been copied? Etc.
+* In case of class documentation: Who has written/maintained this code? At which date has this code been written?
+* Less important, but could be included as well: Why does this code work? Why is this code efficient? Why did you choose this type of design? What other solutions have you tried out for this code? What major decisions have been made regarding this code? 
+
+## Good Example
+Below is an example how you should document your code:
+```
+/**
+ * Get the intersection between two (finite) LineSegment. Note that this function knows how to deal with LineSegments that are points and how to deal with
+ * parallel LineSegments. Moreover this function guarantees to return an intersection point if the LineSegments intersects. In case:
```
RobotJente, line 25: I think this line doesnt add any information. What does knowing how to deal with a linesegment that is a point mean? 
```suggestion
 * Get the intersection between two (finite) LineSegment. This function guarantees to return an intersection point if the LineSegments intersects. In case:
```

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,64 @@
...
+ * - There is no intersection then std:nullopt is returned.
+ * - There is a single intersection point then that intersection point is returned.
+ * - There are multiple intersection points (infinitely many) which happens when the LineSegments overlap then it prefers to returns 1: start, 2: end, 3: 
+ *  line.start of which the first most option is selected that lies on both LineSegments. Note that if none of these points lie on both LineSegments and if both 
+ *  the LineSegments are parallel then the LineSegments cannot intersect.
```
RobotJente: I think this line also doensn't add information, you don't need to explain here what happens if there is no intersection. This information seems out of place here in my view
```suggestion
 * - There are multiple intersection points (infinitely many) which happens when the LineSegments overlap then it prefers to returns 1: start, 2: end, 3: 
 *  line.start of which the first most option is selected that lies on both LineSegments
```

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,64 @@
...
+```
+The documentation basically tells the same as the name of the function. Hence the goal of documentation, to inform others about usage, is missing here. What kind of info is computed is totally unclear? What are requirements for calling this function is also missing? When should this function be called?
+
+# Design
+
+# Unit Testing
+## Why?
+Writing unit test cases for your code is very important, because
+- Even the simplest code can contain bugs and even when fully reasoned that the code works, it can still contain some bugs.
+- It encourage you to inspect your code better for bugs, because during the process of writing test you might already find bugs (without executing the tests).
```
RobotJente:
```suggestion
- It encourages you to inspect your code better for bugs, because during the process of writing test you might already find bugs (without executing the tests).
```

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,64 @@
...
+- It improves the trust in the code, since it is properly tested and therefore it is less likely that your code suddenly fails during the RoboCup.
+- It will help you find the location of bugs.
+
+## How?
+When writing unit test cases, your goal is to create unit test cases that might/will fail. So for example you should test the path planning algorithm for the case that the ball cannot be reached (because it is surrounded by enemy robots) and for LineSegments you should test LineSegments that are actually points (have an equal start position as end position). Using random unit test cases or just writing a few test cases will in most cases not found the most common type of bugs. Hence you should be more systematic in writing test cases. For this there are 2 types of approaches: black box testing and white box testing.
```
RobotJente:
```suggestion
When writing unit test cases, your goal is to create unit test cases that might/will fail. So for example you should test the path planning algorithm for the case that the ball cannot be reached (because it is surrounded by enemy robots) and for LineSegments you should test LineSegments that are actually points (have an equal start position as end position). Using random unit test cases or just writing a few test cases will in most cases not find the most common type of bugs. Hence you should be more systematic in writing test cases. For this there are 2 types of approaches: black box testing and white box testing.
```

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,64 @@
...
+ * - There are multiple intersection points (infinitely many) which happens when the LineSegments overlap then it prefers to returns 1: start, 2: end, 3: 
+ *  line.start of which the first most option is selected that lies on both LineSegments. Note that if none of these points lie on both LineSegments and if both 
+ *  the LineSegments are parallel then the LineSegments cannot intersect.
+ *
+ * @param line The other (finite) LineSegment.
+ * @return std::nullopt if the LineSegments do not intersect. Otherwise return an intersection point.
+ */
+[[nodiscard]] std::optional<Vector2> intersects(const LineSegment &line) const;
+```
+Note that this documentation gives a clear description of what the code does, by explaining what happens for each situation. But beside that it also explain why this code works, by telling that the function knows how to deal with parallel & point cases and that you don't have to check if line.end is on the line if start, end and line.start are not on both lines.
```
RobotJente: As i explained above I don't think that is necessarily good documentation, you don't need to explicitly describe every behaviour of a function, just give an overview. I'm really in favour of short, concise documentation. 
```suggestion
Note that this documentation gives a clear description of what the code does, by explaining what happens for each situation. But besides that it also explains why this code works, by telling that the function knows how to deal with parallel & point cases and that you don't have to check if line.end is on the line if start, end and line.start are not on both lines.
```

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,64 @@
...
+* What does the code compute? What is the meaning of the output? What are the postconditions on the output? What is the meaning of the input arguments? Which preconditions should hold on the input arguments? Which requirements should be satisfied before using the code? Etc.
+* Is there some known bug/strange behavior in the code which hasn't been solved yet? How does the code respond in case of an exception? What happens when preconditions/requirements have been violated? Etc.
```
RobotJente: We don't really use exceptions
```suggestion
* Is there some known bug/strange behavior in the code which hasn't been solved yet? What happens when preconditions/requirements have been violated? Etc.
```

RobotJente (dismissed):

wiki/PreventiveMaintenance.md
HaicoDorenbos: We should use exceptions, however adjusting the system such that it does use exceptions is too difficult

wiki/PreventiveMaintenance.md
HaicoDorenbos: Well I don't think it is a major problem and I guess it takes more effort to update it at both places then it would give as benefit.

wiki/PreventiveMaintenance.md
HaicoDorenbos: It gives partially answer to the important question: Why does this code work? Because it knows how to deal with points... Therefore it convinces the software developers that I took edge/odd cases into account

wiki/PreventiveMaintenance.md
HaicoDorenbos: I guess you misread this. It explains what happens when there are multiple intersections which is quite important.

wiki/PreventiveMaintenance.md
Yuhanun: Iirc it's a grammatical fix

wiki/PreventiveMaintenance.md
RobotJente: I think it's worth changing since this is an example of what we consider good documentation, so I want it to be as good as it can be

wiki/PreventiveMaintenance.md
RobotJente: I was talking about the part where it explains that if the lines are parallel then they don't intersect. This could be seen as relevant information, but I feel like if a paragraph gets too long people will start understanding it less easily. That's just my personal opinion, though. So I prefer to keep docs to say only what is absolutely necessary in order to understand the function. Of course, that's also a subjective measure. So there isn't really a right answer, I guess.

RobotJente (commented):
I like where this is going :)

wiki/PreventiveMaintenance.md
HaicoDorenbos: But you are not changing anything related to that comment now...

HaicoDorenbos:
@RobotJesse Parallel lines can intersect. Two lines that are equal are parallel and they do intersect...

Yuhanun:
We should either merge this or close this PR so we start clean this year.

HaicoDorenbos:
Shall we then just merge it?

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,213 @@
...
+* It makes it easier for other software developers to review your pull requests, since they only have to check if the code corresponds with the specification (documentation) and check if the specification is suitable. Without documentation reviewers have to figure out the specification themselves which takes more time to review the pull request.
+* It makes it easier to find bugs in the code, since you just have to compare the documentation with the implementation. If the documentation does not match the implementation then the implementation is supposed to be wrong. Without documentation you have no clue whether something in the code has been implemented intentionally or unintentionally.
+* It stimulates re-usage, without documentation other software developers have problems figuring out which functionality is provided by which methods, what preconditions are required and which post-conditions have been satisfied.
+* It makes it easier to adhere to the other points: Design, Testing, Consistency, Code cloning/Dead code. For example without documentation it is not even possible to test your code, because you have no specification which you compare the implementation against.
```
Yuhanun, line 9:
```suggestion
* It makes it easier to adhere to the other points: Design, Testing, Consistency, Code cloning/Dead code. For example without documentation it is not even possible to test your code, because it wouldn't make sense if someone else inherits your code and the tests don't seem to match the implementation.
```

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,213 @@
...
+* What does the code compute? What is the meaning of the output? What are the postconditions on the output? What is the meaning of the input arguments? Which preconditions should hold on the input arguments? Which requirements should be satisfied before using the code? Etc.
+* Is there some known bug/strange behavior in the code which hasn't been solved yet? How does the code respond in case of an exception? What happens when preconditions/requirements have been violated? Etc.
+* What is the meaning of abbreviations used inside the code? Is there a website which explains (partially) the meaning of the code? From which source has this code been copied? Etc.
+* In case of class documentation: Who has written/maintained this code? At which date has this code been written?
```
Yuhanun:
```suggestion
* In case of class documentation: Who has written/maintained this code? At which date has this code been written? It's important to know who your point of contact should be in case something isn't sure.
```

wiki/PreventiveMaintenance.md
HaicoDorenbos: I don't think you get the point. You cannot test something of which you haven't specified the behavior.

JaroKuiken:
I committed the requested changes that made sense imo and resolved the rest. Can someone from the old team check so we can merge?

rolfvdhulst (approved):
Seems fine to me.

wiki/PreventiveMaintenance.md
```diff
@@ -0,0 +1,211 @@
...
+- Output: Make sure that the output of similar methods have the same meaning. So when returning a location on the field by different methods, you should make sure that these locations are comparable.
+- Exceptions: Exceptions should be dealt with in the same way by similar methods. So either throw an exception, send an error message, ignore it, etc. but never use a combination of these options. Also odd cases should be dealt with in the same way by similar methods, so for example if you compute intersections between objects and there a multiple of them then you should be consistent in what you return for all these methods in different classes.
+- Metrics: When using a distance metric in similar methods make sure that you use the same distance metric for all of them. For example in case you check whether two locations (Vector2) are the same then either use Manhattan distance or Euclidean distance everywhere (checking if float values are equal can cause issues due to rounding of float values).
+
+Important regarding consistency is that you should discuss a lot with your fellow software developers on which guidelines/definitions you will agree on. 
+
+# Code Clones
+
+# Dead Code
```
Yuhanun, line 211: these two aren't done yet?

JaroKuiken: Oh didn't see these. I guess not, but honestly, do you feel like finishing it? I'm more interested in not having it as a PR anymore tbh

Yuhanun: Agreed, @HaicoDorenbos any chance you could do these two today?

PR #1158, opened 2020-07-07
Removed weird map from worldData

timovdk:
There was a weird, unused map at the top of worldData


HaicoDorenbos (approved):
If everything still compiles then you can merge it...

RobotJente (approved):
We used this during the world refactor because we needed the previous velocity for the PID controllers, but we solved it another way it seems. As far as I can tell this map is indeed unused

Yuhanun:
thank you.

PR #1163, opened 2020-07-16
Path Planning wiki page

Ratoone:
Added a path planning wiki page that contains information about the current implementation + possible future improvements


Ratoone:
Ignore the 100,000 commits, they are the difference between dev and demo/dev

Yuhanun:
what's all the additional commits?

wiki/PathPlanning.md
```diff
@@ -0,0 +1,67 @@
...
+* there is no path
+* the target changed
+* a collision is detected between the moving robot and its next path way point
+
+## Improvements
+Consider experimenting with / researching different algorithms for different tasks. Consider also using different algorithms for each robot (and think before whether it would be better to have separate algorithms for each robot or not). Ideas:
+1. Path Planning
+- Voronoi (slow for many robots, also large avoidance - an implementation already exists in the code base)
+- Split the search area into grids and check grid occupancy
+- RRT* augmented with Particle Swarm (or other RTT implementations - many teams use RTT for their path planning)
```
timovdk: RRT or RTT?

wiki/PathPlanning.md
```diff
@@ -0,0 +1,67 @@
...
+- Magnetic field lines (basically smart picking for RTT - draw magnetic field lines from the current position to the target and sample points from these lines to try to build a path)
```
timovdk: Same ^

wiki/PathPlanning.md
```diff
@@ -0,0 +1,67 @@
...
+2. Path Tracking
+- Pure Pursuit (use a look ahead distance to try to track a path)
+- Fast Marching (personally I didn't understand this, but there are papers on it)
+- Bezier (interpolate the path points and take the path derivative(?) to track it)
+- Potential Field (each obstacle gives a repelling force proportional to distance, while the goal has an attracting force - maybe combine with planning)
+- Couple the PID controllers together - also taking into account the robots maximum velocity
+- Bang bang control (accelerate to maximum velocity and slow down when needed - Rolf may want to write some more details here)
```
timovdk: @rolfvdhulst

wiki/PathPlanning.md
rolfvdhulst: RRT is correct.

wiki/PathPlanning.md
rolfvdhulst: This is actually under the wrong header kind of since it's more a path generation technique than a path tracking.

wiki/PathPlanning.md
timovdk:
```suggestion
- RRT* augmented with Particle Swarm (or other RRT implementations - many teams use RRT for their path planning)
- Magnetic field lines (basically smart picking for RRT - draw magnetic field lines from the current position to the target and sample points from these lines to try to build a path)
```

timovdk (approved):

wiki/PathPlanning.md
Ratoone: Don't you already need a path to do this? Doing collision checking and everything else

rolfvdhulst: No, if you are under that impression you're completely misunderstanding the premise of bangbang trajectories. Multiple other teams are using them as a concept to generate smooth trajectories quickly, rather than performing control.

rolfvdhulst: Actual bang-bang control of our motors is a very bad idea, and I can probably write you an essay on why that is the case.

HaicoDorenbos:
Why does Travis fail, since you only added documentation? You can create a new branch from development and check if Travis also fails for that branch. If it does so then the problem is not related to your changes. If it does not then you can always create a separate PR with the same changes. I encountered this situation once and this solved the problem for me.

timovdk:
> Why does Travis fail, since you only added documentation? You can create a new branch from development and check if Travis also fails for that branch. If it does so then the problem is not related to your changes. If it does not then you can always create a separate PR with the same changes. I encountered this situation once and this solved the problem for me.

Every open PR has this problem, the build fails when compiling sdl2. It happens only on Travis; locally everyone can still compile. @Yuhanun will look into this he said :)

RobotJente (commented):
Looks good in general. Do you have an opinion on which improvement you think will be the best to start with @Ratoone?

Ratoone:
Probably the collision part. But I'll let the new team decide priorities

PR #1166, opened 2020-07-17
STP Documentation and Cleanup - Headers

### LINKED ISSUE #1164: Document STP base and derived classes

timovdk:
Considering the size of this PR already, I will refactor the implementations in a separate PR

- Reformatting of code using clang-format (according to @HaicoDorenbos his standard that can be found in roboteam_suite/formatting) ((basically clang-format google style with extended column size))
- Revision of documentation
- General cleanup things
- Renaming (Open for suggestions)

closes #1164

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -85,7 +85,7 @@ IncludeIsMainSourceRegex: ''
 IndentCaseLabels: true
 IndentGotoLabels: true
 IndentPPDirectives: None
-IndentWidth:     2
+IndentWidth:     4
 IndentWrappedFunctionNames: false
 JavaScriptQuotes: Leave
 JavaScriptWrapImports: true
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,168 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
+AccessModifierOffset: -1
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: Never
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortLambdasOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: WithoutElse
+AllowShortLoopsOnASingleLine: true
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterCaseLabel:  false
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Attach
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     180
+CommentPragmas:  '^ IWYU pragma:'
+CompactNamespaces: false
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DeriveLineEnding: true
+DerivePointerAlignment: true
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: true
+ForEachMacros:
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
+IncludeBlocks:   Regroup
+IncludeCategories:
+  - Regex:           '^<ext/.*\.h>'
+    Priority:        2
+    SortPriority:    0
+  - Regex:           '^<.*\.h>'
+    Priority:        1
+    SortPriority:    0
+  - Regex:           '^<.*'
+    Priority:        2
+    SortPriority:    0
+  - Regex:           '.*'
+    Priority:        3
+    SortPriority:    0
+IncludeIsMainRegex: '([-_](test|unittest))?$'
+IncludeIsMainSourceRegex: ''
+IndentCaseLabels: true
+IndentGotoLabels: true
+IndentPPDirectives: None
+IndentWidth:     2
+IndentWrappedFunctionNames: false
+JavaScriptQuotes: Leave
+JavaScriptWrapImports: true
+KeepEmptyLinesAtTheStartOfBlocks: false
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: None
+ObjCBinPackProtocolList: Never
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakAssignment: 2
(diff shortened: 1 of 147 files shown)


HaicoDorenbos:
Wow, 147 files changed is quite a lot and when looking at the PR title I guess it could have been split up in multiple PR's. So next time could you split up your PR in smaller ones? :blush:

Yuhanun:
Set indentation to 4.

.clang-format
```diff
@@ -0,0 +1,168 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
```
Yuhanun, line 3: Why?

timovdk: Haico picked this in his formatting folder in roboteam_suite so I thought let's stick to it. If you prefer a different style, let me know 😇

Yuhanun: It's commented out?

timovdk: Yes, since the clang-format command generates this file based on the Google style, so you can still change/see the different settings, instead of looking online. I just did `clang-format style=google`

timovdk:
Changes are a lot more obvious now that indentation width is back to 4 (thanks @Yuhanun). Don't mind the formatting of the .cpp files, this will be done in follow up PRs

include/roboteam_ai/stp/Skill.h
```diff
@@ -6,28 +6,29 @@
...
    protected:
     /**
-     * Current status of the last time update() was called
+     * Status of the last time update() was called
      */
     Status currentStatus;
 
     /**
-     * Robot command that will be used for operations, such as publishing
+     * Robot command that will eventually be send to the robot
```
RobotJente:
```suggestion
     * Robot command that will eventually be sent to the robot
```

RobotJente (commented):
I started reviewing this and then saw you were going to split it into multiple PR's, so I'll wait for that. But I did find this typo! :yum:

Yuhanun:
Trying to fix, not sure why it fails, though.

timovdk:
> I started reviewing this and then saw you were going to split it into multiple PR's, so I'll wait for that. But I did find this typo! :yum: 

I'm not splitting this PR though, only future ones. It's basically 1 change in every file.

include/roboteam_ai/stp/Play.hpp
```diff
@@ -23,56 +21,56 @@ namespace rtt::ai::stp {
...
-     * Updated the stored world
-     * @param world
+     * Updates the stored world pointer and field using this world pointer
+     * @param pointer to World
      */
     void updateWorld(world_new::World* world) noexcept;
 
     /**
-     * Updates all the roles
+     * Updates all the roles that have robots assigned to them
```
timovdk:
```suggestion
     * Updates (or ticks) all the roles that have robots assigned to them
```

include/roboteam_ai/stp/Play.hpp
```diff
@@ -23,56 +21,56 @@ namespace rtt::ai::stp {
...
+     * Virtual function that is called in initialize().
+     * This function should contain all play-specific init code
+     */
     virtual void onInitialize() noexcept {};
 
     /**
-     * Updated the stored world
-     * @param world
+     * Updates the stored world pointer and field using this world pointer
+     * @param pointer to World
```
timovdk:
```suggestion
     * Updates the stored world pointer and after that, updates the field instance using the updated world pointer
     * @param pointer to World
```

include/roboteam_ai/stp/Play.hpp
```diff
@@ -23,56 +21,56 @@ namespace rtt::ai::stp {
...
 
     /**
-     * Updates all the roles
+     * Updates all the roles that have robots assigned to them
      */
     virtual void update() noexcept;
 
     /**
-     * Calculates all the info (mostly positions) the roles in this play need to execute
+     * Calculates all the info the roles need in order to execute correctly
```
timovdk:
```suggestion
     * Calculates all the info the roles need in order to execute correctly.
     * This is a purely virtual function, so it is implemented in every play.
```

include/roboteam_ai/stp/Tactic.h
```diff
@@ -30,24 +28,11 @@ class Tactic {
...
-    /**
-     * called on update of this tactic
-     */
-    virtual void onUpdate(Status const &status) noexcept = 0;
-
-    /**
-     * called on terminate of this tactic
-     */
-    virtual void onTerminate() noexcept = 0;
-
```
RobotJente, line 50: I assume you removed this because the onupdates do not have a purpose, right?

include/roboteam_ai/stp/Tactic.h
```diff
@@ -82,19 +67,19 @@ class Tactic {
...
+    Status update(StpInfo const &info) noexcept;
 
     /**
      * Calls onTerminate
      */
-    virtual void terminate() noexcept;
+    void terminate() noexcept;
 
     /**
      * Ensure proper destruction of Tactic classes
```
RobotJente, line 85: I'm not sure, but doesn't C++ need the virtual keyword to access the right method in the derived class, and will go to the base class method without it?

src/stp/Tactic.cpp
```diff
@@ -4,58 +4,67 @@
...
+  }
+
+  // Update skill info
+  auto skill_info = calculateInfoForSkill(info);
+
+  if (skill_info) {
+    // Update the current skill with the new SkillInfo
+    auto status = skills.update(skill_info.value());
+
+    (void)status;
```
RobotJente: I don't understand how this change works, does it preserve the functionality? And why did you change it?

src/stp/new_tactics/DriveWithBall.cpp
```diff
@@ -10,58 +10,44 @@
...
+  // Fail if we don't have the ball or there is no movement position
+  return !info.getRobot()->hasBall() || !info.getPositionToMoveTo();
 }
 
 bool DriveWithBall::shouldTacticReset(const StpInfo& info) noexcept {
-    // Should reset if the angle the robot is at is no longer correct
-    auto robotAngle = info.getRobot()->get()->getAngle();
-    auto ballToRobotAngle = (info.getBall()->get()->getPos() - info.getRobot()->get()->getPos()).angle();
-
-    return fabs(robotAngle + Angle(ballToRobotAngle)) <= stp::control_constants::GO_TO_POS_ANGLE_ERROR_MARGIN;
```
RobotJente: I don't think this preserves the meaning

RobotJente (changes requested):
In general I really like the changes you've made in this PR. Clear docs, and I like how you fixed the ShootOrChip enum naming. Nice work! I have some points that I think need to be checked but otherwise it's great :)

include/roboteam_ai/stp/Tactic.h
timovdk: Yes, to make the structure more clear. And besides, STP should not be stateful and these functions kind of lead the developer to believe they can be.

include/roboteam_ai/stp/Tactic.h
timovdk: `initialize(), terminate() and update()` are only called through the respective functions from Role, which are called from Play. These do not have overrides (and should not  have them either), so no need for `virtual`

src/stp/Tactic.cpp
timovdk: Well, the return of update is never used, but it is marked [[no-discard]] in the state machine. This cast to void basically uses up the return to suppress the compile warning. I did not want to touch the state machine so I opted for this solution.

src/stp/new_tactics/DriveWithBall.cpp
timovdk: I will revert this then!

include/roboteam_ai/stp/Tactic.h
RobotJente: You're right! Makes sense :)

src/stp/Tactic.cpp
RobotJente: To make it clearer to the person reading the code that this is not something to worry about:
```suggestion
    (void)status; // return of update is never used, but it is marked [[no-discard]] in the state machine. This cast to void uses up the return to suppress the compile warning.
```

RobotJente (approved):
Conditionally approved :) I think the branch should be tested one last time as it introduces quite a lot of changes

timovdk:
> Conditionally approved :) I think the branch should be tested one last time as it introduces quite a lot of changes

Do you also want to test @RobotJesse ? I tested just now and found 2 small issues and fixed them.

HaicoDorenbos:
Regarding the code style format: I did not agree on this code style format (for example the line widths are quite long). However I agree that using a code style format (even if it is a bad one) is better than not using a code style format at all, similar to this meme: https://res.cloudinary.com/practicaldev/image/fetch/s--H8LtA_7W--/c_limit%2Cf_auto%2Cfl_progressive%2Cq_auto%2Cw_880/http://i.imgur.com/vOWAAUK.png

timovdk:
> Regarding the code style format: I did not agree on this code style format (for example the line widths are quite long). However I agree that using a code style format (even if it is a bad one) is better than not using a code style format at all, similar to this meme: https://res.cloudinary.com/practicaldev/image/fetch/s--H8LtA_7W--/c_limit%2Cf_auto%2Cfl_progressive%2Cq_auto%2Cw_880/http://i.imgur.com/vOWAAUK.png

Well, you made the formatting tool in roboteam_suite, from which I copied the settings. The only thing I changed (by request from @Yuhanun) is the indent width (which I completely agree with btw). So basically you did not agree to it, but invented it haha

RobotJente (approved):

PR #1171, opened 2020-07-24
Refactored skills

### LINKED ISSUE #1172: Refactor skills

timovdk:
I refactored the skills.
closes #1172


src/stp/skills/GoToPos.cpp
```diff
@@ -31,21 +25,19 @@ Status GoToPos::onUpdate(const StpInfo &info) noexcept {
     Vector2 targetVelocity = robotCommand.vel.stretchToLength(targetVelocityLength);
 
     // Set velocity and angle commands
-    command.mutable_vel()->set_x(targetVelocity.x);
-    command.mutable_vel()->set_y(targetVelocity.y);
+    command.mutable_vel()->set_x(static_cast<float>(targetVelocity.x));
+    command.mutable_vel()->set_y(static_cast<float>(targetVelocity.y));
```
RobotJente, line 29: We cast here because targetVelocity.x is a double and we need a float, right? Otherwise we get nice yellow warnings and this avoids those?

RobotJente (approved):
I'm not sure if you changed anything in the shoot skill since git seems to think it is a new file, but other than that it looks good. Nice work!

src/stp/skills/GoToPos.cpp
timovdk: Yep, exactly

PR #1173, opened 2020-07-24
Refactored tactics

timovdk:
Branched from refactor/stp/skills


RobotJente (commented):
Also looks fine to me, again two files seem to be completely new but they look like they did before... I guess its a git thing?

timovdk:
> Also looks fine to me, again two files seem to be completely new but they look like they did before... I guess its a git thing?

Yes, you are right

RobotJente (approved):

PR #1175, opened 2020-07-24
Refactored plays

timovdk:


RobotJente (approved):
Looks good, I like that you split the massive calculateInfos in the pass plays. Good job! :)

PR #1191, opened 2020-09-25
Slightly hacky fix for play selecting through interface

JaroKuiken:
### Summary
Whenever a play is selected in the interface, a boolean is switched to true which then forces the ApplicationManager to decide on a new play.

We might need to **make sure this is really only used during testing, not in an actual game**.

### Related issues
Somewhere in the STP board there is a checkmark that can be checked now, hooray


include/roboteam_ai/stp/PlayDecider.hpp
```diff
@@ -25,6 +25,7 @@ class PlayDecider {
      * @param play Play to lock to
      */
     static void lockPlay(Play *play);
+    static bool change;
```
Yuhanun: What I would suggest is having a Play* here, which if it's nullptr is not selected from the interface. That way in decidePlay you can just

```cpp
if (interfacePlay) {
    currentPlay = interfacePlay;
   // update n stuff...
}
```

Not exactly sure what the `change` stands for either. Does it mean the internal AI can change the play? does it mean the interface has selected a play that should be changed to?... if so, where is this play stored? 

Where does the playdecider check whether a play is available from the interface?

src/ApplicationManager.cpp
```diff
@@ -150,6 +150,15 @@ void ApplicationManager::runOneLoopCycle() {
 void ApplicationManager::decidePlay(world::World *_world) {
     playChecker.update(_world);
 
+    // Here for manual change with the interface
+    if(rtt::ai::stp::PlayDecider::change) {
+        auto validPlays = playChecker.getValidPlays();
+        currentPlay = playDecider.decideBestPlay(_world, validPlays);
+        currentPlay->updateWorld(_world);
+        currentPlay->initialize();
+        rtt::ai::stp::PlayDecider::change = false;
```
Yuhanun: Wouldn't this make it so that the net iteration it'll switch back to whatever play it wants to?

src/stp/PlayDecider.cpp
```diff
@@ -5,12 +5,19 @@
...
-    if (lockedPlay) {
-        return lockedPlay;
+
+    bool PlayDecider::change = false;
+
+    Play *PlayDecider::decideBestPlay(world::World *pWorld, std::vector<Play *> plays) noexcept {
+        if (lockedPlay) {
+            return lockedPlay;
+        }
+        return *std::max_element(plays.begin(), plays.end(), [&](auto &largest, auto &play) { return largest->score(pWorld) < play->score(pWorld); });
```
Yuhanun, line 13: nice usage of C++ functions <3

src/stp/PlayDecider.cpp
```diff
@@ -5,12 +5,19 @@
...
     }
-    return *std::max_element(plays.begin(), plays.end(), [&](auto &largest, auto &play) { return largest->score(pWorld) < play->score(pWorld); });
-}
 
-void PlayDecider::lockPlay(Play *play) { lockedPlay = play; }
+    // This is only used by the interface to force new plays
+    void PlayDecider::lockPlay(Play *play) {
+        PlayDecider::change = true;
+        lockedPlay = play;
```
Yuhanun: Ah this is where it changes play, consider removing the `bool change`, instead just use the play*.

src/ApplicationManager.cpp
```diff
@@ -150,12 +150,20 @@ void ApplicationManager::runOneLoopCycle() {
...
+        currentPlay->initialize();
+    }
+
     // A new play will be chosen if the current play is not valid to keep
     if (!currentPlay || !currentPlay->isValidPlayToKeep(_world)) {
         auto validPlays = playChecker.getValidPlays();
         if (validPlays.empty()) {
             RTT_ERROR("No valid plays")
-            currentPlay = playChecker.getPlayForName("Defend Shot");
+            currentPlay = playChecker.getPlayForName("Defend Shot"); //TODO Try out different default plays so playing against yourself doesnt deadlock
```
rolfvdhulst: Please don't use 'deadlock' unless you want people to later throw this code out again.

src/stp/PlayDecider.cpp
```diff
@@ -5,12 +5,13 @@
 #include "stp/PlayDecider.hpp"
 
 namespace rtt::ai::stp {
-Play *PlayDecider::decideBestPlay(world::World *pWorld, std::vector<Play *> plays) noexcept {
-    if (lockedPlay) {
-        return lockedPlay;
+
+    Play *PlayDecider::decideBestPlay(world::World *pWorld, std::vector<Play *> plays) noexcept {
+        return interfacePlay ? interfacePlay : *std::max_element(plays.begin(), plays.end(), [&](auto &largest, auto &play) { return largest->score(pWorld) < play->score(pWorld); });
```
rolfvdhulst: This line is way too long/hard to read. Split it up in multiple lines, especially if you define a lambda function.

rolfvdhulst: also, I would suggest const auto& inside of the lambda,

rolfvdhulst: Also, you should probably check if the vector of plays is non-empty as otherwise this will be UB and will probably segfault.

Yuhanun: score() is non-const iirc.

The vector of `plays` is guaranteed to not be empty as decideBestPlay() is ran after initialization and in initialization we add plays to this vector.


Aggreed with the too hard to read, not sure why this is like this im fairly sure i didn't write it this way? (maybe tho idk)

rolfvdhulst: score() not being const seems like a problem? That smells of side-effects...

rolfvdhulst: But anyways, fine by me

Yuhanun: yeah idk man. ask jesse i guess?

JaroKuiken:
So the rest of this PR can be approved i think. Unless you guys don't agree with the newer version of the fix?

Yuhanun (approved):
LGTM

PR #1194, opened 2020-10-07
fix the bloody tests

rolfvdhulst:
This gives you a good idea of why we should always double check or TRIPLE check PR's...
Glorious typos

It compiles again, atleast locally. Test server is another thing...


JaroKuiken (approved):
Pff, can barely see the difference in upper or lowercase even when its highlighted. Nice one Rolf

Yuhanun:
Fixing docker right now.

Yuhanun:
Okay so "Pull Request" builds are trying to build _current_ development, kind of useless. I'm going to disable that one.

Yuhanun:
LGTM merging.

PR #1195, opened 2020-10-08
Reinstated old play selecting fix, interface runs better with it

JaroKuiken:
### Summary
Same fix as commit: 51ddb66fbc862b2c05e709153b71610d85f63beb that was made in development before. The 'slightly less hacky' fix, made the interface very slow, so we will use this hackier fix until we rework the interface itself with the system architecture redesign.


rolfvdhulst (approved):
Seems fine to me.

PR #1197, opened 2020-10-22
Removed all occurences of geneva

emielsteerneman:
Removed any and all occurences of the geneva drive. All targets are still compiling succesfully.


test/ControlTests/ShotControllerTest.cpp
```diff
@@ -180,82 +149,13 @@ TEST(ShotControllerTest, getshotdata_test) {
...
-        rtt::world::world->updateWorld(world);
-        auto robot = world::world->getUs().at(0);
-        robot->setGenevaState(3);
-        robot->setWorkingGeneva(true);
-        robot->setTimeToChangeOneGenevaState(-1);
-
-        shotdata = shotController.getRobotCommand(*robot, testhelpers::WorldHelper::getRandomFieldPosition(field), true, BallSpeed::MAX_SPEED, true, ShotPrecision::HIGH);
-        EXPECT_FALSE(shotdata.kicker);
-        EXPECT_FLOAT_EQ(shotdata.geneva, 3);
-    }
```
rolfvdhulst, line 209: This tests more than just the geneva. Have you tested if ShotController still works without the geneva? I don't like these tests being removed.

rolfvdhulst: Same goes for all other tests in this file.

rolfvdhulst (changes requested):
Are there really no changes in ShotController at all? Seems weird to me, since you edited the test so much, and since ShotController computes geneva state. I heavily suspect you introduced some bugs.

test/ControlTests/ShotControllerTest.cpp
RobotJente: I'm not 100% sure but I don't think the shotcontroller is used anymore. @Ratoone can you confirm this?

emielsteerneman: The only occurrences of "ShotController" anywhere is in the file ShotControllerTest.cpp, and that file does not belong to any target.

rolfvdhulst: Ah ok. Then IMO we remove the entire class aswell, as it's functionality relied on geneva a lot on the past.

emielsteerneman: Can we merge this first?

rolfvdhulst: Why not remove the class? I'll even do it for you if you don't want to. Keeping code that is unused, has no use and which tests have been removed makes no sense.

emielsteerneman: If you could do it, that'd be great! I have no idea about the structure of the code anymore

rolfvdhulst (approved):
LGTM

JaroKuiken:
Never mind, in a different PR. LGTM

JaroKuiken (approved):

PR #1198, opened 2020-10-30
Improvement for stop GameState Plays

JaroKuiken:
Added BallOnOurSide and (the newly created) BallOnTheirSideInvariant To Defensive and AggressiveFormation respectively and gave them the same score. This, so we have a defensive position at a stop GameState when the ball is on our side and vice versa when its on their side


rolfvdhulst (commented):
Code looks fine. I'm slightly concerned about the behavior though; because you added the invariant and changed the scores to the general plays AggressiveFormation and DefensiveFormation, does this not change their behavior in other cases too? That's probably not desirable. I'm not sure *if* or when these other tactics are used, but since they already exist, I assume they are.

JaroKuiken:
Both Plays are only used in the stop gamestate, nowhere else, and they are also the only ones that should get chosen when the game is in the stop gamestate. They will only be chosen when in the stop gamestate, so increasing their score will only make sure that other Plays won't get chosen when in this state. I'm also quite sure that the added invariants will only influence which of these two Plays get chosen while in the stop gamestate, they are only used by these two Plays and the edge case where the ball is exactly in the middle is also covered in line 9 of the TheirSideInvariant.cpp.

rolfvdhulst:
If that's the functionality, then we should rename the Formations 'AggressiveStopFormation' and 'DefensiveStopFormation' respectively. That way it's clear that they belong to this game state. People using tactics for multiple purposes really cost us in Robocup 2018.

JaroKuiken:
True, good idea

rolfvdhulst (approved):
You forgot to rename the formations in CMakeLists, so now they don't compile ;). 

LGTM for the rest though,.

JaroKuiken:
Ah yeah, I did fix the CMake when I accidentally did the renaming in the wrong branch. But when I reverted that commit and did the renaming in this branch I forgot about it. Anyway, should be fixed now

PR #1199, opened 2020-11-03
Changed robot formation for 8m penalties

Frankavj:
The penalties are now taken at 8m from the goal, instead of closer. This means that the old formation of the non-attacking / non-goalkeeping robots needed to be changed as they would be in the way. To ensure that the penalty plays would also work for penalty shoot-outs, the robots have been moved all the way to the side.


src/stp/plays/PenaltyThemPrepare.cpp
```diff
@@ -28,23 +28,23 @@ PenaltyThemPrepare::PenaltyThemPrepare() : Play() {
 uint8_t PenaltyThemPrepare::score(world::World* world) noexcept { return 100; }
 
 void PenaltyThemPrepare::calculateInfoForRoles() noexcept {
-    auto width = field.getFieldWidth();
-    auto length = field.getFieldLength();
+    const auto xPosition = -4 * control_constants::ROBOT_RADIUS;
+    const auto yPosition = Constants::STD_TIMEOUT_TO_TOP() ? field.getFieldWidth() / 2.2 : -field.getFieldWidth() / 2.2;
```
JaroKuiken: I tried to find if the STD_TIMEOUT_TO_TOP() ever changes, but as far as i can see it's always false. Did i miss something?

Frankavj: nope gaat fout ik ga t proberen te fixen

rolfvdhulst: This used to be not a 'constant' but a value we set through the interface. You typically don't change it but it's useful during a match to be able to change the side somewhere.

rolfvdhulst: What's up with the 2.2? That's what we typically call 'magic values'. It would be cleaner if you would write it like e.g:
double distanceFromCenterLine = field.getFieldWidth()/2.2; so you don't repeat yourself. and then do:
double yPosition = Constants::STD_TIMEOUT_TO_TOP() ? distanceFromCenterLine : -distanceFromCenterLine;
If you want the position to be slightly in the field, then it's probably better to do somethinglike:
field.getFieldWidth() / 2 + robotradius, because that will scale better for various field sizes.

rolfvdhulst: I do like that you made these variables const, however!

rolfvdhulst: Also, I think auto is a bit convoluted here, I would just use double.

src/stp/plays/PenaltyUsPrepare.cpp
```diff
@@ -17,19 +17,24 @@ PenaltyUsPrepare::PenaltyUsPrepare() : Play() {
...
         std::make_unique<role::Formation>(role::Formation("formation_8"))};
 }
 
 uint8_t PenaltyUsPrepare::score(world::World* world) noexcept { return 100; }
 
 void PenaltyUsPrepare::calculateInfoForRoles() noexcept {
-    auto width = field.getFieldWidth();
-    auto length = field.getFieldLength();
+    const auto xPosition = -4 * control_constants::ROBOT_RADIUS;
+    const auto yPosition = Constants::STD_TIMEOUT_TO_TOP() ? field.getFieldWidth() / 2.2 : -field.getFieldWidth() / 2.2;
```
rolfvdhulst: Same story as above; dont repeat yourself, and don't use magic values (e.g. 2.2).

rolfvdhulst (changes requested):

src/stp/plays/PenaltyUsPrepare.cpp
rolfvdhulst: you can remember this as; you want to be DRY (Dont Repeat Yourself), and not WET (Write Everything Twice)

src/stp/plays/PenaltyThemPrepare.cpp
```diff
@@ -28,23 +28,23 @@ PenaltyThemPrepare::PenaltyThemPrepare() : Play() {
...
-    stpInfos["formation_2"].setPositionToMoveTo(Vector2(-length / 8 + 2, width / 4));
-    stpInfos["formation_3"].setPositionToMoveTo(Vector2(-length / 8 + 2, -width / 4));
-    stpInfos["formation_4"].setPositionToMoveTo(Vector2(-length * 3 / 8 + 2, 0.0));
-    stpInfos["formation_5"].setPositionToMoveTo(Vector2(-length * 3 / 8 + 2, width / 5));
-    stpInfos["formation_6"].setPositionToMoveTo(Vector2(-length * 3 / 8 + 2, -width / 5));
-    stpInfos["formation_7"].setPositionToMoveTo(Vector2(-length / 4 + 2, width / 3));
-    stpInfos["formation_8"].setPositionToMoveTo(Vector2(-length / 4 + 2, -width / 3));
-    stpInfos["formation_9"].setPositionToMoveTo(Vector2(-length / 8 + 2, 0.0));
+    stpInfos["formation_0"].setPositionToMoveTo(Vector2(xPosition, yPosition));
+    stpInfos["formation_1"].setPositionToMoveTo(Vector2(xPosition - 4 * control_constants::ROBOT_RADIUS, yPosition));
```
JaroKuiken: Oh just saw this, maybe xPosition, then 2*xPosition, 3*xPosition, etc is nicer? Because what you're subtracting every time is 1 more xPosition

JaroKuiken: Or start with xPosition = 0, and then call the '-4*radius' thing an offset and subtract an extra offset every time? Idk, just a style thing not really important

src/stp/plays/PenaltyUsPrepare.cpp
```diff
@@ -33,8 +33,9 @@ PenaltyUsPrepare::PenaltyUsPrepare() : Play() {
 uint8_t PenaltyUsPrepare::score(world::World* world) noexcept { return 100; }
 
 void PenaltyUsPrepare::calculateInfoForRoles() noexcept {
-    const auto xPosition = -4 * control_constants::ROBOT_RADIUS;
-    const auto yPosition = Constants::STD_TIMEOUT_TO_TOP() ? field.getFieldWidth() / 2.2 : -field.getFieldWidth() / 2.2;
+    const double xPosition = -4 * control_constants::ROBOT_RADIUS;
+    const double distanceToCenterLine = field.getFieldWidth() / 2 - 2*control_constants::ROBOT_RADIUS;
```
JaroKuiken, line 37: Is this supposed to put them in a horizontal line above or below the x-axis of the field? Curious to know what the plan is

Frankavj: Each robot's y-position is negative half the field, minus the robot diameter. So the middle point of the robots is one robot diameter into the field from the sideline at the "bottom" of the field.

src/stp/plays/PenaltyThemPrepare.cpp
Frankavj: I copied this part of the code from the time out play, and I think you make a good point but if we use 2xPosition etc, it's maybe better to use a for-loop to remove double code, so ive implemented that now (also in time out)

JaroKuiken: Even nicer, that looks satisfying :+1:

src/stp/plays/PenaltyThemPrepare.cpp
```diff
@@ -28,23 +28,18 @@ PenaltyThemPrepare::PenaltyThemPrepare() : Play() {
...
-    stpInfos["formation_4"].setPositionToMoveTo(Vector2(-length * 3 / 8 + 2, 0.0));
-    stpInfos["formation_5"].setPositionToMoveTo(Vector2(-length * 3 / 8 + 2, width / 5));
-    stpInfos["formation_6"].setPositionToMoveTo(Vector2(-length * 3 / 8 + 2, -width / 5));
-    stpInfos["formation_7"].setPositionToMoveTo(Vector2(-length / 4 + 2, width / 3));
-    stpInfos["formation_8"].setPositionToMoveTo(Vector2(-length / 4 + 2, -width / 3));
-    stpInfos["formation_9"].setPositionToMoveTo(Vector2(-length / 8 + 2, 0.0));
+    const std::string formation = "formation_";
+    for(int i = 1; i <= 10; i++) {
+        stpInfos[formation + std::to_string(i - 1)].setPositionToMoveTo(Vector2(i * xPosition, yPosition));
+    }
```
rolfvdhulst, line 42: May be a bit overkill, but it's fine :)

rolfvdhulst (approved):

src/stp/plays/PenaltyThemPrepare.cpp
JaroKuiken: Was this fixed? As the constant is still used in the newer version

PR #1200, opened 2020-11-05
Control Module

JaroKuiken:
Creating Control Module class and moving acceleration&velocity limiting functionalities of the Skill class to the CM


include/roboteam_ai/control/ControlModule.h
```diff
@@ -0,0 +1,61 @@
...
+#include "world/views/RobotView.hpp"
+
+namespace rtt::ai::control {
+
+/**
+ * Control module as seen in the system architecture.
+ * Its job: Receive RobotCommands from Skills, check these and limit and/or change things wherever necessary.
+ * For now, it limits the velocity and acceleration that a Skill sends so the robot doesn't do a backflip.
+ * And it checks if the command that was sent makes sense (e.g., no kicking and chipping at the same time.
+ */
```
rolfvdhulst: Minor, but use the java-style documentation: (@brief, @author) for classes. That way your CLion can display it when you hover over the class :). I'm happy you wrote some documentation, by the way :D

src/control/ControlModule.cpp
```diff
@@ -0,0 +1,84 @@
...
+        command = command_;
+
+        // If we are not left, commands should be rotated (because we play as right)
+        if (!SETTINGS.isLeft()) {
+            rotateRobotCommand();
+        }
+
+        limitRobotCommand();
+
+        io::io.publishRobotCommand(command, data);
```
rolfvdhulst: I don't like this line; the idea is that we collect all of the commands done over a tick, and then send them all at once; What I'd like to do is store these commands in e.g. a vector, and then have a different function which actually publishes them all at once (and checks if there are no double commands!)

src/control/ControlModule.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    void ControlModule::limitVel() noexcept {
+        auto limitedVel = Vector2(command.vel().x(), command.vel().y());
+
+        limitedVel = control::ControlUtils::velocityLimiter(limitedVel);
+
+        if (std::isnan(limitedVel.x) || std::isnan(limitedVel.y)) {
+            // TODO: Pass on the skill name to Control Module as well, so this error can be given correctly
+            //RTT_ERROR("Robot will have NAN: " + std::string{stp::getName()} + "!\nrobot: " +
+            //          std::to_string(robot.value()->getId()))
```
rolfvdhulst: TODO? :) I'd just print it without the skill name for now.

src/control/ControlModule.cpp
rolfvdhulst: You don't have to worry about the publishing part if you don't want to, but if you can collect them in a vector that'd be nice.

rolfvdhulst (changes requested):
Code looks good! You basically just ported the code though; this should behave the exact same, which was not really the point. Don't forget to fix your TODO's :). Also, I don't like the statics (surprise), but for now this is fine.  We're unlikely to create more than one instance of the control module anyways.

JaroKuiken:
> Code looks good! You basically just ported the code though; this should behave the exact same, which was not really the point.

Ah you mean storing all the RobotCommands until we have 11 and then send all of them at once? Ill get on that now

> Don't forget to fix your TODO's :). Also, I don't like the statics (surprise), but for now this is fine. We're unlikely to create more than one instance of the control module anyways.

Ah yeah, will do. Ill leave it as statics for now, maybe fix it later.

src/control/ControlModule.cpp
JaroKuiken: Ill first start with trying to store them in a vector, then see if i feel like trying the publishing :P

rolfvdhulst (approved):

Floris-Hoek (approved):

include/roboteam_ai/control/ControlModule.h
```diff
@@ -0,0 +1,61 @@
...
+
+    public:
+        /**
+         * Limits the current robot command and adds it to the list of commands to be sent
+         */
+        static void addRobotCommand(std::optional<rtt::world::view::RobotView> robot, const proto::RobotCommand& command, const rtt::world::World *data) noexcept;
+        /**
+         *
+         */
+        static void sendAllCommands();
```
Yuhanun, line 57: Glorified global variables?

include/roboteam_ai/utilities/IOManager.h
```diff
@@ -32,54 +28,29 @@ using namespace rtt::world;
...
-
-    static std::mutex worldStateMutex;
-    static std::mutex geometryMutex;
-    static std::mutex robotFeedbackMutex;
-    static std::mutex refereeMutex;
-    static std::mutex demoMutex;
+    proto::State getState();
 
-    bool hasReceivedGeom = false;
+    mutable std::shared_mutex stateMutex;
```
Yuhanun: why is this a shared_mutex and please don't mark it mutable, keep the class as mutable instead.

Yuhanun: you shouldn't even need this to be `mutable`, if you mark the class as `const` you can't modify the data anyway and then you don't need a mutex since it's read-only.

src/control/ControlModule.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+        // Limit robot velocity when the robot has the ball
+        if (robot->hasBall() && limitedVel.length() > stp::control_constants::MAX_VEL_WHEN_HAS_BALL) {
+            // Clamp velocity
+            limitedVel = control::ControlUtils::velocityLimiter(limitedVel, stp::control_constants::MAX_VEL_WHEN_HAS_BALL,
+                                                                0.0, false);
+        }
+
+        command.mutable_vel()->set_x(static_cast<float>(limitedVel.x));
+        command.mutable_vel()->set_y(static_cast<float>(limitedVel.y));
```
Yuhanun, line 46: i wonder why we're using double, when we cast it to float before sending :P

src/interface/widgets/ManualControlWidget.cpp
```diff
@@ -12,7 +12,7 @@ ManualControlWidget::ManualControlWidget(QWidget *parent) : QWidget(parent) {
     allowCheckBox->setText("Allow Manual takeover");
     allowCheckBox->setChecked(false);
     layout->addWidget(allowCheckBox);
-    manager = new rtt::input::JoystickManager(&rtt::ai::io::io);
+    manager = new rtt::input::JoystickManager();
```
Yuhanun: no.

src/manual/JoystickManager.cpp
```diff
@@ -1,12 +1,13 @@
...
 #include <roboteam_utils/Print.h>
 #include "include/roboteam_ai/world/World.hpp"
+#include "world/views/RobotView.hpp"
 
 using namespace std::chrono;
 
 namespace rtt::input {
 
-JoystickManager::JoystickManager(ai::io::IOManager *ioManager) : ioManager(ioManager) { }
+JoystickManager::JoystickManager() { }
```
Yuhanun: = default;

src/utilities/IOManager.cpp
```diff
@@ -10,128 +10,95 @@
...
-        robotCommandPublisher = new proto::Publisher<proto::RobotCommand>(proto::ROBOT_COMMANDS_SECONDARY_CHANNEL);
+        robotCommandPublisher = new proto::Publisher<proto::AICommand>(proto::ROBOT_COMMANDS_SECONDARY_CHANNEL);
         settingsPublisher = new proto::Publisher<proto::Setting>(proto::SETTINGS_SECONDARY_CHANNEL);
     } else {
-        feedbackSubscriber = new proto::Subscriber<proto::RobotFeedback>(proto::FEEDBACK_PRIMARY_CHANNEL, &IOManager::handleFeedback, this);
-        robotCommandPublisher = new proto::Publisher<proto::RobotCommand>(proto::ROBOT_COMMANDS_PRIMARY_CHANNEL);
+        robotCommandPublisher = new proto::Publisher<proto::AICommand>(proto::ROBOT_COMMANDS_PRIMARY_CHANNEL);
         settingsPublisher = new proto::Publisher<proto::Setting>(proto::SETTINGS_PRIMARY_CHANNEL);
     }
+    central_server_connection = new networking::PairReceiver<16970>();
```
Yuhanun, line 35: no.

Yuhanun: Doesn't need to be heap allocated in general.

include/roboteam_ai/utilities/IOManager.h
```diff
@@ -32,54 +28,29 @@ using namespace rtt::world;
...
-    proto::Subscriber<proto::RobotFeedback> *feedbackSubscriber;
-    void handleFeedback(proto::RobotFeedback &feedback);
-
-    proto::Publisher<proto::RobotCommand> *robotCommandPublisher;
+    proto::Publisher<proto::AICommand> *robotCommandPublisher;
     proto::Publisher<proto::Setting> *settingsPublisher;
 
     rtt::ai::Pause *pause;
 
+    rtt::networking::PairReceiver<16970> *  central_server_connection;
```
Yuhanun, line 40:
```suggestion
    rtt::networking::PairReceiver<16970>  central_server_connection{ };
```

src/utilities/IOManager.cpp
Yuhanun, line 35:
```suggestion
```

include/roboteam_ai/utilities/IOManager.h
Yuhanun: Ah nevermind i see what's going on now, although I can't say I like it :P

src/utilities/IOManager.cpp
```diff
@@ -10,128 +10,95 @@
...
+      received = false;
+      //we don't print the errors as they mark there are no more messages
+    }
+  }
+  if(numReceivedMessages>0){
+    std::cout<<"received " << numReceivedMessages <<" packets from central server"<<std::endl;
+  }
+  //TODO: actually change settings at the relevant places within our AI
+  //then, send the current state once
+  //TODO: make sure to write/add relevant debug information/visualizations (strategy debug, etc.)
```
Yuhanun: todo?>

src/utilities/Pause.cpp
```diff
@@ -24,8 +25,9 @@ void Pause::haltRobots(rtt::world::World const* data) {
         cmd.set_dribbler(0);
         cmd.set_use_angle(1);
         cmd.set_w(static_cast<float>(robot->getAngle()));
-        io::io.publishRobotCommand(cmd, data);
+        commands.push_back(cmd);
```
Yuhanun:
```suggestion
        commands.emplace_back(std::move(cmd));
```

src/world/Field.cpp
```diff
@@ -36,7 +36,7 @@ void Field::initFieldLines(const proto::SSL_GeometryFieldSize &sslFieldSize) {
 }
 
 void Field::initFieldArcs(const proto::SSL_GeometryFieldSize &sslFieldSize) {
-    for (const proto::SSL_FieldCicularArc &arc : sslFieldSize.field_arcs()) {
+    for (const proto::SSL_FieldCircularArc &arc : sslFieldSize.field_arcs()) {
```
Yuhanun:
```suggestion
    for (auto const& arc : sslFieldSize.field_arcs()) {
```

Yuhanun (changes requested):
Please don't heap allocate where it's not necessary

src/control/ControlModule.cpp
rolfvdhulst: I mean, it doesn't matter much. Computing things in double is usually faster though, unless you're sending big matrices or large amount of numeric data where the memory overhead wins.

include/roboteam_ai/control/ControlModule.h
rolfvdhulst: Yeah for now. This was the easiest way to do it because the previous alternative was also just a global variable/static. But personally, I'd still like to change this in the future.

src/utilities/IOManager.cpp
rolfvdhulst: I don't like this, because now the connection is started at an undetermined time when the application is started. I'd rather have a heap allocation so it's easier in the future to close and restart the connection rather than this model.

rolfvdhulst: It makes it easier to see when and how the connection is started.

rolfvdhulst: If it was only a constructor, that's fine, but this also starts the connection, so that's why I disagree.

include/roboteam_ai/control/ControlModule.h
rolfvdhulst: (This was Jaro's code, mostly)

src/utilities/IOManager.cpp
rolfvdhulst: Yes? I can't really do these unless it's clear what's supposed to be in the interface.

include/roboteam_ai/utilities/IOManager.h
rolfvdhulst: No, you're right this is not necessary. I changed it during testing because I thought there was a deadlock, but it turned out to be a ZMQ blocking-call related problem.

rolfvdhulst: I changed it back to a normal std::mutex again.

rolfvdhulst: I'll change it back to a normal std::mutex again.

include/roboteam_ai/control/ControlModule.h
Yuhanun: Sure.

include/roboteam_ai/utilities/IOManager.h
Yuhanun: Sure.

src/utilities/IOManager.cpp
Yuhanun: Sure, fair enough.

Yuhanun (approved):
LGTM

PR #1201, opened 2020-11-05
System refactor

rolfvdhulst:


JaroKuiken:
#1200 has all these changes and more, so this PR is outdated

PR #1208, opened 2021-03-09
Feature/world objects

Floris-Hoek:


CMakeLists.txt
```diff
@@ -178,7 +178,7 @@ set(CONTROL_SOURCES
         ${PROJECT_SOURCE_DIR}/src/control/positionControl/pathPlanning/PathPlanningAlgorithm.cpp
         ${PROJECT_SOURCE_DIR}/src/control/positionControl/pathTracking/PathTrackingAlgorithm.cpp
         ${PROJECT_SOURCE_DIR}/src/control/positionControl/PositionControlUtils.cpp
-        )
+        src/control/positionControl/BBTrajectories/WorldObjects.cpp include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h)
```
rolfvdhulst: These should be separated on multiple lines, and don't put .h files in the cmakelists.

CMakeLists.txt
```diff
@@ -259,7 +259,10 @@ set(AI_SOURCES
         ${INTERFACE_SOURCES}
         ${WORLD_SOURCES}
         ${MANUAL_SOURCES}
-        )
+        include/roboteam_ai/control/positionControl/BBTrajectories/BBTrajectory1D.h
+        src/control/positionControl/BBTrajectories/BBTrajectory1D.cpp
+        include/roboteam_ai/control/positionControl/BBTrajectories/BBTrajectory2D.h
+        src/control/positionControl/BBTrajectories/BBTrajectory2D.cpp src/control/positionControl/BBTrajectories/WorldObjects.cpp include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h src/control/positionControl/BBTrajectories/WorldObjects.cpp include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h src/control/positionControl/BBTrajectories/WorldObjects.cpp include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h)
```
rolfvdhulst: No .h files, and use multiple lines.

src/world/World.cpp
```diff
@@ -111,4 +111,13 @@ namespace rtt::world {
...
     size_t World::getHistorySize() const noexcept { return history.size(); }
 
+     void World::forwardWorldToBB() {
+        BB::WorldObjects::setWorld(this);
+    }
+
+//    void World::forwardRobots(std::vector<view::RobotView> robots){
+//        std::transform(getWorld()->getRobotsNonOwning().begin(), getWorld()->getRobotsNonOwning().end(), robots.begin(), [](const auto& robot) -> rtt::world::view::RobotView { return (robot); });
+//        BB::WorldObjects::setRobots(robots);
+//    }
```
rolfvdhulst: No commented code, just remove it.

src/world/World.cpp
```diff
@@ -111,4 +111,13 @@ namespace rtt::world {
 
     size_t World::getHistorySize() const noexcept { return history.size(); }
 
+     void World::forwardWorldToBB() {
+        BB::WorldObjects::setWorld(this);
+    }
```
rolfvdhulst: No statics :(. Is it a possibility you just pass the bang-bang control module a world pointer every tick?

src/utilities/Dealer.cpp
```diff
@@ -91,6 +91,10 @@ double Dealer::scoreForFlags(const std::vector<Dealer::DealerFlag> &dealerFlags,
 
 // Get the score of one flag for a role for one robot
 double Dealer::getScoreForFlag(v::RobotView robot, Dealer::DealerFlag flag) {
+    std::cout << "RobotId: " << robot->getId() << std::endl;
+    std::cout << "Priority: " << static_cast<int>(flag.priority) << endl;
+    //TODO: Make sure we eventually use priorities when creating different BBTrajectories
```
rolfvdhulst: you could make this todo an issue.

src/stp/tactics/TestTactic.cpp
```diff
@@ -24,20 +24,10 @@ std::optional<StpInfo> TestTactic::calculateInfoForSkill(StpInfo const &info) no
...
-        case 2:
-            skillStpInfo.setPositionToMoveTo(Vector2(-length / 8, -width / 8));
-            break;
-        case 3:
-            skillStpInfo.setPositionToMoveTo(Vector2(-length / 8, width / 8));
-            break;
-        case 4:
-            skillStpInfo.setPositionToMoveTo(Vector2(length / 8, width / 8));
-            skills.reset();
+            skillStpInfo.setPositionToMoveTo(Vector2(2, 2));
```
rolfvdhulst: All the changes in this file are just testing artifacts, I think? Probably just undo them before you merge.

include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h
```diff
@@ -0,0 +1,123 @@
...
+    class WorldObjects {
+    private:
+        const rtt::ai::rtt_world::Field *field = nullptr;
+
+        std::vector<Vector2> calculatedPaths[11] = {}; // Holds the paths of the different robots
+        rtt::ai::GameStateManager gameStateManager;
+        rtt::ai::GameState gameState;
+        rtt::ai::RuleSet ruleset;
+
+        static world::World *world;
```
rolfvdhulst: Does this need to be static? :( Please try to not make it static.

include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h
```diff
@@ -0,0 +1,123 @@
...
+    struct CollisionData {
+        Vector2 obstaclePosition;
+        Vector2 collisionPosition;
+    };
+
+    class WorldObjects {
+    private:
+        const rtt::ai::rtt_world::Field *field = nullptr;
+
+        std::vector<Vector2> calculatedPaths[11] = {}; // Holds the paths of the different robots
```
rolfvdhulst: This is not how you initialize a vector; in the constructor of world objects, you can do calculatedPaths.reserve(11) or preferably calculatedPaths.reserve(16), and just puth std::vector<Vector2> calculatedPaths; here. I'm not even sure personally if this compiles, but I've never seen anyone use this syntax before.

include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h
```diff
@@ -0,0 +1,123 @@
...
+
+        void setField(const rtt::ai::rtt_world::Field &field);
+
+        // Checks if a specified robot can enter the defense area
+        bool canEnterDefenseArea(int robotId);
+
+        // Checks if a specified robot can move out of the field
+        bool canMoveOutsideField(int robotId);
+
+        static void setWorld(world::World *world_);
```
rolfvdhulst: no statics :(

include/roboteam_ai/world/World.hpp
```diff
@@ -162,6 +162,8 @@ class World {
      */
     [[nodiscard]] ai::control::PositionControl *getRobotPositionController() noexcept;
 
+    void forwardWorldToBB();
```
rolfvdhulst: void functions that do not take any arguments are usually bad because it's unclear what is happening with the internal state of world. This function should probably be const, as well, as you don't want BB trajectories editing the world.

src/control/positionControl/BBTrajectories/WorldObjects.cpp
```diff
@@ -0,0 +1,183 @@
...
+        this->calculateDefenseAreaCollisions(collisionDatas, collisionTimes, pathPoints, robotId, timeStep);
+
+        // Check if robot is closer to the ball than it is allowed to be
+        this->calculateBallCollisions(collisionDatas, collisionTimes, pathPoints, timeStep);
+
+        // Loop through all pathPoints for each enemy robot, and check if a point in the path will collide with an enemy robot
+        this->calculateEnemyRobotCollisions(BBTrajectory, collisionDatas, collisionTimes, pathPoints, timeStep);
+
+        // For each path already calculated, check if this path collides with those paths
+        this->calculateOurRobotCollisions(collisionDatas, collisionTimes, pathPoints, robotId, timeStep);
```
rolfvdhulst: All of the this-> should be unneeded.

src/control/positionControl/BBTrajectories/WorldObjects.cpp
```diff
@@ -0,0 +1,183 @@
...
+                        collisionTimes.emplace_back(j * timeStep);
+                        return;
+                    }
+                }
+            }
+        }
+    }
+
+
+    void WorldObjects::setField(const rtt::ai::rtt_world::Field &field_) { this->field = &field_; }
```
rolfvdhulst: This is very dangerous; if the field pointer in world changes for whatever reason, you're going to get a segfault at some point in this class because you copied it. This could happen in many circumstances. I'd personally probably prefer field being passed to every function, rather than relying on the lifetime of your pointer being safe like you do now.

include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h
```diff
@@ -0,0 +1,123 @@
...
+         * @param collisionDatas, std::vector which rtt::BB::CollisionData can be added to
+         * @param collisionTimes, std::vector which times can be added to
+         * @param pathPoints, std::vector with path points
+         * @param robotId
+         * @param timeStep in seconds
+         */
+        void calculateOurRobotCollisions(std::vector<CollisionData> &collisionDatas, std::vector<double> &collisionTimes,
+                                         const std::vector<Vector2> &pathPoints, int robotId, double timeStep);
+
+        void setField(const rtt::ai::rtt_world::Field &field);
```
rolfvdhulst: This seems dangerous, as you have a pointer field; it's probably better to pass world and field pointers and/or/references to every single function call.

src/control/positionControl/BBTrajectories/WorldObjects.cpp
```diff
@@ -0,0 +1,183 @@
...
+            for (int i = 0; i < theirRobots.size() - 1; i++) {
+                double currentTime = j * timeStep;
+                double maxCollisionCheckTime = 2;
+                if (currentTime <= maxCollisionCheckTime) {
+                    // TODO: Currently enemy position in future is calculated with simple model (x = x + v*t), maybe improve?
+                    Vector2 theirVel = theirRobots[i]->getVel();
+                    Vector2 theirPos = theirRobots[i]->getPos() /*+ theirVel * currentTime*/;
+                    Vector2 posDif = BBTrajectory.getPosition(currentTime) -
+                                     theirPos; //TODO: Is currentTime defined the same way as the time from the BBTrajectory time?
+                    if (posDif.length() < 3 /*<-- kawaiii*/ * ai::Constants::ROBOT_RADIUS_MAX()) {
```
rolfvdhulst: No kawaii in our code base plz

src/control/positionControl/BBTrajectories/WorldObjects.cpp
```diff
@@ -0,0 +1,183 @@
...
+                        if (abs(projectLength) > 1.5 && theirVel.length() < ourVel.length()) {
+                            collisionDatas.emplace_back(CollisionData{theirPos, pathPoints[j]});
+                            collisionTimes.emplace_back(currentTime);
+                            return;
+                        }
+                    }
+                }
+            }
+        }
+    }
```
rolfvdhulst: This has a lot of levels of indentation and is difficult to read. Put some of the code into functions.

src/control/positionControl/BBTrajectories/WorldObjects.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    void
+    WorldObjects::calculateOurRobotCollisions(std::vector<CollisionData> &collisionDatas, std::vector<double> &collisionTimes,
+                                              const std::vector<Vector2> &pathPoints, int robotId, double timeStep) {
+
+        int ourRobotAmount = world->getWorld()->getUs().size();
+
+        for (int j = 0; j < pathPoints.size(); j++) {
+            for (int i = 0; i < ourRobotAmount; i++) {
```
rolfvdhulst: Can be confusing, since people usually use i for the outer loop and j for the inner loop.

src/control/positionControl/BBTrajectories/WorldObjects.cpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    void
+    WorldObjects::calculateEnemyRobotCollisions(rtt::BB::BBTrajectory2D BBTrajectory, std::vector<CollisionData> &collisionDatas,
+                                                std::vector<double> &collisionTimes, const std::vector<Vector2> &pathPoints,
+                                                double timeStep) {
+        auto theirRobots = world->getWorld()->getThem();
+
+        for (int j = 0; j < pathPoints.size(); j++) {
+            for (int i = 0; i < theirRobots.size() - 1; i++) {
```
rolfvdhulst: Typically, people will reverse the i and j here, using i for the outer loop and j for the inner loop.

src/control/positionControl/CollisionDetector.cpp
```diff
@@ -52,8 +52,8 @@ std::vector<Vector2> CollisionDetector::getRobotPositions() {
     return robotPositions;
 }
 
-void CollisionDetector::setField(const rtt::world::Field& field) { this->field = &field; }
+void CollisionDetector::setField(const rtt::world::Field& field_) { this->field = &field_; }
```
rolfvdhulst, line 55: This will give the same problem as above; very bug and segfault prone.

rolfvdhulst: Though it seems, this was not your fault as the previous team did this.

src/stp/plays/TestPlay.cpp
```diff
@@ -5,6 +5,7 @@
 #include "stp/plays/TestPlay.h"
 
 #include "stp/roles/TestRole.h"
+#include "stp/roles/Halt.h"
```
rolfvdhulst: You should probably revert all the changes in this file, seems they are just for testing.

src/stp/skills/GoToPos.cpp
```diff
@@ -16,13 +16,21 @@ Status GoToPos::onUpdate(const StpInfo &info) noexcept {
...
 
+    std::pair<RobotCommand, std::optional<Vector2>> robotCommandCollisionPair;
     // Calculate commands from path planning and tracking
-    auto robotCommand = info.getCurrentWorld()->getRobotPositionController()->computeAndTrackPath(
+    // _______Use this for the old pathplanning and tracking_______
+//    robotCommandCollisionPair.first = info.getCurrentWorld()->getRobotPositionController()->computeAndTrackPath(
+//        info.getField().value(), info.getRobot().value()->getId(), info.getRobot().value()->getPos(), info.getRobot().value()->getVel(), targetPos, info.getPidType().value());
+    // _______Use this one for the BBT pathplanning_______
+    robotCommandCollisionPair = info.getCurrentWorld()->getRobotPositionController()->computeAndTrackPathBBT(
         info.getField().value(), info.getRobot().value()->getId(), info.getRobot().value()->getPos(), info.getRobot().value()->getVel(), targetPos, info.getPidType().value());
```
rolfvdhulst: No commented code; prefer writing something like this instead;

bool use_old_pathplanning = false;
if(use_old_pathplanning){
... (old code)...
}else{
... (new code) ...
}

rolfvdhulst: This way, we can be sure the old code also still compiles.

src/control/positionControl/PositionControl.cpp
```diff
@@ -6,44 +6,215 @@
...
+                                 Vector2 targetPosition, stp::PIDType pidType) {
+        //TODO: Create a function in WorldObjects which can give the position of the first collision to STP
+        //TODO: find a good value for the timeStep
+        double timeStep = 0.1;
+
+        worldObjects.setField(field);
+        std::optional<BB::CollisionData> firstCollision;
+        std::pair<RobotCommand, std::optional<Vector2>> robotCommandCollisionPair;
+        // Currently calculate all paths again on each tick because the way the path is used in control is not made for the BBT
+        if (true || (!computedPathsBB.contains(robotId) ||
```
rolfvdhulst: the true is suspicious? Is it related to the comment above?

src/control/positionControl/PositionControl.cpp
```diff
@@ -6,44 +6,215 @@
...
+        // If you are closer to the target than the first point of the approximated path, remove it
+        if((targetPosition - currentPosition).length() < (targetPosition-computedPaths[robotId].front()).length()){
+            computedPaths[robotId].erase(computedPaths[robotId].begin());
+        }
+
+        robotCommandCollisionPair.first = RobotCommand();
+        robotCommandCollisionPair.first.pos = computedPaths[robotId].front();
+        Position trackingVelocity = pathTrackingAlgorithm.trackPathDefaultAngle(currentPosition, currentVelocity,computedPaths[robotId], robotId,pidType);
+        robotCommandCollisionPair.first.vel = Vector2(trackingVelocity.x, trackingVelocity.y);
+        robotCommandCollisionPair.first.angle = trackingVelocity.rot;
```
rolfvdhulst: This seems like you need to define a type instead of using a std::pair. You should almost never be doing this, as you have std::make_pair for making pairs.

src/control/positionControl/PositionControl.cpp
```diff
@@ -6,44 +6,215 @@
...
+
+        return robotCommandCollisionPair;
```
rolfvdhulst: This function is perhaps a bit long.

src/control/positionControl/PositionControl.cpp
```diff
@@ -6,44 +6,215 @@
...
 
-    RobotCommand command = RobotCommand();
-    command.pos = computedPaths[robotId].front();
-    Position trackingVelocity = pathTrackingAlgorithm.trackPathDefaultAngle(currentPosition, currentVelocity, computedPaths[robotId], robotId, pidType);
-    command.vel = Vector2(trackingVelocity.x, trackingVelocity.y);
-    command.angle = trackingVelocity.rot;
+        // PointToDrawFrom is picked by drawing a line from the target position to the obstacle and extending that
+        // line further towards our currentPosition.
+        Vector2 pointToDrawFrom = firstCollision->obstaclePosition +
+                                  (firstCollision->obstaclePosition - targetPosition).normalize() * 1;
```
rolfvdhulst: redundant *1?

src/control/positionControl/PositionControl.cpp
```diff
@@ -6,44 +6,215 @@
...
+        }
+        return intermediatePoints;
+    }
+
+    std::priority_queue<std::pair<double, Vector2>, std::vector<std::pair<double, Vector2>>, std::greater<>>
+    PositionControl::scoreIntermediatePoints(std::vector<Vector2> &intermediatePoints,
+                                            std::optional<BB::CollisionData> &firstCollision) {
+        double intermediatePointScore;
+        std::priority_queue<std::pair<double, Vector2>, std::vector<std::pair<double, Vector2>>, std::greater<>> intermediatePointsSorted;
+        for (auto i : intermediatePoints) {
```
rolfvdhulst: Use const auto&; now you're copying, which is unnecessarily expensive.

src/control/positionControl/PositionControl.cpp
```diff
@@ -6,44 +6,215 @@
...
+                auto newStartCollisions = worldObjects.getFirstCollision(intermediateToTarget, robotId);
+
+                if (newStartCollisions.has_value()) {
+                    continue;
+                } else {
+                    return intermediateToTarget;
+                }
+            }
+        }
+        return {};
```
rolfvdhulst: Is this returning nullopt or an empty trajectory? Can be very unclear, should probably make it more explicit.

src/control/positionControl/PositionControl.cpp
```diff
@@ -6,44 +6,215 @@
...
+    }
+
+    std::optional<BB::BBTrajectory2D>
+    PositionControl::calculatePathFromNewStart(std::optional<BB::CollisionData> intermediatePathCollision,
+                                               BB::BBTrajectory2D pathToIntermediatePoint,
+                                               Vector2 &targetPosition, int robotId, double timeStep) {
+        BB::BBTrajectory2D intermediateToTarget;
+        if (!intermediatePathCollision.has_value()) {
+            timeStep *= 2;
+            for (int i = 0; i < floor(pathToIntermediatePoint.getTotalTime() / timeStep); i++) {
```
rolfvdhulst: Put the floor calculation outside of the loop.

rolfvdhulst (changes requested):

src/control/positionControl/PositionControl.cpp
Floris-Hoek: Yes, I did not get it working yet. Right now it only works when recalculating the path every time but I wanted to have the logic in place when that is fixed. So the true is only there for now to calculate the path every tick.

Floris-Hoek: I added an extra comment to clarify it

src/control/positionControl/PositionControl.cpp
Floris-Hoek: 1 is the length of the extension. Made it more readable by adding an extension variable with value 1

src/control/positionControl/CollisionDetector.cpp
Floris-Hoek: Do you want me to change this? It is currently only used by the old pathplanning

include/roboteam_ai/control/positionControl/BBTrajectories/WorldObjects.h
Floris-Hoek: This is how you would initialize an array right? But I changed the way the paths are stored to an unordered_map. The calculatedPaths is no longer a member of WorldObjects and is now passed by the functions.

Yuhanun: What this does is initialize an array of 11 empty vectors.

src/control/positionControl/PositionControl.cpp
Floris-Hoek: I removed the robotCommandCollisionPair and replaced it with a struct in WorldObjects

src/control/positionControl/PositionControl.cpp
Floris-Hoek: Discussed it with Rolf and is okay for now

src/control/positionControl/BBTrajectories/WorldObjects.cpp
Floris-Hoek: Made the function more readable but left it as one function

JaroKuiken (approved):
The TODO's could be done, but they don't look necessary imo

rolfvdhulst (approved):
LGTM now.

PR #1209, opened 2021-03-10
STP < Upgrading of Dealer

mmlievenseRTT:
- [ ] The tests are passing

### Summary
Fix for Dealer. Making a split between Priority of roles and Weight of flags.

### Issues before
1. Dealer had a weird and flunky method of choosing a Keeper (add a field between everyone and the location except for the robot with the Keeper ID).
2. Dealer would not consistently pick the closest Robot to the ball when asked (very important fix).
3. FlagPriorities had in them the priority of a role and the weight for a flag/factor (like close to goal).

### Changes
1. The role map has now a difference between role priority and factor weights. The rolls are given away in order from highest priority (keeper) to lowest priority (low_priority). This is done after calculating the scoreMatrix (only is does once) by removing given-away roles and assigned robots from the scorelist.
2. Similar to the keeper assignment a boolean helps with this issue. 
3. In the Play the user can given a role a priority and flags weights seperatly.


CMakeLists.txt
```diff
@@ -100,41 +100,41 @@ set(ROLES_SOURCES
...
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/Halt.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/KickOffThem.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/KickOffThemPrepare.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/KickOffUs.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/KickOffUsPrepare.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/PenaltyThem.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/PenaltyThemPrepare.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/PenaltyUs.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/PenaltyUsPrepare.cpp
+#        ${PROJECT_SOURCE_DIR}/src/stp/plays/referee_specific/TimeOut.cpp
```
Yuhanun, line 137: What do these have to do with the dealer?

include/roboteam_ai/utilities/Dealer.h
```diff
@@ -48,12 +50,22 @@ class Dealer {
...
+        double sumWeights;
+    };
+
     struct DealerFlag {
         DealerFlagTitle title;
         DealerFlagPriority priority;
         explicit DealerFlag(DealerFlagTitle title, DealerFlagPriority priority);
     };
-    using FlagMap = std::map<std::string, std::vector<DealerFlag>>;
+    using FlagMap = std::map<std::string, std::pair<DealerFlagPriority, std::vector<DealerFlag>>>;
```
Yuhanun: std::unordered_map + wrap the pair in a struct like RobotRoleScore n stuff

```cpp
struct ... {
    DealerFlagPriority prio;
    std::vector<DealerFlag> flags;
};
```

include/roboteam_ai/utilities/Dealer.h
```diff
@@ -93,15 +104,15 @@ class Dealer {
      * @param stpInfoMap
      * @return The score matrix
      */
-    std::vector<std::vector<double>> getScoreMatrix(const std::vector<v::RobotView> &allRobots, const FlagMap &flagMap,
+    std::vector<std::pair<std::vector<double>, int>> getScoreMatrix(const std::vector<v::RobotView> &allRobots, const FlagMap &flagMap,
```
Yuhanun: wrap the pair in a struct.

src/utilities/Dealer.cpp
```diff
@@ -19,93 +23,124 @@ Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priorit
...
+            }
         }
     }
-    return result;
+    return output;
 }
 
 // Populate a matrix with scores
-std::vector<vector<double>> Dealer::getScoreMatrix(const std::vector<v::RobotView> &allRobots, const Dealer::FlagMap &flagMap,
+std::vector<std::pair<std::vector<double>, int>> Dealer::getScoreMatrix(const std::vector<v::RobotView> &allRobots, const Dealer::FlagMap &flagMap,
```
Yuhanun: wrap this type in a struct.

src/utilities/Dealer.cpp
```diff
@@ -19,93 +23,124 @@ Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priorit
...
+                        output.insert({roleNames[originalRolesIndex[j]], allRobots[originalIDsIndex.back()]});
+                    }
+                }
+                if (output.size() == allRobots.size()) return output;               // case if there are less then 11 bots to distribute
+                std::sort(currentIDs.begin(), currentIDs.end());                    // Sort to delete from back to front
+                // Delete assigned roles and robots from score
+                for (int j = currentRoles.size() - 1; j >= 0; j--) {
+                    scores.erase(scores.begin() + currentRoles[j]);                 // remove role from score (col)
+                    indexRoles.erase(indexRoles.begin() + currentRoles[j]);         // remove from index list
+                }
```
Yuhanun: there's a .clear() 🤔

src/utilities/Dealer.cpp
```diff
@@ -19,93 +23,124 @@ Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priorit
...
             // The better the flags, the lower the score
-            ScoreResult flagScore = scoreForFlags(dealerFlags, robot);
-            row.push_back((distanceScore + flagScore.score)/(flagScore.weight+1));
+            /// dealerFlag.first = Priority & .second = Flags
+            auto robotScore = getRobotScoreForRole(dealerFlags.second, robot);
+            // Simple normalizer. DistanceScore has weight 1, the other factors can have various weights.
+            role.push_back((robotDistanceScore + robotScore.sumScore) / (robotScore.sumWeights + 1));  // the +1 is the distanceScore weight
         }
-        scores.push_back(row);
+        scores.emplace_back(role, (int) dealerFlags.first);
```
Yuhanun:
```suggestion
        scores.emplace_back(role, static_cast<int>(dealerFlags.first));
```

src/utilities/Dealer.cpp
```diff
@@ -19,93 +23,124 @@ Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priorit
...
-}
+    std::unordered_map<std::string, v::RobotView> output;
+    std::vector<std::pair<std::vector<double>, int>> scores = getScoreMatrix(allRobots, flagMap, stpInfoMap);
+    // Make index of roles and ID to keep track which are the original indexes of each
+    std::vector<int> indexRoles;
+    std::vector<int> indexID;
+    for (int i = 0; i < scores.size(); i++) {
+        indexRoles.push_back(i);
+        indexID.push_back(i);
+    }
```
Yuhanun: reserve maybe?

Yuhanun (changes requested):
Those types aren't okay, we should really wrap them, you've got some long functions too, try split those up.

include/roboteam_ai/utilities/Dealer.h
mmlievenseRTT: Will fix, didn't think of the idea that i could use a struct in a map

src/utilities/Dealer.cpp
mmlievenseRTT: Result of pair in map, will fix

src/utilities/Dealer.cpp
mmlievenseRTT: It has to remove only certain elements of the vector, for example: element: erase 7 4 2 from 1 2 3 4 5 6 7 8 makes 1 3 5 6 8.

src/utilities/Dealer.cpp
mmlievenseRTT: Thanks!

src/utilities/Dealer.cpp
mmlievenseRTT: i'll look into that

CMakeLists.txt
mmlievenseRTT: I am actually not sure why the CMakeList didnt get branched, ill make sure it does next time.

src/utilities/Dealer.cpp
```diff
@@ -19,93 +23,126 @@ Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priorit
...
+void Dealer::distribute_remove(DealerDistribute& current, std::vector<int>& indexRoles, std::vector<int>& indexID, std::vector<RoleScores>& scores){
+    std::sort(current.currentIDs.begin(), current.currentIDs.end());                    // Sort to delete from back to front
+    // Delete assigned roles and robots from score
+    for (int j = current.currentRoles.size() - 1; j >= 0; j--) {
+        scores.erase(scores.begin() + current.currentRoles[j]);                 // remove role from score (col)
+        indexRoles.erase(indexRoles.begin() + current.currentRoles[j]);         // remove from index list
+    }
+    for (int col = 0; col < scores.back().robotScores.size(); col++) {        // go through each score role (row)
+        for (int j = current.currentIDs.size() - 1; j >= 0; j--) {
+            scores[col].robotScores.erase(scores[col].robotScores.begin() + current.currentIDs[j]);     // remove the robot
```
mmlievenseRTT: > Those types aren't okay, we should really wrap them, you've got some long functions too, try split those up.

Like this?

mmlievenseRTT (commented):
Refering to @Yuhanun 's comment on long function

src/utilities/Dealer.cpp
```diff
@@ -19,93 +23,126 @@ Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priorit
...
-                                                                         const std::vector<int> &assignment) const {
-    vector<string> orderedRoleNames;
+void Dealer::distribute_init(std::vector<int>& indexRoles, std::vector<int>& indexID, std::vector<std::string>& roleNames, const Dealer::FlagMap &flagMap) {
+    indexRoles.reserve(flagMap.size());
+    indexID.reserve(flagMap.size());
+    roleNames.reserve(flagMap.size());
+    for (int i = 0; i < flagMap.size(); i++) {
+        indexRoles.push_back(i);
+        indexID.push_back(i);
+    }
```
rolfvdhulst: This reads a bit like a constructor, maybe consider making a data format for this (but this is fine, too).

src/utilities/Dealer.cpp
```diff
@@ -19,93 +23,126 @@ Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priorit
...
+    for (int col = 0; col < scores.back().robotScores.size(); col++) {        // go through each score role (row)
+        for (int j = current.currentIDs.size() - 1; j >= 0; j--) {
+            scores[col].robotScores.erase(scores[col].robotScores.begin() + current.currentIDs[j]);     // remove the robot
         }
     }
-    return result;
+    for (int j = current.currentIDs.size() - 1; j >= 0; j--) {
+        indexID.erase(indexID.begin() + current.currentIDs[j]);                 // remove from index list
+    }
```
rolfvdhulst: I dislike this function, you should probably refactor it a bit.

I don't understand what you are doing line 81-83. It seems you're assuming these vectors have the same length but it's not super clear to me. Similarly, you're doing weird things in other places.

Secondly, by calling erase manually every time,you are performing a lot resizes (even though you do it back to front). You can just delete the entire range of elements.

Thirdly, use iterators! This makes it much more clear what you're iterating over. I suggest you look into .rend() and .rbegin() of std::vector<>. If you don't know what an iterator is, just let me know and I'll explain you.

src/utilities/Dealer.cpp
```diff
@@ -19,93 +23,126 @@ Dealer::DealerFlag::DealerFlag(DealerFlagTitle title, DealerFlagPriority priorit
...
+    std::unordered_map<std::string, v::RobotView> output;
+    std::vector<RoleScores> scores = getScoreMatrix(allRobots, flagMap, stpInfoMap);
+    // Make index of roles and ID to keep track which are the original indexes of each and get roleNames
+    std::vector<int> indexRoles;
+    std::vector<int> indexID;
+    std::vector<std::string> roleNames;
+    distribute_init(indexRoles,indexID,roleNames,flagMap);
+
+            // Loop through the order of role priorities (column)
+    for (int i = sizeof(DealerFlagPriority); i >= 0; i--){
```
rolfvdhulst: maybe add a comment to make clear every priority level can represent a group of robots.

rolfvdhulst (commented):
Most of the code looks fine, except for the distribute_remove function I commented on. I like the general code, I can tell you're improving :). Also; the .reserve() calls everywhere are nice. You really need to look into what iterators and reverse iterators are, though :)

src/utilities/Dealer.cpp
mmlievenseRTT: Is there something as to many structs? haha
I couldnt find a way (but i am sure there is) to get a incrementing vector constructed..

src/utilities/Dealer.cpp
mmlievenseRTT: In 81-83 i am removing the roles that have been given away already. I.e.:
Start: Currently:
scores (A represent the index of the role, a1 represent score for robot1 for role A, this is scores[A] = a1,a2,a3,a4)
[0] A -> a1 a2 a3 a4 | 2
[1] B -> b1 b2 b3 b4 | 0
[2] C -> c1 c2 c3 c4  | 2
[3] D -> d1 d2 d3 d4 | 1
Now you want to give away roles A and C as they have the highest priority, currentScores are:
[0] A -> a1 a2 a3 a4
[1] C -> c1 c2 c3 c4
C has [2] in the real matrix and therefore this needs to be remembered as it is linked to an ordered vector with roleNames. When the roles are given away (to robot 2 and 4 in this example) you get, assignments:
[0] 2
[1] 4
linking A to 2 and C to 4.
Finally you want to remove given away roles and robots so you want:
[0] B -> b1 b3 | 0
[1] D -> d1 d3 | 1
Now, if you deal robot 3, the index [1] will be incorrect and linking the assignment to the role will need to keep in mind the original ID index.

That's why there are so many vectors to keep in mind the index of the original and current role and id...
I agree, i did like to refactor this aswell, however i am not sure how to effectivily do that with 2D vectors. And not sure if the result will be more efficient (more indexes to go over with i.e. NaN elements for the removed roles and Id's).

I can solve it with pairs and triplets as you can save the info in there, but not sure if thats desired haha :3

mmlievenseRTT: Is remove better than erase? 
https://stackoverflow.com/questions/4115279/most-efficient-way-of-erasing-deleting-multiple-stdvector-elements-while-retai
I'll use the iterators there if it is

src/utilities/Dealer.cpp
mmlievenseRTT: I was going to make a good template for every play soon with all this information (bit like the front-end of STP), but i guess the more places the better

src/utilities/Dealer.cpp
rolfvdhulst: std::iota does exactly what you need (#include <numeric>). 

You could simply write:

indexRoles.resize(flagMap.size());
std::iota(indexRoles.begin(),indexRoles.end(),0);

But this is fine too (it's clear enough)

rolfvdhulst: Too many structs can be a thing, but typically encapsulation helps make your code more readable and easier to argue about.

src/utilities/Dealer.cpp
rolfvdhulst: I was mentioning is because I was a bit confused as first. Also I just realized that sizeof(DealerFlagPriority) is probably a bad idea to do.

rolfvdhulst: It may be better to have an explicit order or vector and loop over that (it's also more clear)

mmlievenseRTT: Dealer.h:
// Lowest to Highest Priority ordering. UNIQUE Classes are the Highest. (Order using in Dealer::distribute() )
enum class DealerFlagPriority { LOW_PRIORITY, MEDIUM_PRIORITY, HIGH_PRIORITY, REQUIRED, KEEPER };

This is the method I implemented the order, the size is allowed to vary aslong as the order is followed like in the comment. How would you suggest it otherwise?

src/utilities/Dealer.cpp
mmlievenseRTT: Look at it with Rolf.

Yuhanun (approved):
lgtm

PR #1214, opened 2021-03-26
STP < Dealer option to force ID

mmlievenseRTT:
### Summary
All the playmaker to define roles that need to go to a specific robot ID.

### Related issues
This seems like a option that a real AI would not want to have, however due to the delay of accurate information between the world and the AI (in particular positioning of the ball) a bypass can be added to increase efficiency of the AI.

**I.e. Passing;** Robot X passes to Robot Y and the play ends. In the next play the AI should decide whom is going to receive the ball, but doesnt know that the previous play this was already decided and the action was finalized. Instead of letting the AI estimate that the ball is heading towards Robot Y, this information is send to the next play and the receiver can be forced on Robot Y.

This should ofc be validated by the play and if the ball is not heading the correct direction, the play should fail and reconsider its options.

### DIFF
--- a/include/roboteam_ai/utilities/Dealer.h
+++ b/include/roboteam_ai/utilities/Dealer.h
@@ -22,7 +22,7 @@
 namespace rtt::ai {
 namespace v = rtt::world::view;
 
-// Set up a struct for dealerflags. Set up a struct for dealerflags.
+/// Set up a struct for dealerFlags. Set up a struct for dealerFlags.
 enum class DealerFlagTitle {
     CLOSE_TO_THEIR_GOAL,
     CLOSE_TO_OUR_GOAL,
@@ -36,7 +36,10 @@ enum class DealerFlagTitle {
     CLOSEST_TO_BALL
 };
 
+/// Generalization for Priority of a role
 enum class DealerFlagPriority { LOW_PRIORITY, MEDIUM_PRIORITY, HIGH_PRIORITY, REQUIRED, KEEPER };
+
+/// The order at which the Priority will be dealt (Keeper first, Low last).
 static std::vector<DealerFlagPriority> PriorityOrder { DealerFlagPriority::KEEPER,
                                                        DealerFlagPriority::REQUIRED,
                                                        DealerFlagPriority::HIGH_PRIORITY,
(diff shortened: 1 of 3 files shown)


include/roboteam_ai/utilities/Dealer.h
```diff
@@ -57,13 +57,14 @@ class Dealer {
...
         std::vector<DealerFlag> flags;
+        int forcedID = -1;
     };
 
     using FlagMap = std::map<std::string, RoleInfo>;
     Dealer(v::WorldDataView world, rtt::world::Field *field);
     virtual ~Dealer() = default;  // needed for test
     // Create a distribution of robots
-    std::unordered_map<std::string, v::RobotView> distribute(const std::vector<v::RobotView> &allRobots, const FlagMap &flagMap,
+    std::unordered_map<std::string, v::RobotView> distribute(const std::vector<v::RobotView> &_allRobots, FlagMap &flagMap,
```
Yuhanun: why are you modifying the FlagMap here?

include/roboteam_ai/utilities/Dealer.h
```diff
@@ -174,8 +175,17 @@ class Dealer {
...
-    void distribute_remove(DealerDistribute &current, std::vector<int> &indexRoles, std::vector<int> &indexID,
+    static void distribute_remove(DealerDistribute &current, std::vector<int> &indexRoles, std::vector<int> &indexID,
                            std::vector<RoleScores> &scores);
+
+    /**
+     * Distributes the forced roles first, so that other rest of the function does not need to compute extra information
+     * @param allRobots a copy of all our robots, where in the robots will be removed
+     * @param flagMap wherein the roles will be removed
+     * @param output
+     */
```
Yuhanun, line 206: Why are these 3 static now?

Yuhanun (commented):
Soem weird stuff with mutability.

include/roboteam_ai/utilities/Dealer.h
mmlievenseRTT: I was not sure how to do it nicely, however the thinking was that:
if you force a role on a Robot, the role and robot should not be considered anymore. By removing them, it would mean that the scoreMatric and everything after that does not consider the Role and Robot, which is more efficient.

The downside of such is that the flagMap will be shorter than in the Play. However the flag map is not used after the dealer anymore. Alternatively, i could make a copy of flagMap and change that, but as it was not used anymore, i chose the not make copies where it was not needed

include/roboteam_ai/utilities/Dealer.h
mmlievenseRTT: Oh think it was a Tidy pop-up, not required indeed

include/roboteam_ai/utilities/Dealer.h
Yuhanun: The thing is that the function is called "distribute", which is now modifying things, and has side effects. 

It's fine but change the name of the function to something that makes it clear it modifies.

mmlievenseRTT: I'd prefer then to just make a copy, removing the reference to the flagMap (so that in play has the original flagMap and the dealer can do with it as it wants)

rolfvdhulst:
Code seems fine now, but haven't really looked into the logic.

rolfvdhulst (approved):

PR #1215, opened 2021-03-26
STP < PlayInfo from previous play

mmlievenseRTT:
### Summary
Added a structure to where information from the previous play can be saved and passed on to the next play.

### Related issues
An example would be that if a ball is passed to a other robot, that the other robot should become the reciever in the net play.
The neat method of doing this is letting the AI decide to whom the ball is passed to, however that would require an instant update of the world with correct velocities of the ball (which is not the case at this point). Therefore passing the already known information between plays should work. This assumes that the previous play has succeeded and that the skills like shootatpos are somewhat accurate, if not the play should fail with.

### DIFF
--- a/include/roboteam_ai/stp/Play.hpp
+++ b/include/roboteam_ai/stp/Play.hpp
@@ -130,14 +130,21 @@ class Play {
      */
     uint8_t getLastScore();
 
-    /// Place to store info in that is needed between Plays
-    using PlayInfo = std::unordered_map<int, StpInfo>;
+    struct PlayInfo {
+
+    };
+
+    /**
+     * Place to store info in that is needed between Plays.
+     * E
+     */
+    using PlayInfos = std::unordered_map<int, StpInfo>;
 
     /**
      * Saves all necessary information (that is needed for a potential next Play), when this Play will be finished
      * @return List of all the necessary information
      */
-    PlayInfo finalizePlay();
+    PlayInfos storePlayInfo();
 
 protected:
     /**
--- a/include/roboteam_ai/stp/Play.hpp
+++ b/include/roboteam_ai/stp/Play.hpp
@@ -114,8 +114,8 @@ class Play {
     virtual const char* getName() = 0;
 
     /**
-
-     * Gets the
+     *
+     * Gets the current StpInfos from a Play
      */
     std::unordered_map<std::string, StpInfo> getStpInfos();
 
@@ -129,6 +129,12 @@ class Play {
      * @return score if no value -> 0
      */
     uint8_t getLastScore();
+
+    /**
+     * Saves all necessary information (that is needed for a potential next Play), when this Play will be finished
+     * @return List of all the necessary information
+     */
+    std::unordered_map<int, StpInfo> finalizePlay();
 
 protected:
     /**
--- a/include/roboteam_ai/stp/Play.hpp
+++ b/include/roboteam_ai/stp/Play.hpp
@@ -24,13 +24,21 @@ class Play {
    public:
 
     //TODO move to gen cpp
+    /**
+     * Generalized Keys for passing information form the old play to the new.
+     * Usage in the storePlayInfo where KeyInfo is the key for the elements in the map.
+     */
     enum class KeyInfo{
-        isPasser = 0,
-        isReceiver,
-        isShooter,
-        hasBall
+        isPasser = 0,   // Robot that passes the ball last play
+        isReceiver,     // Robot that should receive the ball (as passer shot to there)
+        isShooter,      // Robot that Shot the ball last play
+        hasBall         // Robot that had the ball last play
     };
 
+    /**
+     * Generalized information structure for the map of storePlayInfo.
+     * Allows for saving specific information from the old play to the new.
+     */
     struct StoreInfo {
         std::optional<int> robotID;
         std::optional<Vector2> robotPosition;
@@ -41,7 +49,7 @@ class Play {
     };
 
     /**
-     * Place to store info in that is needed between Plays.
+     * Place to store info in that is needed between Plays. Used in storePlayInfo.
      */
     using PlayInfos = std::unordered_map<KeyInfo, StoreInfo>;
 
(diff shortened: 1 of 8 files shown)


include/roboteam_ai/stp/Play.hpp
```diff
@@ -22,6 +22,37 @@ namespace rtt::ai::stp {
...
+
+    struct StoreInfo {
+        std::optional<int> robotID;
+        std::optional<Vector2> robotPosition;
+        std::optional<Vector2> moveToPosition;
+        std::optional<Vector2> defendPosition;
+        std::optional<Vector2> shootAtPosition;
+        std::optional<Vector2> passToRobot;
+    };
+
```
Yuhanun: Docs

src/stp/Play.cpp
```diff
@@ -117,24 +118,23 @@ std::unordered_map<Role*, Status> const& Play::getRoleStatuses() const { return
 
 bool Play::isValidPlayToKeep(PlayEvaluator& playEvaluator) noexcept {
     if (!interface::MainControlsWidget::ignoreInvariants) {
-        return std::all_of(keepPlayEvaluation.begin(), keepPlayEvaluation.end(), [playEvaluator] (auto& x) { return playEvaluator.checkEvaluation(x); });
+        return std::all_of(keepPlayEvaluation.begin(), keepPlayEvaluation.end(), [&playEvaluator] (auto& x) { return playEvaluator.checkEvaluation(x); });
```
Yuhanun: Consider formatting this:

```
return std::all_of(begin(), end(), [data](arg) {
    return ...;
});
```

src/stp/Play.cpp
```diff
@@ -117,24 +118,23 @@ std::unordered_map<Role*, Status> const& Play::getRoleStatuses() const { return
...
     } else {
         return true;
     }
 }
 
 bool Play::isValidPlayToStart(PlayEvaluator& playEvaluator) const noexcept {
     if (!interface::MainControlsWidget::ignoreInvariants) {
-        return std::all_of(startPlayEvaluation.begin(), startPlayEvaluation.end(), [playEvaluator] (auto& x) { return playEvaluator.checkEvaluation(x); });
+        return std::all_of(startPlayEvaluation.begin(), startPlayEvaluation.end(), [&playEvaluator] (auto& x) { return playEvaluator.checkEvaluation(x); });
```
Yuhanun: same here

src/stp/Play.cpp
```diff
@@ -117,24 +118,23 @@ std::unordered_map<Role*, Status> const& Play::getRoleStatuses() const { return
...
     if (!interface::MainControlsWidget::ignoreInvariants) {
-        return std::all_of(keepPlayEvaluation.begin(), keepPlayEvaluation.end(), [playEvaluator] (auto& x) { return playEvaluator.checkEvaluation(x); });
+        return std::all_of(keepPlayEvaluation.begin(), keepPlayEvaluation.end(), [&playEvaluator] (auto& x) { return playEvaluator.checkEvaluation(x); });
     } else {
         return true;
     }
 }
 
 bool Play::isValidPlayToStart(PlayEvaluator& playEvaluator) const noexcept {
     if (!interface::MainControlsWidget::ignoreInvariants) {
```
Yuhanun: Also this condition is uhh

ignoreInvariants || all_of(...);

include/roboteam_ai/stp/Play.hpp
mmlievenseRTT: wops, will do

src/stp/Play.cpp
mmlievenseRTT: I only see a change in an enter in the lambda, is that correct?

Yuhanun: Correct. it's kinda gross on 1 line.

src/stp/Play.cpp
mmlievenseRTT: lol, old code. ill make it more compact

mmlievenseRTT: Think a refactor of the ApplicationManager should be done soon, its a real mess

src/stp/Play.cpp
```diff
@@ -3,12 +3,13 @@
...
 #include "stp/Play.hpp"
-
 #include "interface/widgets/MainControlsWidget.h"
 
 namespace rtt::ai::stp {
 
-void Play::initialize(std::unordered_map<std::basic_string<char>, StpInfo>) noexcept {
+void Play::initialize(PlayInfos& _previousPlayInfos) noexcept {
+    previousPlayInfos = _previousPlayInfos;
+    if (!previousPlayInfos->empty()) RTT_DEBUG(std::to_string(previousPlayInfos->begin()->second.robotID.value_or(-1)));
```
rolfvdhulst: I personally don't like 1-line if statements with curly braces, it's unclear what belongs to the scope. Especially since you call a macro which gets expanded, this can be bug-sensitive.

src/stp/Play.cpp
```diff
@@ -116,25 +117,20 @@ void Play::distributeRoles() noexcept {
...
 
 bool Play::isValidPlayToStart(PlayEvaluator& playEvaluator) const noexcept {
-    if (!interface::MainControlsWidget::ignoreInvariants) {
-        return std::all_of(startPlayEvaluation.begin(), startPlayEvaluation.end(), [playEvaluator] (auto& x) { return playEvaluator.checkEvaluation(x); });
-    } else {
-        return true;
-    }
+    return (interface::MainControlsWidget::ignoreInvariants || std::all_of(startPlayEvaluation.begin(), startPlayEvaluation.end(), [&playEvaluator] (auto& x) {
+        return playEvaluator.checkEvaluation(x);
+    }));
```
rolfvdhulst, line 129: probably separate the second half behind the or (||) on a new line.

rolfvdhulst (approved):
Two minor comments which are very quick fixes/not crucial, rest looks good.

src/stp/Play.cpp
Yuhanun:
```suggestion
    if (!previousPlayInfos->empty()) 
        RTT_DEBUG(std::to_string(previousPlayInfos->begin()->second.robotID.value_or(-1)));
```

PR #1219, opened 2021-04-23
Add orbit skill

mamiksik:
An attempt to add orbiting skill. Not complete -> problem with drifting while orbiting.

The general idea for the skill is to move in the direction of the normal vector to the directional vector from ball to robot.

Acceleration is controlled by sinus of angle be desired with a double period. -> Full angle = Maximal acceleration. 0 angle = no acceleration.

### DIFF
--- a/src/stp/skills/Orbit.cpp
+++ b/src/stp/skills/Orbit.cpp
@@ -1,0 +1,54 @@
+//
+// Created by mamiksik on 22-04-21.
+//
+
+#include "stp/skills/Orbit.h"
+
+#include <ApplicationManager.h>
+
+#include "control/ControlUtils.h"
+
+namespace rtt::ai::stp::skill {
+
+Status Orbit::onUpdate(const StpInfo &info) noexcept {
+    Vector2 directionVector = (info.getBall()->get()->getPos() - info.getRobot()->get()->getPos());
+    Angle normalAngle = directionVector.rotate(M_PI).rotate(M_PI_2).toAngle();
+
+    Vector2 initialVelocity = info.getRobot()->get()->getVel();
+    double initX = cos(initialVelocity.toAngle()) * initialVelocity.length();
+    double initY = sin(initialVelocity.toAngle()) * initialVelocity.length();
+
+    double margin = stp::control_constants::ROBOT_RADIUS + 2 * stp::control_constants::BALL_RADIUS;
+    Vector2 adjustDistance = directionVector.normalize().scale(info.getRobot()->get()->getDistanceToBall() - margin);
+
+    double multiplier = -sin(0.5 * (directionVector - info.getAngle()).toAngle());
+
+    double x = cos(normalAngle) * multiplier + cos(adjustDistance.toAngle()) * adjustDistance.length();  //- initX adjustment for initial velocity?;
+    double y = sin(normalAngle) * multiplier + sin(adjustDistance.toAngle()) * adjustDistance.length();  //- initY adjustment for initial velocity?;
+
+    double errorMargin = stp::control_constants::GO_TO_POS_ANGLE_ERROR_MARGIN * M_PI;
+    command.set_w(info.getAngle());
+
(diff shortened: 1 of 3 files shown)


Yuhanun (approved):
lgtm

src/stp/skills/Orbit.cpp
```diff
@@ -0,0 +1,54 @@
...
+    auto initX = cos(initialVelocity.toAngle()) * initialVelocity.length();
+    auto initY = sin(initialVelocity.toAngle()) * initialVelocity.length();
+
+    auto margin = stp::control_constants::ROBOT_RADIUS + 2 * stp::control_constants::BALL_RADIUS;
+    auto adjustDistance = directionVector.normalize().scale(info.getRobot()->get()->getDistanceToBall() - margin);
+
+    auto multiplier = -sin(0.5 * (directionVector - info.getAngle()).toAngle());
+
+    auto x = cos(normalAngle) * multiplier + cos(adjustDistance.toAngle()) * adjustDistance.length(); //- initX adjustment for initial velocity?;
+    auto y = sin(normalAngle) * multiplier + sin(adjustDistance.toAngle()) * adjustDistance.length(); //- initY adjustment for initial velocity?;
```
rolfvdhulst: It's fine for code usually, but I would typically discourage 'overuse' of auto like this. These data types should be either double or Vector2; naming them explicitly helps a lot with readability IMO.

rolfvdhulst: For really complicated types, it's fine to use auto, but here, I don't think it's necessary.

mamiksik: I will add explicit types.

Yuhanun: I don't necessarily agree with rolf here, though.

I think any type that's not a built-in type (aka arithmetic + ref/pointer qualified thereof), should be `auto`.

so yeah these variables should be explicitly typed, but if you get to std::vector etc i'd already auto them.

rolfvdhulst: That's a style preference. Personally, I like auto only when it's sort-of clear from the context what the type is. For templated containers this often is the case with good naming and auto is fine, but I would personally still use the type for simple classes. When using long, chained calls, like happens here often, I don't want to spend 2 minutes trying to decipher what the last call is and what type it gives, so I'd be tempted to use auto a bit less often. For simple single function calls, auto becomes a lot more acceptable.

mamiksik: ehh, don't we have a set code style?

rolfvdhulst: No, impossible to enforce when the team changes every year. But for now you can do what you like, as long as you at least do not use auto for built-in types.

mamiksik: I added types and amended the original commit. What about using a linter such as https://github.com/cpplint/cpplint?

rolfvdhulst: I believe we already had a linter, but we never use/enforce it anymore these days :/

mamiksik: I guess adding it to CI, for next year, would be a nice way to keep code consistent. btw: can we close/merge the PR if everything is fine?

Yuhanun: We've tried adding it to CI but it's not like, perfect.

If we can get CI to format any file that changed then sure, feel free to set it up :)

mamiksik: I think linter should not format code for you.  Also, not all linting rules are formating, aren't they? For example, PEP8 has a code style rule for if-else return. 

This code does not pass PEP8:
```python3
if x is Null:
    return "Fail"
else:
    return "Success"
``` 

This does:
```python
if x is Null:
    return "Fail"

return "Success"
```


=> I think CI should just disallow you to merge PR while linting check does not pass.

JaroKuiken (approved):
Ill merge

PR #1230, opened 2021-05-22
Refactore free kick them

mamiksik:
What about method shouldRoleSkipEndTactic()? Is there a substitute for it?

### DIFF
--- a/src/stp/plays/referee_specific/FreeKickThem.cpp
+++ b/src/stp/plays/referee_specific/FreeKickThem.cpp
@@ -2,31 +2,74 @@
 // Created by jordi on 07-05-20.
 //
 
-#include "stp/plays/referee_specific/FreeKickThem.h"
-#include "stp/invariants/game_states/FreeKickThemGameStateEvaluation.h"
+#include "include/roboteam_ai/stp/plays/referee_specific/FreeKickThem.h"
 #include "include/roboteam_ai/stp/roles/passive/Defender.h"
+#include "include/roboteam_ai/stp/computations/PositionComputations.h"
+#include "include/roboteam_ai/stp/roles/Keeper.h"
 #include "include/roboteam_ai/stp/roles/passive/Formation.h"
-#include "stp/roles/Keeper.h"
 
 namespace rtt::ai::stp::play {
+const char* FreeKickThem::getName() { return "Free Kick Them"; }
 
 FreeKickThem::FreeKickThem() : Play() {
-    startPlayInvariants.clear();
-    startPlayInvariants.emplace_back(std::make_unique<invariant::FreeKickThemGameStateInvariant>());
+    /// Evaluations that have to be true to be considered when changing plays.
+    startPlayEvaluation.clear();  // DONT TOUCH.
+    startPlayEvaluation.emplace_back(eval::FreeKickThemGameState);
 
-    keepPlayInvariants.clear();
-    keepPlayInvariants.emplace_back(std::make_unique<invariant::FreeKickThemGameStateInvariant>());
+    /// Evaluations that have to be true to allow the play to continue, otherwise the play will change. Plays can also end using the shouldEndPlay().
+    keepPlayEvaluation.clear();  // DONT TOUCH.
+    keepPlayEvaluation.emplace_back(eval::FreeKickThemGameState);
 
+    /// Role creation, the names should be unique. The names are used in the stpInfos-map.
     roles = std::array<std::unique_ptr<Role>, rtt::ai::Constants::ROBOT_COUNT()>{
-        std::make_unique<role::Keeper>("keeper"),       std::make_unique<role::Defender>("defender_0"), std::make_unique<role::Defender>("defender_1"),
-        std::make_unique<role::Defender>("defender_2"), std::make_unique<role::Defender>("blocker_0"),  std::make_unique<role::Defender>("blocker_1"),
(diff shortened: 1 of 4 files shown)


src/stp/plays/referee_specific/FreeKickThem.cpp
```diff
@@ -2,31 +2,76 @@
...
+Dealer::FlagMap FreeKickThem::decideRoleFlags() const noexcept {
+    Dealer::FlagMap flagMap;  // DONT TOUCH.
+
+    /// Flags that have a factor and a weight linked to it, can be given to a role
+    Dealer::DealerFlag closeToOurGoalFlag(DealerFlagTitle::CLOSE_TO_OUR_GOAL, DealerFlagPriority::HIGH_PRIORITY);
+    Dealer::DealerFlag closeToBallFlag(DealerFlagTitle::CLOSE_TO_BALL, DealerFlagPriority::HIGH_PRIORITY);
+
+    /// Creation flagMap. Linking roles to role-priority and the above created flags, can also force ID {roleName, {priority, flags, forceID}}
+    flagMap.insert({"keeper", {DealerFlagPriority::KEEPER, {}}});
+    flagMap.insert({"defender_0", {DealerFlagPriority::HIGH_PRIORITY, {closeToOurGoalFlag}}});
```
JaroKuiken: The DealerFlagPriority's are supposed to be used to give the Dealer an indication of which Roles are most important to a Play. By using them the Dealer can decide the best division of Roles over robots, but if you give each role 'HIGH_PRIORITY' the Dealer won't divide the Roles in an efficient way. My suggestion: Look at the different Roles in this Play, decide which ones are absolutely necessary and give these the 'REQUIRED' priority. Then look at which Roles are most important after that, and give those the 'HIGH_PRIORITY'. Then after that the Roles which are 'MEDIUM_PRIORITY', etc.

src/stp/plays/referee_specific/FreeKickThem.cpp
```diff
@@ -2,31 +2,76 @@
...
+    flagMap.insert({"blocker_3", {DealerFlagPriority::HIGH_PRIORITY, {closeToBallFlag}}});
+    flagMap.insert({"blocker_4", {DealerFlagPriority::HIGH_PRIORITY, {closeToBallFlag}}});
+    flagMap.insert({"blocker_5", {DealerFlagPriority::HIGH_PRIORITY, {}}});
+    flagMap.insert({"offender", {DealerFlagPriority::HIGH_PRIORITY, {closeToBallFlag}}});
+
+    return flagMap;  // DONT TOUCH.
+}
+
+uint8_t FreeKickThem::score(PlayEvaluator& playEvaluator) noexcept {
+    calculateInfoForScoredRoles(playEvaluator.getWorld());  /// DISABLE IF NOT USED
```
JaroKuiken: This function is not used and as an empty function in your header, so you can disable/comment it out

src/stp/plays/referee_specific/FreeKickThem.cpp
```diff
@@ -2,31 +2,76 @@
...
+    /// List of all factors that combined results in an evaluation how good the play is.
+    scoring = {{playEvaluator.getGlobalEvaluation(eval::FreeKickThemGameState), 1.0}};
+    return (lastScore = playEvaluator.calculateScore(scoring)).value();  // DONT TOUCH.
+}
 
 void FreeKickThem::calculateInfoForRoles() noexcept {
+    /// Function where are roles get their information, make sure not to compute roles twice.
+    calculateInfoForScoredRoles(world);  // DONT TOUCH.
```
JaroKuiken: Same here, can be taken out

src/stp/plays/referee_specific/FreeKickThem.cpp
```diff
@@ -2,31 +2,76 @@
...
-#include "stp/invariants/game_states/FreeKickThemGameStateEvaluation.h"
+#include "include/roboteam_ai/stp/plays/referee_specific/FreeKickThem.h"
 #include "include/roboteam_ai/stp/roles/passive/Defender.h"
+#include "include/roboteam_ai/stp/computations/PositionComputations.h"
+#include "include/roboteam_ai/stp/roles/Keeper.h"
 #include "include/roboteam_ai/stp/roles/passive/Formation.h"
-#include "stp/roles/Keeper.h"
 
 namespace rtt::ai::stp::play {
+const char* FreeKickThem::getName() { return "FreeKickThem Template"; }
```
JaroKuiken: Why change the name? No need to change the previous getName() at the bottom of the class

src/stp/plays/referee_specific/FreeKickThem.cpp
```diff
@@ -96,32 +144,8 @@ void FreeKickThem::calculateInfoForDefenders() noexcept {
...
-    flagMap.insert({"blocker_4", {closeToBallFlag}});
-    flagMap.insert({"blocker_5", {closeToBallFlag}});
-    flagMap.insert({"offender", {closeToBallFlag}});
-
-    return flagMap;
+void FreeKickThem::calculateInfoForOffenders() noexcept {
+    stpInfos["offender"].setPositionToMoveTo(Vector2(field.getFieldLength() / 4, 0.0));
 }
 
-const char *FreeKickThem::getName() { return "Free Kick Them"; }
```
JaroKuiken, line 125: You can keep this in, no need to change the name!

JaroKuiken (changes requested):
DealerFlagPriority's and some small things

JaroKuiken (approved):

PR #1243, opened 2021-07-27
MacOS support and build improvements

daw1012345:
This was supposed to be just a quick job of adding MacOS support, but I noticed a couple of things that could be improved with the build system.

- Refactor to use modern CMake
- Work around Qt5 issues on MacOS
- Add instructions on how to build on MacOS
- Add missing and remove redundant dependencies in README
- Add CCache install instructions to README
- Set CMake to use unity builds
- Remove unused CMake files (cotire was included but not actually used)
  - Use modern CMake features to speed up build times
- Recommend and set default build system to Ninja
- Use newer method of fetching/discovering dependencies
   - Remove unnecessary third party git submodules 
- Change structure of some roboteam repositories (e.g. where full CMake projects were nested in the `include/` directory)

Only meant to be merged together with corresponding PRs in other repos.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,8 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.16)
 project(roboteam_ai)
 
 add_subdirectory(src)
 
-include(LocateQt5)
 set(CMAKE_AUTORCC OFF)
 set(CMAKE_AUTOUIC OFF)
 set(CMAKE_INCLUDE_CURRENT_DIR OFF) #Find includes in corresponding build directories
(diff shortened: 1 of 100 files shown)


CMakeLists.txt
```diff
@@ -304,13 +301,26 @@ set(AI_INCLUDES
         PRIVATE include/roboteam_ai
         )
 
+# Apple-specific fixes and separate out old interface for performance
+if(APPLE)
+    # AUTOMOC is borked on CMAKE 3.21.0 + moc 5.15.2 + libprotoc 3.17.3
```
Yuhanun, line 306: "borked" nice xd

Yuhanun (approved):

PR #1244, opened 2021-08-30
Fix preexisting tests

daw1012345:
Reenable tests and tweak to work with newer codebase


CMakeLists.txt
```diff
@@ -10,17 +10,8 @@ set(CMAKE_INCLUDE_CURRENT_DIR OFF) #Find includes in corresponding build directo
...
-            /usr/local/lib/libgtest.a
-            /usr/local/lib/libgtest_main.a
-            /usr/local/lib/libgmock.a
-            /usr/local/lib/libgmock_main.a
-            )
 else (NOT APPLE)
     set(CMAKE_AUTOMOC ON)
-    set(GTEST_LIB
-            PRIVATE gtest
-            PRIVATE gmock)
```
rolfvdhulst: Unrelated changes? and we'll need it?

CMakeLists.txt
```diff
@@ -315,17 +305,17 @@ target_link_libraries(ai_iface ${AI_LIBS})
 target_include_directories(ai_iface ${AI_INCLUDES})
 
 add_library(ai_lib ${AI_SOURCES})
-target_link_libraries(ai_lib ${AI_LIBS})
+target_link_libraries(ai_lib ${AI_LIBS} GTest::gtest GTest::gtest_main)
```
rolfvdhulst: Why in ai_lib? Just add it to AI_LIBS?

test/main.cpp
```diff
@@ -1,15 +0,0 @@
...
-
-#define RUNNING_TEST
-
-int main(int argc, char **argv) {
-    testing::InitGoogleTest(&argc, argv);
-    QApplication app(argc, argv);  // initialize qt5
-    rtt::ai::Constants::init();
-    rtt::SETTINGS.init(0);
-
-    return RUN_ALL_TESTS();
```
rolfvdhulst: Why delete this? Now we cannot run all tests easily.

rolfvdhulst (dismissed):
Try not to leave commented code everywhere, and there seem to be some unrelated changes.

CMakeLists.txt
```diff
@@ -10,17 +10,8 @@ set(CMAKE_INCLUDE_CURRENT_DIR OFF) #Find includes in corresponding build directo
...
-            /usr/local/lib/libgtest.a
-            /usr/local/lib/libgtest_main.a
-            /usr/local/lib/libgmock.a
-            /usr/local/lib/libgmock_main.a
-            )
 else (NOT APPLE)
     set(CMAKE_AUTOMOC ON)
-    set(GTEST_LIB
-            PUBLIC gtest
-            PUBLIC gmock)
```
alexander-de-ranitz, line 23: unrelated changes?

alexander-de-ranitz:
I currently get a CMake error on this branch (see below). I think adding find_package(GTest) should fix it though.
 
```
CMake Error at roboteam_ai/CMakeLists.txt:316 (add_executable):
  Target "roboteam_ai_tests" links to target "GTest::gtest" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

daw1012345:
> I currently get a CMake error on this branch (see below). I think adding find_package(GTest) should fix it though.

This PR is dependent on the PR in roboteam_suite. Have you checked out that branch?

alexander-de-ranitz:
> 

Yeah that was it, MB :)

However, I do currently have two more issues:
1) for some magical reason, my CLion does not recognize the roboteam_ai_test as a GTest executable. However, it does work properly when it is named roboteam_ai_tests (or anything else pretty much). No clue why this happens, but since the other tests are also called something_tests, maybe we can change this from roboteam_ai_test to roboteam_ai_tests too.
2) I get the following linking error: ```undefined reference to `rtt::ai::control::ControlModule::addRobotCommand(std::optional<rtt::world::view::RobotView>, proto::RobotCommand const&, rtt::world::World const*)'```
This is fixed for me by moving roboteam_ai_control to the bottom of the target_link_libraries list.

alexander-de-ranitz (approved):
Looks good! The tests run just fine, almost all of them pass :)

For some reason, my CLion still does not recognise the GTest properly, but that seems like an issue only I have, and I can just run it from terminal instead.

PR #1250, opened 2021-10-11
Add missing include <optional>

Florisheinen1:
### Summary
A few files use the standard library <optional> but do not include it. Some compilers allow for it, as other classes do import it into the namespace, but more strict compilers will deny a build. I added the include statements in these files in this branch.


itsasimpleplan (approved):
If something is used in a class, then it should be included in the head of the file. It should now also work for Floris/

PR #1252, opened 2021-10-25
Fix logic in isValidPlayToKeep to return false when shouldEndPlay is true

alexander-de-ranitz:
There were two issues with this function:
1) shouldEndPlay was not negated (!), which meant that is shouldEndPlay was true, isValidPlayToKeep was also true, which is the opposite of the desired behaviour. Fixed by adding !.
2) isValidPlayToKeep returned true if shouldEndPlay OR all of the playEvaluations were true. However, we need to make sure that BOTH conditions are fullfilled. Hence, the OR was changed into AND.

Without these changes, this function would return the wrong value if shouldEndPlay() is true, as the function would return true when it should be false, as the play should end. Furthermore, this function used OR where AND should be used, as both shouldEndPlay needs to be false AND all evaluations must be true in order to be a valid play to keep.

- [ X] The tests are passing

### Summary
Negated shouldEndPlay and changed shouldEndPlay OR [all evaluations are true] to  ... AND ...

**functionality**  
- Is the code is properly tested?
- Is the code working as expected
- Is the code is understandable and easy to read
- Are proper abstractions made? (no duplicate code, nice use of functions/objects)

**style**  
- Are smart pointers used?
- unused variables?
- magic numbers?
- well documented?
- No unused imports?
- Proper indentation
- Proper use of namespaces

### DIFF
--- a/src/stp/Play.cpp
+++ b/src/stp/Play.cpp
@@ -114,10 +114,10 @@ namespace rtt::ai::stp {
     std::unordered_map<Role *, Status> const &Play::getRoleStatuses() const { return roleStatuses; }
 
     bool Play::isValidPlayToKeep(PlayEvaluator &playEvaluator) noexcept {
-        return (interface::MainControlsWidget::ignoreInvariants || shouldEndPlay() ||
+        return (interface::MainControlsWidget::ignoreInvariants || (!shouldEndPlay() &&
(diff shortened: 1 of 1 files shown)


itsasimpleplan (approved):
Seems like a valid change. This way the problems with 'ignore invariants' in the interface are hopfully less.

PR #1254, opened 2021-10-28
Do not touch ball before angle of robot is correct

alexander-de-ranitz:
If the ball is close to the robot but the robot is not facing towards the ball, it can happen that the robot hits the ball with its side in getBall, causing the ball to be pushed away. To avoid this, we prevent the robot from getting too close to the ball while its angle is incorrect. As long as the angle is too far off, we set the robot to go to its own position, which makes the tactic go to the next skill: rotate (since goToPos is immediately successful). The robot then rotates until it is within a margin of HAS_BALL_ANGLE_ERROR_MARGIN * M_PI, when the tactic resets and the robot is told to go to the position of the ball. This ensures that the robot always faces the right direction before actually making contact with the ball.

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages

### DIFF
--- a/src/stp/tactics/active/GetBall.cpp
+++ b/src/stp/tactics/active/GetBall.cpp
@@ -20,31 +20,37 @@ std::optional<StpInfo> GetBall::calculateInfoForSkill(StpInfo const &info) noexc
 
     if (!skillStpInfo.getRobot() || !skillStpInfo.getBall()) return std::nullopt;
 
-    Vector2 robotPosition = info.getRobot().value()->getPos();
-    Vector2 ballPosition = info.getBall().value()->getPos();
+    Vector2 robotPosition = skillStpInfo.getRobot().value()->getPos();
+    Vector2 ballPosition = skillStpInfo.getBall().value()->getPos();
+    double ballDistance = (ballPosition - robotPosition).length();
 
     // If this robot is not the keeper, don't get the ball inside a defense area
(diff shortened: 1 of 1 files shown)


Florisheinen1 (approved):
This code compiles, and looks like it makes sense. I would not be able to see if this breaks other parts of the code, so @itsasimpleplan will have to check it too.

src/stp/tactics/active/GetBall.cpp
```diff
@@ -20,31 +20,37 @@ std::optional<StpInfo> GetBall::calculateInfoForSkill(StpInfo const &info) noexc
...
 
-    skillStpInfo.setPositionToMoveTo(newRobotPosition);
-
     return skillStpInfo;
 }
 
 bool GetBall::isTacticFailing(const StpInfo &info) noexcept { return false; }
 
-bool GetBall::shouldTacticReset(const StpInfo &info) noexcept { return !info.getRobot()->hasBall(); }
+bool GetBall::shouldTacticReset(const StpInfo &info) noexcept { return info.getRobot()->get()->getAngleDiffToBall() < control_constants::HAS_BALL_ANGLE_ERROR_MARGIN * M_PI; }
```
itsasimpleplan, line 53: Example of info being used

itsasimpleplan (changes requested):
The code itself looks good, however it does not seem consistent. 
Sometimes skillStpInfo is used, sometimes just info, while they seem to be the same object. 
Maybe change it for this file or atleast the parts where you worked with it to make it more consistent

PR #1267, opened 2021-11-29
Fix/keeper

alexander-de-ranitz:
The keeper id, stored in the interfaceGameState, was not updated when the keeper was distributed. This meant that whenever the robot with the default keeper id (0) was not in the field, we would now know who our keeper was in STP. This is problematic for path planning. This PR makes sure that we save the keeper id if this role is distributed. The default keeper id is also set to -1, which forces the dealer to pick a keeper based on cost, as this is the desired behavior in pretty much every scenario. If we want to force a certain robot to be keeper, the default keeper id can be set to that id.

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages


Florisheinen1 (approved):
lgtm

PR #1281, opened 2021-12-21
Changed MAX_CHIP_POWER from 8 to 6.5

itsasimpleplan:
The MAX_CHIP_POWER should be the same as the MAX_KICK_POWER, which should both be 6.5. This is namely the upper boundry that control uses for this value. Setting it to 8 wouldn't make sense.

### Comments for the reviewer
- 

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages


emielsteerneman:
Definitely needed, since [REM expects](https://github.com/RoboTeamTwente/roboteam_embedded_messages/blob/fc1b7a5c53cfa281180c134b8bc3cdb28a8ec8f5/generator/packets.py#L18) a power between [0, 6.5]. It doesn't support 8

alexander-de-ranitz (approved):

PR #1313, opened 2022-01-20
Fix segfault and small issue in defendShot/Pass

alexander-de-ranitz:
World and field are only set in the play after it has been selected. When scoring each play, this has not been done yet. Therefore, the internal world and field of each play should not be accessed in the score function, as this could cause a segfault if these are not set yet. To avoid this, we use the world as set in the playEvaluator, which is updated before scoring.

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages


GagataS (approved):
LGTM

PR #1315, opened 2022-01-22
Stp info collision settings

itsasimpleplan:
### Comments for the reviewer
The struct 'AvoidObjects' in StpInfoEnums consists of 3 boolean variables. Previously there was only one function that could set all of these at the same time and one that could get them all at the same time. This is not ideal if you only want to set/get only one boolean. 
There have been added setters and getters for each of the boolean variables in the struct. This way it is possible to only set one of the variables to true/false. This makes it easy to for example say that the keeper should not avoid the defend area, but do the rest as the other robots. 
The setter and getter for the complete struct is still there, since this is used in the path planning algorithm. 

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages

### DIFF
--- a/include/roboteam_ai/stp/StpInfo.h
+++ b/include/roboteam_ai/stp/StpInfo.h
@@ -81,8 +81,16 @@ struct StpInfo {
     std::string getRoleName() const { return roleName; }
     void setRoleName(std::string name) { roleName = name; }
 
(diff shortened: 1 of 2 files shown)


alexander-de-ranitz (approved):
Looks good

PR #1316, opened 2022-01-22
Fix/dealer

emielsteerneman:
### Comments for the reviewer
Fixed the bug where the dealer threw an assert error when there were more robots than roles. Added  robot.getPositionToDefend() to Dealer::getRobotScoreForDistance, although it's unsure yet if this is an improvement. 

###### Code Quality

###### Testing
- [x] All tests are passing.

I compared the role assignment to the dealer before this fix, and it's the same. I added print statements to manually check the cost matrix.

###### Commit Messages

### DIFF
--- a/include/roboteam_ai/utilities/Dealer.h
+++ b/include/roboteam_ai/utilities/Dealer.h
@@ -70,14 +70,14 @@ class Dealer {
     virtual ~Dealer() = default;  // needed for test
 
     /**
-     * Distributes the role between the friendly robots considering all information given in the role_map
-     * !!! robots and role_map are copies as they are modified to reduce computations after the forcedID's.
+     * Distributes the role between the friendly robots considering all information given in the role_to_flags
+     * !!! robots and role_to_flags are copies as they are modified to reduce computations after the forcedID's.
      * @param robots vector containing all friendly robots
-     * @param role_map information map with all information for each role (with key -> roleName)
+     * @param role_to_flags information map with all information for each role (with key -> roleName)
      * @param stpInfoMap
      * @return a vector with the roleName and the Robot that should get the role
      */
-    std::unordered_map<std::string, v::RobotView> distribute(std::vector<v::RobotView> robots, FlagMap role_map, const std::unordered_map<std::string, stp::StpInfo> &stpInfoMap);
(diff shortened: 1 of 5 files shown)


emielsteerneman:
Fixed a ~segfault~ bad allocation exception in [distribute_forcedIDs](https://github.com/RoboTeamTwente/roboteam_ai/blob/818879115147ea7fdbdf3aa5c39e4c9910370f51/src/utilities/Dealer.cpp#L160). 

The function incorrectly assumed that all robots in the robot map were sorted by id (which is not guaranteed) and in the  range [0 .. n_robots] (which is incorrect if a robot is swapped out).

include/roboteam_ai/utilities/Dealer.h
```diff
@@ -69,14 +70,14 @@ class Dealer {
...
-     * @param flagMap information map with all information for each role (with key -> roleName)
+     * Distributes the role between the friendly robots considering all information given in the role_map
+     * !!! robots and role_map are copies as they are modified to reduce computations after the forcedID's.
+     * @param robots vector containing all friendly robots
+     * @param role_map information map with all information for each role (with key -> roleName)
      * @param stpInfoMap
      * @return a vector with the roleName and the Robot that should get the role
      */
-    std::unordered_map<std::string, v::RobotView> distribute(std::vector<v::RobotView> allRobots, FlagMap flagMap, const std::unordered_map<std::string, stp::StpInfo> &stpInfoMap);
+    std::unordered_map<std::string, v::RobotView> distribute(std::vector<v::RobotView> robots, FlagMap role_map, const std::unordered_map<std::string, stp::StpInfo> &stpInfoMap);
```
Yuhanun: Keep in mind that a robot view is still a reference to the original robot

emielsteerneman: I'm not sure what you're saying. Do you mean the segfault danger is still there when the original robots disappear?

Yuhanun: Yeah exactly, std::vector<robot> would be better, you can always construct a view from that

Yuhanun:
LGTM

Yuhanun (approved):
LGTM

PR #1324, opened 2022-01-28
Pass during kickoff

alexander-de-ranitz:
### Summary

This PR makes sure that we do something resembling a pass during kickoff. The main thing to note is that this passing position is still hard-coded; in the future, it would be nice to have this more dynamic and take into account where the enemies are.

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages


TijmenWesteneng (approved):
Ugly, but better than the previous option, so let's merge before spain.

PR #1332, opened 2022-02-16
Add position scoring test

alexander-de-ranitz:
### Summary
This program can generate points spread across a randomly generated field, to then score these points. These results can then be saved to txt files for further analysis. 

This can be useful for visualizing scoring, but also comparing the best score found by this program (which should be close to the optimum when using a lot of points) to that found by our AI (which has to be fast).

###### Code Quality

###### Testing
- [x] All tests are passing.

###### Commit Messages


Florisheinen1 (approved):
LGTM.
Normally, I would require the checking of whether a proto object even has a set variable, like the ball inside world. Using that information without it being set will result in a segfault, but seeing this is a test, and the start conditions are set, including the ball, it is not necessary here.

PR #1333, opened 2022-02-23
replaced get ball position with get expected ball position in getRobo…

maxthielen:
Simply making use of the already built function considering ball velocity when returning the location of the ball.


alexander-de-ranitz:
Perhaps it is better to add a new function to see which robot is closest to the expected endpoint of the ball. With this change, the function getRobotClosestToBall does not actually return the robot closest to the ball, which is confusing and could cause errors.

PR #1334, opened 2022-02-27
Update macOS support to reflect recent developments

daw1012345:
- Default version of protobuf, qt5 (via homebrew) now work with automoc - enable AUTOMOC on macOS

No code changes, only enable AUTOMOC for all platforms (previously only enabled on Linux)


Florisheinen1 (commented):
Still works on Linux. If @7lopol7 also gets it to work, then this can be merged

rttwente (approved):
LGTM. Seeing as only Dawid actually *uses* macos for rtt, if it works for both him and us, we do not need the approval of lotte.

PR #1342, opened 2022-03-25
Defend shot pass11v11

GagataS:
### Comments for the reviewer
- DefendShot and DefendPass for 11v11

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages


src/stp/plays/defensive/DefendShot.cpp
```diff
@@ -23,52 +23,118 @@ DefendShot::DefendShot() : Play() {
...
 }
 
 uint8_t DefendShot::score(PlayEvaluator &playEvaluator) noexcept {
     auto world = playEvaluator.getWorld();
     auto field = world->getField().value();
     auto enemyRobot = world->getWorld()->getRobotClosestToBall(world::them);
+    auto goalVisibility = FieldComputations::getPercentageOfGoalVisibleFromPoint(field, true, enemyRobot->get()->getPos(), world->getWorld().value(), enemyRobot->get()->getId(), false);
     auto position = distanceFromPointToLine(field.getBottomLeftCorner(), field.getTopLeftCorner(), enemyRobot->get()->getPos());
-    return 255 * (field.getFieldLength() - position) / field.getFieldLength();
+    return 255 * (field.getFieldLength() - position) / field.getFieldLength() * goalVisibility/100;
```
alexander-de-ranitz: Perhaps we could use the GoalShotScore for this? Something like PositionScoring::scorePosition(enemyRobot->get->getPos() , gen::GoalShot, field, world).score;

This is also what we do in our Attack play, so it might make sense to employ a similar strategy here

src/stp/plays/defensive/DefendShot.cpp
```diff
@@ -23,52 +23,118 @@ DefendShot::DefendShot() : Play() {
...
+
+    stpInfos["defender_2"].setPositionToDefend(field.getOurGoalCenter());
+    stpInfos["defender_2"].setEnemyRobot(enemyClosestToOurGoalTwo);
     stpInfos["defender_2"].setBlockDistance(BlockDistance::CLOSE);
+
+    std::map<double, Vector2> enemyMap;
+
+    for (auto enemy : enemyRobots){
+        double score = FieldComputations::getDistanceToGoal(field, true, enemy->getPos());
+        enemyMap.template insert(std::pair<double, Vector2>(score, enemy->getPos()));
```
alexander-de-ranitz, line 123:
```suggestion
        enemyMap.insert({score, enemy->getPos()});
```

src/stp/plays/defensive/DefendShot.cpp
```diff
@@ -23,52 +23,118 @@ DefendShot::DefendShot() : Play() {
...
+    enemyRobots.erase(std::remove_if(enemyRobots.begin(), enemyRobots.end(),
+                                     [&](const auto enemyRobot) -> bool { return enemyClosestToOurGoalOne && enemyRobot->getId() == enemyClosestToOurGoalOne.value()->getId(); }),
+                      enemyRobots.end());
+
+    auto enemyClosestToOurGoalTwo = world->getWorld()->getRobotClosestToPoint(field.getOurGoalCenter(), enemyRobots);
+
+    enemyRobots.erase(std::remove_if(enemyRobots.begin(), enemyRobots.end(),
+                                     [&](const auto enemyRobot) -> bool { return enemyClosestToOurGoalTwo && enemyRobot->getId() == enemyClosestToOurGoalTwo.value()->getId(); }),
+                      enemyRobots.end());
+
```
alexander-de-ranitz, line 96: Instead of using erase and remove_if, you should use erase_if. This is more concise and readable, but also safer, as calling erase on the iterator returned by remove_if can give an error if remove_if does not remove anything, as it will return the past-the-end iterator, which can not be erased.

src/stp/plays/defensive/DefendShot.cpp
alexander-de-ranitz:
```suggestion
        enemyMap.template insert(std::pair<double, Vector2>(score, enemy->getPos()));
        enemyMap.insert({score, enemy->getPos()});
```

src/stp/plays/defensive/DefendShot.cpp
```diff
@@ -23,52 +23,118 @@ DefendShot::DefendShot() : Play() {
...
     stpInfos["defender_2"].setBlockDistance(BlockDistance::CLOSE);
+
+    std::map<double, Vector2> enemyMap;
+
+    for (auto enemy : enemyRobots){
+        double score = FieldComputations::getDistanceToGoal(field, true, enemy->getPos());
+        enemyMap.template insert(std::pair<double, Vector2>(score, enemy->getPos()));
+    }
+
+    for (int i = 1; i <= 4; i++){
```
alexander-de-ranitz: There should be checks to make sure enemyMap is not empty, as that would cause errors when trying to erase an element of the vector.

src/stp/tactics/passive/BlockBall.cpp
```diff
@@ -53,6 +53,9 @@ Vector2 BlockBall::calculateTargetPosition(const world::view::BallView &ball, Ve
             // Default distance of 0.5m. If the ball is closer than that to the enemy, stand right in front of the ball instead.
             distance = std::min(0.5, targetToBall.length() - control_constants::ROBOT_RADIUS);
             break;
+        case BlockDistance::PARTWAY:
+            distance = targetToBall.length() * 0.4;
```
alexander-de-ranitz, line 57: When should PARTWAY be used instead of HALFWAY? Is there really a significant difference between 0.4 or 0.5 times the vector length?

src/stp/plays/defensive/DefendShot.cpp
```diff
@@ -23,52 +23,118 @@ DefendShot::DefendShot() : Play() {
...
 }
 
 void DefendShot::calculateInfoForRoles() noexcept {
+    calculateInfoForWallers();
     calculateInfoForDefenders();
     calculateInfoForHarassers();
     calculateInfoForKeeper();
 }
 
+void DefendShot::calculateInfoForWallers() noexcept{
```
alexander-de-ranitz: Do you think using the determineWallPosition function in PositionComputations could be used here? I'm not exactly sure how the results would differ, but it might be interesting to look into

src/stp/plays/defensive/DefendShot.cpp
```diff
@@ -23,52 +23,118 @@ DefendShot::DefendShot() : Play() {
 
     roles = std::array<std::unique_ptr<Role>, stp::control_constants::MAX_ROBOT_COUNT>{
         std::make_unique<role::Keeper>(role::Keeper("keeper")),
-        std::make_unique<role::Defender>(role::Defender("defender_1")),
-        std::make_unique<role::Defender>(role::Defender("defender_2")),
-        std::make_unique<role::Harasser>(role::Harasser("harassing_defender")),
+        std::make_unique<role::Defender>(role::Defender("waller_1")),
```
alexander-de-ranitz: The role names seem a bit confusing. Wallers and midfielders have the Defender role, whilst defenders (and the harasser) have the Harasser role. Do you think the naming could be improved here? Or are the names of the roles them selves not descriptive enough?

alexander-de-ranitz (dismissed):

src/stp/plays/defensive/DefendPass.cpp
```diff
@@ -53,29 +76,53 @@ void DefendPass::calculateInfoForRoles() noexcept {
...
 
-    stpInfos["defender_2"].setPositionToDefend(remainingEnemy->get()->getPos());
-    stpInfos["defender_2"].setBlockDistance(BlockDistance::HALFWAY);
+    for (int i = 1; i <= 3; i++) {
+        stpInfos["midfielder_" + std::to_string(i)].setPositionToDefend(enemyMap.rbegin()->second);
+        stpInfos["midfielder_" + std::to_string(i)].setBlockDistance(BlockDistance::HALFWAY);
+
+        if (!enemyMap.empty()) {
+            enemyMap.erase(prev(enemyMap.end()));
+        } else {
```
alexander-de-ranitz, line 119: When this is true (enemyMap is empty), wouldn't the code in line 117 still try to access this empty map? Perhaps this check for emptiness should be done at the start of the loop.

src/stp/plays/defensive/DefendPass.cpp
```diff
@@ -53,29 +76,53 @@ void DefendPass::calculateInfoForRoles() noexcept {
...
+    stpInfos["defender_helper_2"].setPositionToDefend(enemyClosestToOurGoalTwo->get()->getPos());
+    stpInfos["defender_helper_2"].setBlockDistance(BlockDistance::HALFWAY);
 
-    auto remainingEnemy = world->getWorld()->getRobotClosestToPoint(field.getOurGoalCenter(), enemyRobots);
+    std::map<double, Vector2> enemyMap;
 
-    stpInfos["defender_1"].setPositionToDefend(enemyClosestToOurGoal->get()->getPos());
-    stpInfos["defender_1"].setBlockDistance(BlockDistance::HALFWAY);
+    for (auto enemy : enemyRobots) {
+        double score = FieldComputations::getPercentageOfGoalVisibleFromPoint(field, true, enemy->getPos(), world->getWorld().value(), enemy->getId(), false) *
```
alexander-de-ranitz: Although I think this way of distributing roles is quite clever, I think there is one issue with how it is done now. Midfielder_1 will not always go to the enemy with the highest goal visibility, which is fine, but as soon as midfielder_1 is standing in front of this enemy, its goal visibility will decrease significantly, which means that midfielder_1 will be assigned to a different robot (since a different enemy now has the highest goal visibility). This causes the robots to keep switching between enemies they are defending instead of actually standing in front of them to defend their pass.

src/stp/plays/defensive/DefendShot.cpp
```diff
@@ -22,53 +22,111 @@ DefendShot::DefendShot() : Play() {
...
+    for (auto enemy : enemyRobots) {
+        double score = FieldComputations::getDistanceToGoal(field, true, enemy->getPos());
+        enemyMap.insert({score, enemy->getPos()});
+    }
+
+    for (int i = 1; i <= 4; i++) {
+        stpInfos["midfielder_" + std::to_string(i)].setPositionToDefend(enemyMap.begin()->second);
+        stpInfos["midfielder_" + std::to_string(i)].setBlockDistance(BlockDistance::CLOSE);
+
+        if (!enemyMap.empty()) {
```
alexander-de-ranitz, line 113: Same as in defendPass, this check is too late since line 115 still uses the possibly empty map

alexander-de-ranitz (changes requested):

alexander-de-ranitz:
The new fix looks fine, but the other issue (see below) hasn't been addressed yet as far as I can see.

In DefendPass L108-L122
Midfielder_1 will not always go to the enemy with the highest goal visibility, which is fine, but as soon as midfielder_1 is standing in front of this enemy, its goal visibility will decrease significantly, which means that midfielder_1 will be assigned to a different robot (since a different enemy now has the highest goal visibility). This causes the robots to keep switching between enemies they are defending instead of actually standing in front of them to defend their pass.

alexander-de-ranitz (approved):

alexander-de-ranitz:
> The new fix looks fine, but the other issue (see below) hasn't been addressed yet as far as I can see.
> 
> In DefendPass L108-L122 Midfielder_1 will not always go to the enemy with the highest goal visibility, which is fine, but as soon as midfielder_1 is standing in front of this enemy, its goal visibility will decrease significantly, which means that midfielder_1 will be assigned to a different robot (since a different enemy now has the highest goal visibility). This causes the robots to keep switching between enemies they are defending instead of actually standing in front of them to defend their pass.

For now, I changed the scoring to use the goalAngle, so that should hopefully fix the issue, since it depends only on the enemies position, not our robots. However, this might not be a great way of scoring, so that should be improved later on.

PR #1343, opened 2022-03-29
Feature/world visualization

rolfvdhulst:
This PR adds a visualization option for raw detection frames. This makes debugging vision issues and Roboteam world infinitely easier.


Florisheinen1 (approved):
LGTM. Only adds visualization functionality, so it should not break other stuff

PR #1346, opened 2022-04-04
Refactor Pass Computations and Add Tests

alexander-de-ranitz:
### Comments for the reviewer
This PR refactors passComputations to be more readable and understandable and also improves its functionality by returning more information and ensuring the keeper and passer are correctly identified and distributed. Furthermore, changes are made to attackingPass (and keeperKickBall) to end the play when a better pass is available, preventing the robots from passing to a location which initially might have been a good pass, but is no longer optimal.

###### Code Quality

###### Testing
- [x] All tests are passing.

###### Commit Messages

### DIFF
--- a/src/stp/computations/PassComputations.cpp
+++ b/src/stp/computations/PassComputations.cpp
@@ -13,43 +13,93 @@
 
 namespace rtt::ai::stp::computations {
 
-gen::ScoredPosition PassComputations::calculatePassLocation(Vector2 ballLocation, const std::vector<Vector2>& robotLocations, Vector2 passerLocation, gen::ScoreProfile profile,
-                                                            const rtt::world::World* world, const world::Field& field) {
-    double gridWidth = field.getFieldWidth();
-    double gridLength = field.getFieldLength();
-    int numPoints = 9;
-    gen::ScoredPosition bestPassLocation;
-    bestPassLocation.score = 0;
-    if (robotLocations.empty()) {
-        RTT_WARNING("No possible receivers!")
-        return bestPassLocation;
+PassInfo PassComputations::calculatePass(gen::ScoreProfile profile, const rtt::world::World* world, const world::Field& field, bool keeperCanPass) {
+    PassInfo passInfo;  // Struct used to store the information needed to execute the pass
+
+    auto us = world->getWorld()->getUs();
+    auto ballLocation = world->getWorld()->getBall()->get()->getPos();
+
+    // Find which robot is keeper (bot closest to goal if there was not a keeper yet), store its id, and erase from us
+    passInfo.keeperId = getKeeperId(us, world, field);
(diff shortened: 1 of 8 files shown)


src/stp/computations/PassComputations.cpp
```diff
@@ -13,43 +13,93 @@
...
             }
         }
     }
-    if (bestPassLocation.score == 0) {
-        RTT_DEBUG("No good pass found! Passing to robot furthest in field.")
-        bestPassLocation.position = *std::max_element(robotLocations.begin(), robotLocations.end(), [](auto& p1, auto& p2) { return p1.x > p2.x; });
+
+    if (passInfo.passScore == 0) {
+        // If no good pass is found, pass to the robot furthest in the field
+        passInfo.passLocation = *std::max_element(possibleReceiverLocations.begin(), possibleReceiverLocations.end(), [](auto& p1, auto& p2) { return p1.x > p2.x; });
```
Florisheinen1, line 63: As far as I can tell, there is no guarantee that `possibleReceiverLocations` has at least 1 location. If that is indeed the case, std::max_element might return an iterator that points to the end of the vector. Because you immediately dereference it, this code might crash. You could solve this in 2 ways:
- Either write a small if statement that checks if the return value of the max_element is a valid iterator, or
- Write in the comments of the function that there is a minimum required robots, so other people know when not to call this function

alexander-de-ranitz: Good catch. Added a check to make sure possibleReceviers won't be empty by the time it gets there.

src/stp/plays/offensive/AttackingPass.cpp
```diff
@@ -35,58 +35,46 @@ AttackingPass::AttackingPass() : Play() {
...
     stpInfos["keeper"].setEnemyRobot(world->getWorld()->getRobotClosestToBall(world::them));
-    // TODO: set good pass position
-    stpInfos["keeper"].setPositionToShootAt(world->getWorld()->getRobotClosestToPoint(field.getOurGoalCenter(), world::us).value()->getPos());
-    stpInfos["keeper"].setKickOrChip(KickOrChip::KICK);
 
     /// Midfielder
     stpInfos["midfielder"].setPositionToMoveTo(PositionComputations::getPosition(std::nullopt, field.getMiddleMidGrid(), gen::SafePosition, field, world));
-}
 
-void AttackingPass::calculateInfoForScoredRoles(world::World* world) noexcept {
```
Florisheinen1, line 77: You forgot to implement this function `calculateInfoForScoredRoles`. Compiling this branch results in an undefined reference. (It is a virtual function inherited from Play that is required to implement here)

alexander-de-ranitz: Whoops, had already fixed it but forgot to push. Should be better now

Florisheinen1 (approved):
LGTM, still compiles

PR #1348, opened 2022-04-06
Change kickspeed to non-zero value

robopc:
The kickspeed was set to 0 by the chip function, which is fixed now. Also added some debug statements to say when you kicked and with what speed.

### Comments for the reviewer
- 

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages

### DIFF
--- a/src/manual/JoystickManager.cpp
+++ b/src/manual/JoystickManager.cpp
@@ -122,6 +122,10 @@ void JoystickManager::tickJoystickHandlers() {
         joystickHandler.second->tick();
         auto const &[_, world] = world::World::instance();
         auto robot = world->getWorld()->getRobotForId(joystickHandler.second->getCommand().id);
+        if (joystickHandler.second->getCommand().kickSpeed > 0){
+            RTT_DEBUG("KickSpeed = ", joystickHandler.second->getCommand().kickSpeed);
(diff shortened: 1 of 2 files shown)


src/manual/JoystickManager.cpp
```diff
@@ -122,6 +122,10 @@ void JoystickManager::tickJoystickHandlers() {
         joystickHandler.second->tick();
         auto const &[_, world] = world::World::instance();
         auto robot = world->getWorld()->getRobotForId(joystickHandler.second->getCommand().id);
+        if (joystickHandler.second->getCommand().kickSpeed > 0){
+            RTT_DEBUG("KickSpeed = ", joystickHandler.second->getCommand().kickSpeed);
+            RTT_DEBUG("KickType = ", joystickHandler.second->getCommand().kickType == KickType::KICK ? "Kick" : (joystickHandler.second->getCommand().kickType == KickType::CHIP) ? "Chip" : "NoKick");
+        }
```
Florisheinen1, line 128: The defining factor whether a robot kicks is the kickType. If the kick type is `NO_KICK`, even if the kickSpeed is above 0, it will not kick. In this case, it might be better to first test if the kickType is not `NO_KICK`, and then print the kickSpeed and the type as either `KICK` or `CHIP`

alexander-de-ranitz (approved):
lgtm

PR #1350, opened 2022-04-13
Add roles to attacking pass for 11v11

alexander-de-ranitz:
### Comments for the reviewer
This PR adds more roles to that attacking pass can work with 11 bots. The roles still need to be fine-tuned, but that should be done when other plays are also made 11v11 and a 11v11 match can be played

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages

### DIFF
--- a/src/stp/plays/offensive/AttackingPass.cpp
+++ b/src/stp/plays/offensive/AttackingPass.cpp
@@ -75,6 +75,10 @@ Dealer::FlagMap AttackingPass::decideRoleFlags() const noexcept {
 }
 
 void AttackingPass::calculateInfoForRoles() noexcept {
+    calculateInfoForDefenders();
+    calculateInfoForMidfielders();
+    calculateInfoForAttackers();
+
     /// Keeper
     stpInfos["keeper"].setPositionToMoveTo(field.getOurGoalCenter());
     stpInfos["keeper"].setEnemyRobot(world->getWorld()->getRobotClosestToBall(world::them));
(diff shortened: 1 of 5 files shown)


src/stp/plays/offensive/AttackingPass.cpp
```diff
@@ -27,9 +28,17 @@ AttackingPass::AttackingPass() : Play() {
...
+                                                                                       std::make_unique<role::Passer>(role::Passer("passer")),
+                                                                                       std::make_unique<role::PassReceiver>(role::PassReceiver("receiver")),
+                                                                                       std::make_unique<role::Defender>(role::Defender("defender_left")),
+                                                                                       std::make_unique<role::Defender>(role::Defender("defender_mid")),
+                                                                                       std::make_unique<role::Defender>(role::Defender("defender_right")),
+                                                                                       std::make_unique<role::Formation>(role::Formation("midfielder_left")),
+                                                                                       std::make_unique<role::Formation>(role::Formation("midfielder_middle")),
```
Florisheinen1: Isnt it better for naming consistency to change "midfielder_middle" to "midfielder_mid"?

src/stp/tactics/passive/Formation.cpp
```diff
@@ -20,6 +20,8 @@ Formation::Formation() {
 std::optional<StpInfo> Formation::calculateInfoForSkill(StpInfo const &info) noexcept {
     StpInfo skillStpInfo = info;
 
+    if (!info.getPositionToMoveTo()) return std::nullopt;
+
```
Florisheinen1, line 24: Here you first make a copy of StpInfo, and then check if you should return `nullopt`. In that case, the copying would have been redundant. It should be more efficient to first check if there is no position to mote to before copying it and setting the dribbler speed to 0.

Florisheinen1: For consistency with other tactics, and because this most likely has very little impact on performance, it has been decided to not change this

Florisheinen1 (approved):
LGTM

PR #1365, opened 2022-04-21
Define constant for def. area avoid margin and use this where applicable

alexander-de-ranitz:
### Comments for the reviewer
It is important we do not enter the defense area (or come within 20cm) with our players, as that could easily lead to fouls (when either touching the ball at any moment, or just being there at all during the stop gamestate). Thus, this PR adds a constant to standardise the distance we keep from the defense area (0.4m).

Note that there might still be more places where this constant could/should be used, but those will be added in time.

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages


alexander-de-ranitz:
Note: the hotfix also breaks some tests regarding the margins of our defense area, logically. Once the hotfix is changed into something better, we should make sure these tests pass as well.

TijmenWesteneng (approved):

PR #1388, opened 2022-04-26
Prevent infinite loop

robopc:
To calculate the points of the path, it has a while loop untill the end of the path. Somehow, infinite times are calculated when switching from non-referee to referee mode while playing many plays (e.g. Attack, DefendPass, but not AttackingPass).
I looked into the past history, and this bug has been in existence for at least a month.

THIS IS NOT A SOLUTION. Someone, (preferably @TijmenWesteneng , as it is BB related ) should look into this.

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages


alexander-de-ranitz (approved):
Seems to work as a hotfix

PR #1390, opened 2022-05-02
Refactor avoidBall to better cover edge cases

alexander-de-ranitz:
In some cases, projecting the target pos outside the area that needs to be avoided is not sufficient, as this projected position could be an invalid position. To account for this, a new function is added that can deal with these edge cases by generating points around the target position and checking if they are suitable positions to go to. This is slower than projecting and might return a point that is slightly further than it needs to be, but it should come up with a reasonable position to go to in (almost) all cases.

This PR is dependent on [this PR in utils](https://github.com/RoboTeamTwente/roboteam_utils/pull/119).

###### Code Quality

###### Testing
- [x] All tests are passing.

###### Commit Messages


rttwente (approved):
LGTM

PR #1400, opened 2022-05-31
Fix widget locking world for longer than necessary

mamiksik:
`rtt::world::World::instance()` returns a guard lock, which gets released
at the end of the scope (used to be once the paintEvent returned). This
causes STP to be delayed if `Input::getDrawings()` vector is large (over
3000). Since the lock release is postponed until the method finishes painting all data in the vector (despite not needing the world anymore).

Releasing the lock imminently after the field finished painting mitigates
this issue. As far as I can tell, that is the intended behavior since
`Input::getDrawings()` holds the separate lock. Thus, it does not require
the world to be locked during drawing.

### Comments for the reviewer
- 

###### Code Quality

###### Testing
- [x] All tests are passing.

###### Commit Messages


mamiksik:
Update, overfilling drawings causes deadlock I think.

src/interface/widgets/widget.cpp
```diff
@@ -21,30 +21,32 @@ void Visualizer::paintEvent(QPaintEvent *event) {
...
-    }
+    {
+        // worldPtr is scoped in order to release the lock soon as possible
+        auto const &[_, worldPtr] = rtt::world::World::instance();
+        auto const &world = worldPtr->getWorld();
+        auto const &field = worldPtr->getField();
 
-    auto s = QString::fromStdString("We have " + std::to_string(world->getUs().size()) + " robots");
-    painter.drawText(24, 48, s.fromStdString("We have " + std::to_string(world->getUs().size()) + " robots"));
+        if (!world.has_value() || !world.value()->weHaveRobots()) {
```
alexander-de-ranitz: I've always been bothered by the fact that we say we done have a world if we don't have any robots. Could we change this to be shown only when there is no world? If we have no robots, we can still draw the field (and enemy robots)

Florisheinen1:
I approve of the changes made, but I also think it is nice to implement the requested change Alex suggested.

src/interface/widgets/widget.cpp
mamiksik: ok I will have look into that

mamiksik:
@alexander-de-ranitz @Florisheinen1 Fixed

Florisheinen1 (approved):
LGTM

PR #1408, opened 2022-06-27
Chore/clean ball

rttwente:
Observer filters the ball data. This should be quite accurate. When AI receives this filtered data, it starts filtering again, resulting in a delay. This is of course completely redundant and unnecessary, so this PR removes this filtering and cleans up the ball class in general by moving constants around and making the class public. The latter can be done because the only way to access the ball is by means of a const reference, so accidentally editing the values is impossible. The added benefit is that it takes less function calls to get to these values, and the values do not have to be copied over and over.

Here are plots to show the delay that the filtering of AI causes.
KickPlot

Above you see the total graph of a kick of a ball, and below a zoom, where you can easily see the added difference of AI to Observer to Simulator

KickPlotZoom

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages

### DIFF
--- a/docs/Ball.md
+++ b/docs/Ball.md
@@ -27,13 +27,13 @@ serves the purpose of representing the ball. Things like position and velocity a
 
 `updateBallAtRobotPosition` -> Updates `expectedEndPosition` and draws to the interface.
 
-`getPos` -> Returns `position`
+`getPosBall` -> Returns `position`
 
-`getVelocity` -> Returns `velocity`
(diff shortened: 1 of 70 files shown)


alexander-de-ranitz:
@Florisheinen1 Your part looks good to me. If you could take a look at the commits I added that'd be great.

Florisheinen1 (approved):
LGTM

PR #1412, opened 2022-07-03
Check for empty optional before access

robopc:
Pretty straightforward, I forgot to check for an empty optional here. This can cause the AI to crash in case the robot we are looking for isn't in world. This is why AI sometimes crashed when a robot was obscured.


Florisheinen1 (approved):
LGTM

PR #1424, opened 2022-07-10
Fix ball placement

alexander-de-ranitz:
### Comments for the reviewer
This PR improves ball placement by making the following changes:
- Only use ballStandback to move away from the ball by 15cm, since avoidBall should be used to actually go to a good ball avoiding position. BallStandBack is only used to make sure stop dribbling the ball after we place it.
- Changing when DriveWithBall to only reset when the angle to the ball > hasBallAngle instead of GoToPosAngleErrorMargin. The latter is a much smaller margin, which can quite easily be exceeded when driving with the ball even when nothing goes wrong.
- Make sure we actually avoid ball collisions if we set shouldAvoidBall to true

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages

### DIFF
--- a/src/control/positionControl/BBTrajectories/WorldObjects.cpp
+++ b/src/control/positionControl/BBTrajectories/WorldObjects.cpp
@@ -3,10 +3,7 @@
 //
 #include "control/positionControl/BBTrajectories/WorldObjects.h"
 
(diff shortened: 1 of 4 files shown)


robopc (approved):
LGTM

PR #1480, opened 2022-09-08
de-hardcode distance for BlockRobot tactic

JAndringa:
distance from enemy robot to blocking position was hardcoded in BlockRobot.cpp, changed it, so now it is relative to the velocity of the velocity of the robot still requires fine tuning

### Comments for the reviewer
- changed the way distance is calculated but requires fine tuning so will probably have unwanted behaviour

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages


silashoevers (approved):
Some values indeed still need some fine tuning as mentioned, but hardcoding has been removed. I have expressed my worries regarding the resulting differences in CLOSE, HALFWAY and FAR but they can be fixed at a later stage.

PR #1481, opened 2022-09-13
fix for roboteam_ai_tests not compiling

### LINKED ISSUE #1482: fix gtests for ai

JAndringa:
fixed syntax and commented out all nonworking functions. getId() in robot.cpp returns null so the that doesn't pass the test but everything else should now be testable. :)

### Comments for the reviewer
- 

###### Code Quality

###### Testing
- [x] All tests are passing. -> No but should be fixed separately as we need the tests to work before we can test.
- [x] I _added new / changed existing_ tests to reflect code changes (state why not otherwise!) -> made the test actually working
- [x] I tested my changes manually (Describe how, to what extent etc.) -> compiles and runs

###### Commit Messages

### DIFF
--- a/include/roboteam_ai/world/World.hpp
+++ b/include/roboteam_ai/world/World.hpp
@@ -163,7 +163,6 @@ class World {
      * Everything EXCEPT positioncontrol and settings is reset.
      */
     void reset() noexcept {
-        updateMap.clear();
         history.clear();
         currentIndex = 0;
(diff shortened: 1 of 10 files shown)


silashoevers (changes requested):
Fix  accidental naming issue in WorldData.cpp

silashoevers (approved):

PR #1490, opened 2022-11-10
Keeper should not chase ball outside defense area

robopc:
keeper would chase the ball even outside our defense area when the play KeeperKickBall is active. changed it so the play terminates when the ball is not in our defense area

### Comments for the reviewer
- 

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages

### DIFF
--- a/src/stp/plays/defensive/KeeperKickBall.cpp
+++ b/src/stp/plays/defensive/KeeperKickBall.cpp
@@ -118,6 +118,9 @@ bool KeeperKickBall::shouldEndPlay() noexcept {
             1.05 * stp::PositionScoring::scorePosition(passInfo.passLocation, gen::AttackingPass, field, world).score)
         return true;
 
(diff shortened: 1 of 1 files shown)


rolfvdhulst (approved):
LGTM.

alexander-de-ranitz:
I see this is already merged by now, but I think it probably shouldn't be. First of all, if I'm not mistaken, this does exactly the opposite of what you want, as it ends the play if the ball IS in the defense area. Secondly, ending the play as soon as the ball leaves the defense area means that we will also end the play when the ball was kicked out of the defense area by our keeper when passing. As receiving the ball is part of the same play, this is probably not what you want to do, since it will prevent the receiver from actually trying to receive the ball.

As a simple solution, I think you could change the condition to something like this: 
`if (!ballKicked() && !ballIsInOurDefenseArea(...)) return true;`

PR #1491, opened 2022-11-28
fixed/improved angular orbit

JAndringa:
angular orbit was not working properly, so I just rewrote it. Needs some finetuning but works well enough for now.

### Comments for the reviewer
- 

###### Code Quality

###### Testing
- [ ] All tests are passing.

###### Commit Messages

### DIFF
--- a/src/control/ControlUtils.cpp
+++ b/src/control/ControlUtils.cpp
@@ -86,7 +86,7 @@ double ControlUtils::determineKickForce(const double distance, stp::ShotType sho
 
     double kickForce;
     if (shotType == stp::ShotType::PASS) {
-        kickForce = 1.5;
+        kickForce = 1.8;
     } else if (shotType == stp::ShotType::TARGET) {
         kickForce = 0.5;
     } else {
--- a/src/stp/skills/Kick.cpp
+++ b/src/stp/skills/Kick.cpp
@@ -6,12 +6,14 @@
 
 #include "stp/constants/ControlConstants.h"
 
+#include "roboteam_utils/Print.h"
+
 namespace rtt::ai::stp::skill {
 
 Status Kick::onUpdate(const StpInfo &info) noexcept {
     // Clamp and set kick velocity
     float kickVelocity = std::clamp(info.getKickChipVelocity(), control_constants::MIN_KICK_POWER, control_constants::MAX_KICK_POWER);
-
+    RTT_DEBUG(kickVelocity)
     // Set kick command
     command.kickType = KickType::KICK;
     command.kickSpeed = kickVelocity;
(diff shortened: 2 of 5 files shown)


umer99roboteam:
The orbit of the robot is much more smooth
As a result the passing has vastly improved and gameplay is faster
The orbiting seems quite fast. The actual robot might loose the loose the ball. However, seems that can be tuned

umer99roboteam (approved):

rolfvdhulst:
Has this been tested with real robots?

src/stp/skills/Kick.cpp
```diff
@@ -6,12 +6,14 @@
...
 
+#include "roboteam_utils/Print.h"
+
 namespace rtt::ai::stp::skill {
 
 Status Kick::onUpdate(const StpInfo &info) noexcept {
     // Clamp and set kick velocity
     float kickVelocity = std::clamp(info.getKickChipVelocity(), control_constants::MIN_KICK_POWER, control_constants::MAX_KICK_POWER);
-
+    RTT_DEBUG(kickVelocity)
```
rolfvdhulst, line 16: Remove before merging.

src/control/ControlUtils.cpp
```diff
@@ -86,7 +86,7 @@ double ControlUtils::determineKickForce(const double distance, stp::ShotType sho
 
     double kickForce;
     if (shotType == stp::ShotType::PASS) {
-        kickForce = 1.5;
+        kickForce = 1.8;
```
rolfvdhulst, line 89: Seemse unrelated to what the pull request is about.

rolfvdhulst (changes requested):
See comments. Nothing major.

alexander-de-ranitz:
As Rolf asked, was this actually tested on the field? Based on "The actual robot might loose the loose the ball", I'm guessing not :)

Based on the changes you made, I can believe this works better in GrSim, but I highly doubt it does what you expect on the field. A lot of the code you changed/removed in this PR was specifically designed and tuned to improve performance on the field (probably at the expense of performance in GrSim, but we didn't really look at this much as we created this quite shortly before the robocup). Things such as padding the angle to account for delay and delegating the decision of when to shoot to the robot instead of the AI can be really useful with actual robots but are quite useless/counterproductive in the simulator. Of course, it is still important everything works well in the simulator, but doing so at the expense of the performance on the field isn't a good approach IMO. Of course, if needed, you could add some `if (playingOnTheField)` and `if (playingInSim)` to deal with cases where we want to do things differently due to fundamental differences between actual robots and the simulation, just be sure not to let the sim and real-life performance get too different.

